PR #5, opened 2015-03-05
simplify c interface

klokane:
Implemented  c-interface.
Serialization of parse result


Makefile
```diff
@@ -12,6 +12,13 @@ V ?= 1
 # Targets
 all: drafter
 
+libdrafter: config.gypi $(BUILD_DIR)/Makefile
+	$(MAKE) -C $(BUILD_DIR) V=$(V) $@
+
+test-libdrafter: config.gypi $(BUILD_DIR)/Makefile
+	$(MAKE) -C $(BUILD_DIR) V=$(V) test-libdrafter
```
pksunkara: `$@` instead of `test-libdrafter`

Makefile
```diff
@@ -34,10 +41,11 @@ distclean:
...
-test: drafter
+test: libdrafter test-libdrafter drafter
+	./bin/test-libdrafter
 
 ifdef INTEGRATION_TESTS
 	bundle exec cucumber
 endif
 
-.PHONY: all drafter test
+.PHONY: all libdrafter drafter test
```
pksunkara: Why is `test-libdrafter` not here?

drafter.gyp
```diff
@@ -39,5 +43,61 @@
...
+      "type": "executable",
+      "sources": [
+        "src/main.cc",
+        "src/config.cc",
+        "src/config.h",
+        "src/reporting.cc",
+        "src/reporting.h",
+      ],
+
+      # FIXME: replace by direct dependecies
```
pksunkara: I am sorry but what is this comment about?

src/SerializeSourceAnnotations.cc
```diff
@@ -0,0 +1,49 @@
...
+
+sos::Object drafter::WrapSourceAnnotations(const snowcrash::Report& report, const snowcrash::SourceMap<snowcrash::Blueprint>& sourceMap)
+{
+    sos::Object object;
+
+    object.set(SerializeKey::AnnotationsVersion, sos::String(AST_ANNOTATION_SERIALIZATION_VERSION));
+    
+    sos::Object ast;
+    ast.set(SerializeKey::ASTVersion, sos::String(AST_SERIALIZATION_VERSION));
+    object.set(SerializeKey::Ast, ast);
```
pksunkara: Shouldn't the `ast` here contain the actual serialised AST instead of just the version?

pksunkara: If we don't need it here, then we should probably not even add `ast` key here.

test/test-drafter.h
```diff
@@ -0,0 +1,32 @@
+#ifndef DRAFTER_DRAFTERTEST_H
+#define DRAFTER_DRAFTERTEST_H
+
+#include "catch.hpp"
+
+#include "stream.h"
+
+class it_fixture_files {
```
pksunkara: Shouldn't the class name be camel cased? ex: `ITFixtureFiles`

test/test-cdrafter.cc
```diff
@@ -0,0 +1,42 @@
+#include "test-drafter.h"
+
+#include "cdrafter.h"
+
+#include <string.h>
+
+
```
pksunkara: Extra line.

test/test-drafter.h
```diff
@@ -0,0 +1,32 @@
...
+
+        stream_type in = CreateStreamFromName<std::istream>(filename);
+        std::stringstream strStream;
+        strStream << in->rdbuf();
+
+        return strStream.str();
+    }
+
+    const std::string get(const std::string& ext) const {
+        return fetchContent(base_ + ext);
```
pksunkara: Maybe make this fetch the fixture from `test/fixtures` by default. So, this function becomes:

```cpp
const std::string get(const std::string& ext, const std::string& fixtureFolder = "test/fixtures/") const;
```

test/test-cdrafter.cc
```diff
@@ -0,0 +1,42 @@
...
+TEST_CASE("c-interface parses blueprint correctly","[c-interface]")
+{
+    it_fixture_files fixture = it_fixture_files("test/fixtures/annotations-with-warning");
```
pksunkara: Won't this test fail on windows since their file system path notation is different? So, reading from "test/fixtures" will give an error.

test/test-cdrafter.cc
```diff
@@ -0,0 +1,42 @@
...
+    REQUIRE(ast != NULL);
+
+    REQUIRE(strcmp(ast, fixture.get(".ast.json").c_str()) == 0);
+    REQUIRE(strcmp(report, fixture.get(".result.json").c_str()) == 0);
+
+    free(ast);
+    free(report);
+}
+
+TEST_CASE("just check result, no memory alloc","[c-interface]")
```
pksunkara: Please rename the test case properly.

src/main.cc
```diff
@@ -0,0 +1,91 @@
...
+ */
+void Serialization(std::tr1::shared_ptr<std::ostream> stream,
+                   const sos::Object& object,
+                   sos::Serialize* serializer)
+{
+    serializer->process(object, *stream);
+    *stream << "\n";
+}
+
+
```
pksunkara: Extra line

src/cdrafter.cc
```diff
@@ -0,0 +1,56 @@
...
+        std::stringstream astStream;
+        serializer.process(drafter::WrapBlueprint(blueprint.node), astStream);
+        astStream << "\n";
+        *ast = ToString(astStream);
+    }
+
+
+    if (report) {
+        std::stringstream reportStream;
+        serializer.process(drafter::WrapSourceAnnotations(blueprint.report, blueprint.sourceMap), reportStream);
```
pksunkara: Why does allocation memory for `report` also gives back the `sourceMap`? I don't understand this.

drafter.gyp
klokane: I will resolve it later.

There are some troubles while include `snowcrash/common.gypi` and dependencies are no working as declared there:
https://code.google.com/p/gyp/wiki/InputFormatReference#Dependencies_and_Dependents

Will fix after cleanup snowcrash

src/SerializeSourceAnnotations.cc
klokane: Oh I miss `...` in https://github.com/apiaryio/api-blueprint-ast/blob/zdne/elements-ast/Parse%20Result.md

AST member, probably, should contain fully parsed AST. 
@zdne  isn't it?

test/test-drafter.h
klokane: agree, will fix

pksunkara:
You have 2 functions currently

* `WrapBlueprint` which wraps `ast` (*Sounds good to me*)
* `WrapSourceAnnotations` which wraps `report` and `sourceMap` (*My mind goes WHAT?*)

I would suggest changing it to either:

* `WrapBlueprint` which wraps `ast`
* `WrapParseResult` which wraps `ast`, `sourceMap`, `report`. (The spec for `ParseResult` is [here](https://github.com/apiaryio/api-blueprint-ast/blob/master/Parse%20Result.md))

or:

* `WrapBlueprint` which wraps `ast`
* `WrapSourcemap` which wraps `sourceMap`
* `WrapReport` which wraps `report`

src/SerializeSourceAnnotations.cc
pksunkara: Check the [comment](https://github.com/apiaryio/drafter/pull/5#issuecomment-77376081) I was writing while you wrote this comment.

src/cdrafter.cc
klokane: due to specification https://github.com/apiaryio/api-blueprint-ast/blob/zdne/elements-ast/Parse%20Result.md

test/test-cdrafter.cc
klokane: Will solve in GYP by define for preprocessor macro

klokane:
Updated

pksunkara:
Just one small thing which is blocking. We can chat about it if you want.

klokane:
Oh I missed following  notice 

> sourcemap (Source Map) - This is the sourcemap tree of the parsed blueprint whose content may not be present if the option to export source maps is not present

in parse result specification https://github.com/apiaryio/api-blueprint-ast/blob/master/Parse%20Result.md

Ok will fix, I meant it is mandatory member

src/SerializeResult.h
```diff
@@ -0,0 +1,27 @@
+////
+//  SerializeSourceAnnotations.h
+//  drafter
+////
```
pksunkara: I think you mistyped this line. Shouldn't it be `//`? Similarly the 1st line of this file.

src/SerializeResult.h
```diff
@@ -0,0 +1,27 @@
...
+
+#include "SectionParserData.h" // required by BlueprintParserOptions
+
+namespace snowcrash { 
+    struct Blueprint;
+    template <typename T> struct ParseResult; 
+}
+
+namespace drafter {
+
```
pksunkara: This line has extra white space.

pksunkara:
@klokane Just two very minor style fixes needed.

src/SerializeResult.h
klokane: Update formating rules :)

PR #8, opened 2015-03-10
Update snowcrash

klokane:
- update snowcrash to MSON branch
- add modules cmdline && sos
- fix warning on gcc 4.6 in tests

### DIFF
--- a/drafter.gyp
+++ b/drafter.gyp
@@ -5,6 +5,19 @@
   
   "targets" : [
     {
+      'target_name': 'libsos',
+      'type': 'static_library',
+      'include_dirs': [
+        'ext/sos/src'
+      ],
+      'sources': [
+        'ext/sos/src/sos.cc',
+        'ext/sos/src/sos.h',
+        'ext/sos/src/sosJSON.h',
+        'ext/sos/src/sosYAML.h'
+      ]
+    },
+    {
       "target_name": "libdrafter",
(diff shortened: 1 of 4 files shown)


drafter.gyp
```diff
@@ -29,18 +42,17 @@
 
       # FIXME: replace by direct dependecies
```
pksunkara: This warning should be removed, right?

drafter.gyp
```diff
@@ -86,19 +98,19 @@
...
```
pksunkara: Similarly, this warning.

pksunkara:
Just 2 minor things I want you to address before I can merge.

klokane:
Still in backlog. Will fix in future

pksunkara:
> Still in backlog.

Okay, but could you explain what exactly they are about? I think I misunderstood what they are about.

klokane:
See chapter *direct_dependent_settings* 

https://code.google.com/p/gyp/wiki/InputFormatReference#Dependencies_and_Dependents

In short it allow "inject" dependencies like include dirs, define and other to other targets. 
eg. there should be direct dependencies in `libdrafter` target and `drafter` shoud use them instead of repeat include paths. But while there is `common.gypi` included, direct dependencies doesn't works. 

Still have no time to investigate for reason.

PR #10, opened 2015-03-12
Remove tr1 from source codes due to troubles w\ building on diferent OS

klokane:
Should fix troubles with tr1
@pksunkara or @zdne  please test on OS X

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -43,12 +43,13 @@ sos::Serialize* CreateSerializer(const std::string& format)
 /**
  * \brief Serialize sos::Object into stream
  */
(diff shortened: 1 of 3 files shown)


klokane:
In original code I use shared_ptr<>, it allows inject `deleter` as 2nd parameter. But due to shared_ptr<> is tr1 part i must to replace it by auto_ptr<>.

This deleter solve problem with closing. Because std::cout/std::cin will be close on delete.

So I create some kind of "proxy" which connect std::cout/cin and redirect write/read operations to its buffer.  This avoid close std io while delete.

PR #14, opened 2015-03-23
Fix fstream behaviour on windows 

klokane:
Add fstream open in `binary` mode.
In text mode on windows is char `\n` automatic transformed into `\r\n`
This fix avoid this behaviour


pksunkara:
Looks good. But it looks like the unit tests are not working on appveyor CI.

klokane:
will check

PR #19, opened 2015-03-25
Prepare v0.1.0 release

pksunkara:
This PR updates snowcrash submodule to version v1.0.0 and prepares drafter release for v0.1.0


tools/homebrew/drafter.rb
```diff
@@ -2,7 +2,7 @@
 
 class Drafter < Formula
   homepage 'http://apiblueprint.org'
-  head 'https://github.com/apiaryio/drafter.git'
+  head 'https://github.com/apiaryio/drafter.git', :tag => 'v0.1.0'
```
pksunkara: @zdne I am wondering if this is needed or we can just use the tip always.

zdne: Having the explicit tag here prevents from installing any commits into master that weren't yet released. So it is somewhat safer. Unless one forgets to change it, that's it.

ext/snowcrash
```diff
@@ -1 +1 @@
-Subproject commit 65e01df77c0e7ad86efb3c0efa4cbc69c846db98
+Subproject commit 76560473da89b5ba4be0a4826288ced2a9712533
```
zdne: I would prefer doing this after the release (just to make sure there are no other commits in the SC master).

pksunkara: As I said, SC master is frozen and once this PR is merged, drafter master will also be frozen. This is what I mean by "preparing" releases. This way, we save some time when we are doing releases instead of doing more grunt work.

README.md
```diff
@@ -12,7 +12,7 @@ API Blueprint is Web API documentation language. You can find API Blueprint docu
 Additionally Drafter provide set of Wrappers for serialization, of parsing result, via  [SOS library](https://github.com/apiaryio/sos) into JSON and YAML format.
 
 ## Status
-- [Format 1A7](https://github.com/apiaryio/api-blueprint/releases/tag/format-1A7) fully implemented 
+- [Format 1A8](https://github.com/apiaryio/api-blueprint/releases/tag/format-1A8) fully implemented
```
zdne: Also I wonder if this line shouldn't be removed from Snow Crash (duplicity).

pksunkara: Snowcrash is an open source library on it's own. Which is why I let it stay in it's README even though this is a bit duplicitous.

PR #30, opened 2015-04-15
Make version check work in Python <2.7

tobiashm:
In Python 2.6 `sys.version_info` is a tuple, so doesn't have a `.major` attribute.
Using the index access method works in both 2.6, 2.7 and 3

### DIFF
--- a/configure
+++ b/configure
@@ -8,7 +8,7 @@
 import sys
 
 
-if sys.version_info.major != 2:
+if sys.version_info[0] != 2:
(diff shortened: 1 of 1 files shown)


tobiashm:
This currently also causes a failure in the redsnow gem, since it uses drafter.
Discussed in https://github.com/apiaryio/redsnow/issues/51

kylef:
Thanks @tobiashm.

PR #46, opened 2015-06-12
refract type expansion

klokane:
Preview of element expansion, call for review of expanded elements.
@pksunkara @smizell @zdne @danielgtaylor 

Please look at 
`test/fixtures/mson-inner-inheritance.apib`
`test/fixtures/mson-inheritance.apib`

and they expanded pairs `json` if expansion is handled correctly
NOTE: there is not implemented copying of meta element as changed in last revision of spec


zdne:
@klokane by 

> Preview of element expansion

You mean what exactly? Can it be merged or not?

klokane:
@zdne still not. I mean I already label this PR as Do Not Merge, will fix it

Still working on Mixins expansion.
But Inheritance should be done - this is request to check validity of expanded elements

zdne:
@klokane Careful the fixtures DOES NOT seem to be correct :skull: 

The elements in `extend`'s content should be of the same type – in this case objects. For example 

**Customer inheriting from User expanded:**


```json
{
  "element": "extend",
  "meta": {
    "id": "Customer"
  },
  "content": [
    {
      "element": "object",
      "meta": {
        "ref": "User"
      },
      "content": [
        {
          "element": "member",
          "meta": {
            "description": "user name"
          },
          "content": {
            "key": {
              "element": "string",
              "content": "login"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    },
    {
      "element": "object",
      "content" : [
        {
          "element": "member",
          "meta": {
            "description": "unique number"
          },
          "content": {
            "key": {
              "element": "string",
              "content": "id"
            },
            "value": {
              "element": "number",
              "content": 0
            }
          }
        }        
      ]
    }
    
  ]
}
```

zdne:
@klokane please check https://github.com/refractproject/refract-spec/blob/master/namespaces/mson-namespace.md#expanded-mson-refract

klokane:
Hi there is revisited version of expansion.
Changes:
- Inheritance: additional object envelope (reported by @zdne)
- Mixins: implemented expansion of  `ref` element

Please as in previous version check if expansion is correct. 
it shoud be in files:
```
test/fixtures/mson-inner-inheritance.apib
test/fixtures/mson-inheritance.apib
test/fixtures/mson-mixin.apib
```

test/fixtures/mson-inheritance.json
```diff
@@ -12,11 +12,15 @@
         {
           "element": "object",
           "meta": {
-            "id": "User"
+            "id": "User",
+            "description": "The User description\n\n"
```
smizell: Is the plan for serializers to account for these extra line breaks?  Seems like if you go from API Blueprint to Refract then back to API Blueprint, you may get extra spaces.

test/fixtures/mson-inner-inheritance.json
```diff
@@ -20,33 +20,76 @@
...
+                            "value": {
+                              "element": "number"
+                            }
+                          }
                         }
-                      }
+                      ]
+                    },
+                    {
+                      "element": "object"
```
smizell: Question: are you all OK with using this empty object like this? Is there a need to include it like this?

test/fixtures/mson-mixin-nonexistent.json
```diff
@@ -0,0 +1,109 @@
+{
+  "_version": "3.0",
+  "metadata": [],
+  "name": "<API name>",
+  "description": "<API description>\n\n",
+  "element": "category",
+  "resourceGroups": [],
```
smizell: Question: is there a reason to not include this kind of data in a root-level element?

smizell:
A couple of questions:

* These examples are expanded versions of the representations, not the original. Is this correct?
* If these are the expanded versions, why do you not extend the content rather than including all of the extend elements? Is there a reason? Do you need the un-extended data?

Had a few questions throughout, but otherwise, the fixtures look good to me.

test/fixtures/mson-inheritance.json
klokane: It is copy of "raw" text in form as it come from parser without any changes.
In general we can add some kind of "filters"

test/fixtures/mson-inner-inheritance.json
klokane: Z wrote in comment above
:
> @klokane Careful the fixtures DOES NOT seem to be correct ￼
>  The elements in extend's content should be of the same type – in this case objects. For exampleof original members. It is not problem to remove if it is empty.

And if original is empty object, result is same :)

It is not problem if orginal emenet is empty, but there is question: which way is correct?

test/fixtures/mson-mixin-nonexistent.json
klokane: Refract Elements are expanded in place of original "DataStructure". 
One of acceptance criteria is be backward compatible.

But in general it ca be anywhere

klokane:
@smizell 
- yes it is
- i'm not sure what do you mean? Can you, please, introduce it in some example?

test/fixtures/mson-mixin-nonexistent.json
smizell: Ah, I see. Thanks.

smizell:
@klokane For example, in one of the examples, you have this:

```json
{
  "element": "extend",
  "meta": {
    "id": "Other Customer"
  },
  "content": [
    {
      "element": "object",
      "meta": {
        "description": "The User description\n\n",
        "ref": "User"
      },
      "content": [
        {
          "element": "member",
          "meta": {
            "description": "user name"
          },
          "content": {
            "key": {
              "element": "string",
              "content": "login"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    },
    {
      "element": "object"
    }
  ]
}
```

Instead of:

```json
{
  "element": "object",
  "meta": {
    "description": "The User description\n\n",
    "ref": "User",
    "id": "Other Customer"
  },
  "content": [
    {
      "element": "member",
      "meta": {
        "description": "user name"
      },
      "content": {
        "key": {
          "element": "string",
          "content": "login"
        },
        "value": {
          "element": "string"
        }
      }
    }
  ]
}
```

Is the reason because you are needing to retain from what they were extended? **Note**: I'm just curious why it's done this way, and don't think it's wrong to do it this way.

klokane:
It is not problem to "merge" it into object. I implement it as it is in specification.

Pros of extend object
- possibility of "backward reconstruction" of all original object
- you can see what property comes from which source

Cons
- you need some additional operation to work with them
- require more effort to create JSON object from them 

BTW, is there way to merge "inheritance tree".

As is in https://github.com/apiaryio/drafter/blob/klokane/refract-type-expansion/test/fixtures/mson-inheritance.apib
```
## User (object) 
- login (string)

## Customer (User)
- id: 0 (number)

## Vip Customer (Customer)
- discount: 15 (default, number) - in percent 0 - 100
```

While merge you have to set name of inherited element into `meta.ref`, but for `Vip Customer` there you have `Customer` and `User`. Which one should be placed into `ref`?

src/RefractAST.cc
```diff
@@ -664,7 +614,7 @@ namespace drafter
         
         //e->meta[key::Title] = IElement::Create(ds.name.symbol.literal);
```
pksunkara: Is this going to be resolved?

src/RefractAST.cc
```diff
@@ -719,15 +670,30 @@ namespace drafter
                 throw std::runtime_error("Unhandled type of DataStructure");
         }
 
-        if (!element) {
+        return element;
+    }
+
+    sos::Object SerializeRefract(refract::IElement* element, const refract::Registry& registry) {
+
+        if(!element) {
```
pksunkara: `if (`

src/RefractAST.cc
```diff
@@ -719,15 +670,30 @@ namespace drafter
...
             return sos::Object();
         }
 
+        refract::ExpandVisitor expander(registry);
+        expander.visit(*element);
 
-        SerializeVisitor serializer;
+        if(refract::IElement* expanded = expander.get()) {
```
pksunkara: `if (`

src/RefractAST.cc
```diff
@@ -719,15 +670,30 @@ namespace drafter
...
+           element = expanded;
+        }
+
+        refract::SerializeVisitor serializer;
         serializer.visit(*element);
-        delete element;
+
+        if(expander.get()) {
```
pksunkara: `if (`

src/SerializeAST.cc
```diff
@@ -437,8 +451,15 @@ sos::Object WrapTypeSection(const mson::TypeSection& section)
 sos::Object WrapDataStructure(const DataStructure& dataStructure)
 {
 #if _WITH_REFRACT_
-    return DataStructureToRefract(dataStructure);
-#else
+    refract::IElement* element = DataStructureToRefract(dataStructure);
+    sos::Object object = SerializeRefract(element, NamedTypesRegistry);
+
+    if(element) {
```
pksunkara: `if (`

src/RefractAST.cc
klokane: Still not.

test/fixtures/mson-mixin-nonexistent.apib
```diff
@@ -0,0 +1,12 @@
...
+
+# Data Structures
+
+## Address (object)
+- city
+- street
+
+## User (object)
+- id
+- Include Address
```
pksunkara: I don't understand why this file is named `mson-mixin-nonexistent` :(

src/SerializeAST.cc
```diff
@@ -754,10 +774,45 @@ bool IsElementResourceGroup(const Element& element)
...
+void findNamedDataStructures(const snowcrash::Elements& elements, DataStructures& found) {
+    for (snowcrash::Elements::const_iterator i = elements.begin()
+       ; i != elements.end()
+       ; ++i) {
+
+        if (i->element == snowcrash::Element::DataStructureElement) {
+            found.push_back(&(i->content.dataStructure));
+        }
+        else if (!i->content.resource.attributes.empty()) {
+            found.push_back(&i->content.resource.attributes);
```
pksunkara: The `content.resource` is an element, right? Then the data structure should be in it's content but you are checking `attributes` here.

Can you add some full API Blueprint tests from drafter.js here so that these kind of things can be checked?

pksunkara:
I had a suspicion on Friday and commented it here. Can you please look into it? If I am right, we need to fix it and add tests. If I am wrong, please add the tests anyway.

test/fixtures/mson-mixin-nonexistent.apib
klokane: Ahh sorry orginaly there were
```
## User(XXX)
```

will fix

klokane:
Sry, it probably lost while rebase, it was same comment as above?

src/SerializeAST.cc
klokane: I discuss it with @zdne while implement searching for named Types.

Quoted && translated (originally in Czech:)
Named types comes from two sources
first:
```
# R [/r]
+ attributes 
    + p
```
second:
```
# Data Structures
## R
+ p
```

did we miss something?

Will add fixture as you suggest

klokane: But, yes it looks it is wrong, first example is expanded as "resource".
```
  "content": [
    {
      "element": "category",
      "content": [
        {
          "element": "resource",
          "name": "R",
          "description": "",
          "uriTemplate": "/resource",
          "model": {},
          "parameters": [],
          "actions": [],
          "content": [
            {
              "element": "object",
              "meta": {
                "id": "R"
              },
              "content": [
                {
                  "element": "member",
                  "content": {
                    "key": {
                      "element": "string",
                      "content": "p"
                    },
                    "value": {
                      "element": "string"
                    }
                  }
                }
              ]
            }
          ]
        }
      ]
    },
```

good catch

klokane: But it look, we are nearly MSON Resources to Refract Elements implementation :)

src/refract/AppendDecorator.h
```diff
@@ -0,0 +1,75 @@
+//
+//  refract/AppendDecotarod.h
```
pksunkara: Wrong.

src/refract/AppendDecorator.h
```diff
@@ -0,0 +1,75 @@
...
+//  librefract
+//
+//  Created by Jiri Kratochvil on 17/06/15.
+//  Copyright (c) 2015 Apiary Inc. All rights reserved.
+//
+#ifndef _REFRACT_APPENDDECORATOR_H_
```
pksunkara: We don't use `_` before and after these. Please change this to `REFRACT_APPENDDECORATOR_H` and similarly all  the other places you have done this.

src/refract/ComparableVisitor.cc
```diff
@@ -0,0 +1,40 @@
+//
+//  refract/SerializeCompactVisitor.cc
```
pksunkara: Wrong.

src/refract/ExpandVisitor.h
```diff
@@ -0,0 +1,54 @@
+//
+//  refract/IsExpandableVisitor.h
```
pksunkara: Wrong.

pksunkara:
> Sry, it probably lost while rebase, it was same comment as above?

@klokane Yes, my comment was regarding [this](https://github.com/apiaryio/drafter/pull/46#discussion_r32892521)

I have identified a few other minor things which need to be fixed. Can you address them?

PS: Please do not edit any commits before this comment for the sake of my sanity since I just rebased C++ Boutique over this. :smile:

src/SerializeAST.cc
```diff
@@ -754,10 +774,45 @@ bool IsElementResourceGroup(const Element& element)
     return element.element == Element::CategoryElement && element.category == Element::ResourceGroupCategory;
 }
 
+
+#if _WITH_REFRACT_
+typedef std::vector<const snowcrash::DataStructure*> DataStructures;
+
+void findNamedDataStructures(const snowcrash::Elements& elements, DataStructures& found) {
+    for (snowcrash::Elements::const_iterator i = elements.begin()
+       ; i != elements.end()
```
pksunkara: This is not the style we follow in our C++ repos.

```
for (snowcrash::Elements::const_iterator i = elements.begin();
     i != elements.end();
     ++i) {
```

src/SerializeAST.cc
```diff
@@ -754,10 +774,45 @@ bool IsElementResourceGroup(const Element& element)
...
+
 sos::Object drafter::WrapBlueprint(const Blueprint& blueprint)
 {
     sos::Object blueprintObject;
 
+#if _WITH_REFRACT_
+    DataStructures found;
+    findNamedDataStructures(blueprint.content.elements(), found);
+
+    for(DataStructures::const_iterator i = found.begin() ; i != found.end() ; ++i) {
```
pksunkara: `for (`

pksunkara: Also unnecessary spaces before semicolons.

drafter.gyp
```diff
@@ -41,17 +41,33 @@
...
         "src/refract/SerializeCompactVisitor.cc",
         "src/refract/SerializeVisitor.h",
         "src/refract/SerializeVisitor.cc",
         "src/refract/ComparableVisitor.h",
+        "src/refract/ComparableVisitor.cc",
+        "src/refract/TypeQueryVisitor.h",
+        "src/refract/TypeQueryVisitor.cc",
+        "src/refract/IsExpandableVisitor.h",
+        "src/refract/IsExpandableVisitor.cc",
+        "src/refract/ExpandVisitor.h",
```
pksunkara: By the way, type expansion is not refract. It should be in drafter.

klokane: Are you sure?
https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#extend-element-element

pksunkara: Basically, This `ExpandVisitor` file is about expanding the named types given other named types right? I think that should be in drafter.

The refract spec only defines the `extend` stuff.

klokane: And refract MSON namespace describe how to do expansion
https://github.com/refractproject/refract-spec/blob/master/namespaces/mson-namespace.md#expanded-element

I mean expansion is part of refract much more than serialization :)

pksunkara: I think a good rule we can use here is, can remove the folder refract from here and make it a dependency library of drafter/snowcrash without any changes to the code in refract?

If no, then we are mixing up files and code. I will leave it up to you.

klokane: Agree, in future it should be separated into "library". 
It is one of reason why it is in own directory

src/refract/ComparableVisitor.cc
```diff
@@ -0,0 +1,40 @@
...
+#include "Visitors.h"
+
+namespace refract
+{
+
+    ComparableVisitor::ComparableVisitor(const std::string& str, const int compare) : compare_to(str), result(false), compare(compare)
+        {
+        }
+
+        void ComparableVisitor::visit(const MemberElement& e)
```
pksunkara: All these functions are indented by 4 extra spaces.

src/RefractAST.cc
```diff
@@ -719,15 +670,30 @@ namespace drafter
...
+        return element;
+    }
+
+    sos::Object SerializeRefract(refract::IElement* element, const refract::Registry& registry) {
+
         if (!element) {
             return sos::Object();
         }
 
+        refract::ExpandVisitor expander(registry);
```
pksunkara: I am feeling really bad about doing the expansion while serializing. It is going to introduce a lot of duplicate code. Can we separate the expansion part out of the serialization?

klokane: I don't understand, where is duplicate code?
By this solution we 
- avoid another looping trough data structures.
- minimal changes in original code as possible.

pksunkara: The duplication happens when I add boutique. We will leave it as it is now. But we should come back to this later and see if this needs reorganization.

klokane: We can make refactoring later.
Now it is premature optimization.

src/SerializeAST.cc
klokane: Result:
I check status and in discussion with Pavan we found it is handled correctly.
In function `findNamedDataStructures()` just for **data structures** not for complete resource.

src/RefractAST.cc
zdne: @pksunkara Agree with @klokane please do not optimize before it is needed

drafter.gyp
zdne: > expansion is not refract

@pksunkara Expansion is essentially described in Refract. And MSON Refract brings additional clarification to it. I do not see problem with what @klokane did here.

zdne:
@pksunkara looking at your review it is not clear what is a blocking issue for this to be merged. Please provide more clarity and mark your styling comments as "nitpicks"!

pksunkara:
@klokane Good to go. Will merge this after tests pass.

The following things are still needed to be done. Please create a separate PR for them:

1. Backport tests from drafter.js
2. Expand arrays in MSON
3. The MSON expanded objects should not replace the original objects.

PR #48, opened 2015-06-23
Typo sins

klokane:
Addresed problems form #46

### DIFF
--- a/test/fixtures/mson-mixin-nonexistent.apib
+++ b/test/fixtures/mson-mixin-nonexistent.apib
@@ -3,10 +3,12 @@
 
 # Data Structures
 
-## Address (object)
-- city
-- street
-
 ## User (object)
+
+FIXME: result of parsing `Include XXX` in snowcrash is now error.
+Because of nenexisting Mixin
+In future it should be possible to parse it correctly to allow Mixin referencing other document
+
+## Properties
 - id
-- Include Address
+- Include XXX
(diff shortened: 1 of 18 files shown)


test/fixtures/mson-mixin-nonexistent.apib
```diff
@@ -9,4 +9,4 @@
 
 ## User (object)
 - id
-- Include Address
+- Include XXX
```
pksunkara: @klokane Snowcrash will give error when parsing this test case. It wouldn't print the AST.

klokane: Hmm, with older version of snowcrash it was working (I forgot update sobmodules)

In future it should work again beacuse Mixins should allow reference remote blueprints. All checking in about referencing probably will be moved into drafter

pksunkara: > with older version of snowcrash it was working

Yes, but I added this check during the bug-fixes.

> In future it should work again

Yes, but that's in the future.

klokane: Hmm interesting
this: 
```
# <API name>
<API description>

# Data Structures

## User (object)
- id
- Include XXX 
```

Will not parse `Include XXX`

While this:
```
# <API name>
<API description>

# Data Structures

## User (object)

comment 

## Properties
- id
- Include XXX
```
does. 

(both cases returns error)

pksunkara: When snowcrash returns an error, the AST it gives should be neglected. The AST should not even be tested when the parser has an error. This is why, the above 2 samples have different AST, because we are returning an error as soon as we see one.

klokane: All tests in `test/test-Refract.cc` calling `HandleFixtureTest()` with second parameter `false` returns error while parsing.
It is part of checking edge condition

PR #50, opened 2015-06-24
refract backport drafter.js tests

klokane:
Ready for review
Changes:
- backported testcases from drafter.js
- MSON array nested types expansion

DO NOT MERGE - it probably break backward compatibility with **Extend JSON** where is @pksunkara working on.

### DIFF
--- a/src/refract/ExpandVisitor.cc
+++ b/src/refract/ExpandVisitor.cc
@@ -223,9 +223,18 @@ namespace refract
     void ExpandVisitor::visit(const NumberElement& e) {}
     void ExpandVisitor::visit(const BooleanElement& e) {}
     
-    // FIXME: can be ArrayElement Expandable?
-    // probably if any of members will be Member|Object
-    void ExpandVisitor::visit(const ArrayElement& e) {}
+    void ExpandVisitor::visit(const ArrayElement& e) {
+        if (!Expandable(e)) {  // do we have some expandable members?
+            return;
+        }
+
+        ArrayElement* a = new ArrayElement;
+        for (ArrayElement::ValueType::const_iterator it = e.value.begin(); it != e.value.end(); ++it) {
+            a->push_back(ExpandOrClone(*it, registry));
+        }
+
+        result = a;
+    }
 
     IElement* ExpandVisitor::get() const {
         return result;
--- a/src/refract/ExpandVisitor.cc
+++ b/src/refract/ExpandVisitor.cc
@@ -83,6 +83,13 @@ namespace refract
                 o->push_back(clone);
             }
 
+            // FIXME: posible solution while referenced type is not found in regisry
+            // \see test/fixtures/mson-resource-unresolved-reference.apib
+            //
+            //if (o->value.empty()) {
+            //   o->meta["ref"] = IElement::Create(name);
+            //}
+
             return o;
         }
 
--- a/test/fixtures/mson-array-typed-content.json
+++ b/test/fixtures/mson-array-typed-content.json
@@ -119,7 +119,32 @@
                   "element": "array",
                   "content": [
                     {
-                      "element": "X"
+                      "element": "extend",
+                      "content": [
+                        {
+                          "element": "object",
+                          "meta": {
+                            "ref": "X"
+                          },
+                          "content": [
+                            {
+                              "element": "member",
+                              "content": {
+                                "key": {
+                                  "element": "string",
+                                  "content": "y"
+                                },
+                                "value": {
+                                  "element": "string"
+                                }
+                              }
+                            }
+                          ]
+                        },
+                        {
+                          "element": "object"
+                        }
+                      ]
                     }
                   ]
                 }
--- a/test/fixtures/mson-array-typed-content.json
+++ b/test/fixtures/mson-array-typed-content.json
@@ -1,0 +1,171 @@
+{
+  "_version": "3.0",
+  "metadata": [],
+  "name": "<API name>",
+  "description": "<API description>\n\n",
+  "element": "category",
+  "resourceGroups": [],
+  "content": [
+    {
+      "element": "category",
+      "content": [
+        {
+          "element": "object",
+          "meta": {
+            "id": "User"
+          },
+          "content": [
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "nontyped array"
+                },
+                "value": {
+                  "element": "array"
+                }
+              }
+            },
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "typed array"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
+                      "element": "number"
+                    },
+                    {
+                      "element": "string"
+                    },
+                    {
+                      "element": "boolean"
+                    }
+                  ]
+                }
+              }
+            },
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "typed with values"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
+                      "element": "string",
+                      "content": "1"
+                    },
+                    {
+                      "element": "string",
+                      "content": "2"
+                    },
+                    {
+                      "element": "string",
+                      "content": "3"
+                    },
+                    {
+                      "element": "number"
+                    },
+                    {
+                      "element": "string"
+                    }
+                  ]
+                }
+              }
+            },
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "strict type with values"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
+                      "element": "number",
+                      "content": 1
+                    },
+                    {
+                      "element": "number",
+                      "content": 2
+                    },
+                    {
+                      "element": "number",
+                      "content": 3
+                    }
+                  ]
+                }
+              }
+            },
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "typed with object"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
+                      "element": "X"
+                    }
+                  ]
+                }
+              }
+            },
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "typed with generic"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
+                      "element": "generic",
+                      "content": "T"
+                    }
+                  ]
+                }
+              }
+            }
+          ]
+        },
+        {
+          "element": "object",
+          "meta": {
+            "id": "X"
+          },
+          "content": [
+            {
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "y"
+                },
+                "value": {
+                  "element": "string"
+                }
+              }
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
--- a/test/fixtures/mson-resource-nested-mixin.json
+++ b/test/fixtures/mson-resource-nested-mixin.json
@@ -1,0 +1,266 @@
+{
+  "_version": "3.0",
+  "metadata": [],
+  "name": "",
+  "description": "",
+  "element": "category",
+  "resourceGroups": [
+    {
+      "name": "",
+      "description": "",
+      "resources": [
+        {
+          "element": "resource",
+          "name": "",
+          "description": "",
+          "uriTemplate": "/",
+          "model": {},
+          "parameters": [],
+          "actions": [
+            {
+              "name": "",
+              "description": "",
+              "method": "GET",
+              "parameters": [],
+              "attributes": {
+                "relation": "",
+                "uriTemplate": ""
+              },
+              "content": [],
+              "examples": [
+                {
+                  "name": "",
+                  "description": "",
+                  "requests": [],
+                  "responses": [
+                    {
+                      "name": "200",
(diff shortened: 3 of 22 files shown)


src/refract/ExpandVisitor.cc
```diff
@@ -83,6 +83,13 @@ namespace refract
                 o->push_back(clone);
             }
 
+            // FIXME: posible solution while referenced type is not found in regisry
+            // \see test/fixtures/mson-resource-unresolved-reference.apib
+            //
+            //if (o->value.empty()) {
+            //   o->meta["ref"] = IElement::Create(name);
+            //}
+
```
pksunkara: This will never happen since snowcrash would have returned an error.

test/fixtures/mson-array-typed-content.json
```diff
@@ -0,0 +1,196 @@
...
+              "element": "member",
+              "content": {
+                "key": {
+                  "element": "string",
+                  "content": "typed array"
+                },
+                "value": {
+                  "element": "array",
+                  "content": [
+                    {
```
pksunkara: We don't need to add the nested types in here. Only value members is enough.

test/fixtures/mson-resource-nested-mixin.json
```diff
@@ -0,0 +1,266 @@
...
+                                      "element": "member",
+                                      "content": {
+                                        "key": {
+                                          "element": "string",
+                                          "content": "c"
+                                        },
+                                        "value": {
+                                          "element": "extend",
+                                          "content": [
+                                            {
```
pksunkara: I think this example is wrong.

test/fixtures/mson-resource-unresolved-reference.apib
```diff
@@ -0,0 +1,11 @@
+# Polls [/]
+
+FIXME: what should be result of expansion for nonexistent structure?
```
pksunkara: Drafter shouldn't even think of trying to expand this example since snowcrash gives an error.

Once the parser (snowcrash/drafter) sees an error, it stops doing everything and returns an error to the user. Why are we testing the expansion here?

klokane: In this case snowcrash DOES NOT returns error

test/fixtures/mson-resource-nested-mixin.json
klokane: Why?

test/fixtures/mson-array-typed-content.json
klokane: There are NO members in this case, just types
` + typed array (array[number,string, boolean])`

If I dont add primitive members w/ null content, nested members info will be lost.

We discuss it yesterday w/ @zdne this is result of our discussion.

src/refract/ExpandVisitor.cc
klokane: No return in this case, just warning see below.

test/fixtures/mson-resource-unresolved-reference.apib
pksunkara: We need to fix that later :(

test/fixtures/mson-array-typed-content.json
pksunkara: Okay.

test/fixtures/mson-resource-nested-mixin.json
pksunkara: This is the object you have here:

```
{
  "element": "object",
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "b"
        },
        "value": {
          "element": "object",
          "content": [
            {
              "element": "member",
              "content": {
                "key": {
                  "element": "string",
                  "content": "c"
                },
                "value": {
                  "element": "extend",
                  "content": [
                    {
                      "element": "object",
                      "content": [
                        {
                          "element": "object",
                          "content": [
                            {
                              "element": "object",
                              "meta": {
                                "ref": "X"
                              },
                              "content": [
                                {
                                  "element": "member",
                                  "content": {
                                    "key": {
                                      "element": "string",
                                      "content": "id"
                                    },
                                    "value": {
                                      "element": "string",
                                      "content": "pavan"
                                    }
                                  }
                                }
                              ]
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              }
            }
          ]
        }
      }
    }
  ]
}
```

I think the mixin is nested too deeply. This is the object I am thinking of:

```
{
  "element": "object",
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "b"
        },
        "value": {
          "element": "object",
          "content": [
            {
              "element": "member",
              "content": {
                "key": {
                  "element": "string",
                  "content": "c"
                },
                "value": {
                  "element": "extend",
                  "content": [
                    {
                      "element": "object",
                      "content": [
                        {
                          "element": "object",
                          "meta": {
                            "ref": "X"
                          },
                          "content": [
                            {
                              "element": "member",
                              "content": {
                                "key": {
                                  "element": "string",
                                  "content": "id"
                                },
                                "value": {
                                  "element": "string",
                                  "content": "pavan"
                                }
                              }
                            }
                          ]
                        }
                      ]
                    }
                  ]
                }
              }
            }
          ]
        }
      }
    }
  ]
}
```

pksunkara: Looks like other mixin examples are similarly deep nested.

zdne: @pksunkara comments like this:

> I think this example is wrong.

are not really helpful. Please try to think of such a comment can be dealt with.

zdne: @pksunkara it should be possible to clearly communicate it with words even without the huge JSON blobs – something along lines "mixin elements are incorrectly nested" goes a long way. The time you spent thinking about what you want to say is well worth it. Both to formulate your thoughts for yourself and for your teammates.

test/fixtures/mson-resource-unresolved-reference.apib
zdne: > We need to fix that later :(

Please keep track of these.

zdne:
So what is the result of this review @pksunkara ? @klokane is it clear what needs to be done in order for this to pass the review?

pksunkara:
@zdne Can you look at the mixin example and decide if it is correct or wrong? I think it is nested a bit extra, but I think refract spec is okay with that. If it is correct, we can merge this PR.

zdne:
@pksunkara Which example? Please always provide links / context etc.

pksunkara:
@zdne Okay. Do you think [this](https://github.com/apiaryio/drafter/blob/shared/refract/test/fixtures/mson-mixin.apib) api blueprint's AST [here](https://github.com/apiaryio/drafter/blob/shared/refract/test/fixtures/mson-mixin.json) is correct? I think the mixin is incorrectly nested in [here](https://github.com/apiaryio/drafter/blob/shared/refract/test/fixtures/mson-mixin.json#L66).

zdne:
@pksunkara @klokane trying to answer your question: No, it does not look correct. The `extend` element is missing the `id` (of `User`) value. Also the `extend element should have two entries – one for (ex) `User` and the other for `Address`.

As a consequence – you won't be able to tell, looking at expanded refract, whether the elements were inherited or "mixed-in".

If I haven't made any typo this is what it should be™:

## Input MSON

```apib
## Address (object)
- city

## User (object)
- id
- Include (Address)
```

## Address Refract (JSON)

```json
{
  "element": "object",
  "meta": {
    "id": "Address"
  },
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "city"
        },
        "value": {
          "element": "string"
        }
      }
    }
  ]
}
```

## User Refract (JSON)

```json
{
  "element": "object",
  "meta": {
    "id": "User"
  },
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "id"
        },
        "value": {
          "element": "string"
        }
      }
    },
    {
      "element": "ref",
      "content": {
        "href": "Address",
        "path": "content"
      }
    }
  ]
}
```

## User Refract Expanded (JSON)

```json
{
  "element": "extend",
  "meta": {
    "id": "User"
  },
  "content": [
    {
      "element": "object",
      "content": [
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "id"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    },
    {
      "element": "object",
      "meta": {
        "ref": "Address"
      },
      "content": [
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "city"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    }
  ]
}
```

pksunkara:
@zdne Thanks for the explanation.

zdne:
@pksunkara sure :+1:

zdne:
> As a consequence – you won't be able to tell, looking at expanded refract, whether the elements were inherited or "mixed-in".

**Update:** You will be. Sort of. But when extended there is no difference between 

```apib
# Address (User)
```
and 

```apib
# Address (object)
- Include (User)
```

You can however achieve a different results with mixins thanks to ordering. In other words using regular inheritance – the base object should always be the first. Not so when mixing in.

zdne:
Not sure if we won't hit a wall here tho.

pksunkara:
@zdne A question I have regarding this, are we not going to preserve the original structure anywhere? If you look at the normal API Blueprint AST, you have `dataStructure` and `resolvedDataStructure` which contains the expanded `dataStructure`. Don't we need to do the same here?

klokane:
@zdne OK will fix it in new Pull Request in way you suggest.

pksunkara:
Since the "mixin fixes" need to be done on other test cases outside this PR, I am merging this PR and ask the "mixin fixes" to be a new PR.

klokane:
If there are no other problem except Mixin, from point of view it can be merged, so I remove label DNM

zdne:
> A question I have regarding this, are we not going to preserve the original structure anywhere? If you look at the normal API Blueprint AST, you have dataStructure and resolvedDataStructure which contains the expanded dataStructure. Don't we need to do the same here?

@pksunkara not sure if I understand your question correctly: is it 

1. How to reconstruct the same source document from a Refract tree

or 

2. Looking at expanded tree – how do I know what elements it was expanded from

?

zdne:
If the later then the `ref` in meta together with the `extend` element are the clues

pksunkara:
@zdne I am thinking about the former, "How to reconstruct same source document from refract API AST"

If you remember some of our earlier discussions, we wanted to represent the API Blueprint in the AST such that the API Blueprint can be reconstructed from it.

klokane:
@zdne working on impl,  I mean expanded version should contain ref one level "higher" (it should be part of first content) because in this case we are not able to hold order of members.

eg.
```
## User (object)
- id
- Include (Address)
- login
```

Becomes into:
```json
{
  "element": "extend",
  "meta": {
    "id": "User"
  },
  "content": [
    {
      "element": "object",
      "content": [
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "id"
            },
            "value": {
              "element": "string"
            }
          }
        },
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "login"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    },
    {
      "element": "object",
      "meta": {
        "ref": "Address"
      },
      "content": [
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "city"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    }
  ]
}
```

And while merge we don't know where to place `Address`
resp. it will be placed on end and original structure after merge will be transformed into

```
## User (object)
- id
- login
- Include (Address)
```

zdne:
@pksunkara for that the answer is source map. Refract is "normalized" the source map information is essentially the information about the graph of a function. That is without source map you should be able to create some source document from a Refract, with source map you should be able to create the exactly same source.

zdne:
@klokane hm good point  (so we have hit a wall). This is not good and we need to come up with a solution. Please try to formulate a problem and proposed solution.

pksunkara:
Regarding the above mixin problem, Why are we not expanding it in place?

zdne:
>  Why are we not expanding it in place?

What that means? Can you give example?

pksunkara:
Basically, after expanding the MSON, maybe we should get something like this:

```
{
  "element": "object",
  "meta": {
    "id": "User"
  },
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "id"
        },
        "value": {
          "element": "string"
        }
      }
    },
    {
      "element": "mixin",
      "meta": {
        "ref": "Address"
      },
      "content": [
        {
          "element": "member",
          "content": {
            "key": {
              "element": "string",
              "content": "city"
            },
            "value": {
              "element": "string"
            }
          }
        }
      ]
    }
  ]
}
```

Of course, I used the `mixin` keyword as a placeholder. But you get the idea when I mean expand it in place.

zdne:
@pksunkara oh so you want to introduce a new element to solve this?

pksunkara:
Maybe. I would like to go through MSON namespace to see if we can solve it any other way.

Also, this is another idea of expanding it in place:

```json
{
  "element": "object",
  "meta": {
    "id": "User"
  },
  "content": [
    {
      "element": "member",
      "content": {
        "key": {
          "element": "string",
          "content": "id"
        },
        "value": {
          "element": "string"
        }
      }
    },
    {
      "element": "member",
      "content": {
        "key": {
           "element": "string",
          "content": "city"
        },
        "value": {
          "element": "string"
        }
      }
    }
  ]
}

```

But I am afraid that the losing the reference might be a concern.

klokane:
@pksunkara https://gist.github.com/klokane/a09b33abb79fa5a14adb

PR #54, opened 2015-06-25
refract fix inheritance expansion

klokane:
This fix disable copying element name while expanding deeply inherited elements


klokane:
PR on wrong branch

PR #55, opened 2015-06-25
Refract - disable copy of element name while expnading inheritance

klokane:
This fix resolve bug with copying element name while expanding "deeply" inherited elements


pksunkara:
Do we need to rebase this?

klokane:
rebased

pksunkara:
Looks like we didn't need to rebase this :smile:

klokane:
Already rebased, wait for build and can merge

PR #61, opened 2015-07-06
Added link to AUR package for Arch Linux installations

Tenzer:
I made a package to install Drafter on Arch Linux, this adds a link to it in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ $ brew install --HEAD \
   https://raw.github.com/apiaryio/drafter/master/tools/homebrew/drafter.rb
 ```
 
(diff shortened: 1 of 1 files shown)


pksunkara:
Does the AUR package be automatically up-to-date whenever we release a version?

Tenzer:
In its current form it simply checks out the code from Git, compiles it and installs it. It's made so it doesn't rely on release versions at all. Updating of the package would be a matter of reinstalling the package in order to get it to build from the latest commit in Git. I doubt that most AUR package managers would update the package regularly, as there isn't really any easy way for them to know when there's updates available.

pksunkara:
Thanks for the PR! :smile:

PR #62, opened 2015-07-14
Refract - implement Ref Element expansion

klokane:
Expansion via `attributes.resolved` as defined in https://github.com/refractproject/refract-spec/blob/master/namespaces/mson-namespace.md#type-reference-ref-element

### DIFF
--- a/src/refract/ExpandVisitor.cc
+++ b/src/refract/ExpandVisitor.cc
@@ -10,8 +10,6 @@
 #include "Visitors.h"
 #include "Registry.h"
 
-#include <iostream>
-
 namespace refract
 {
 
@@ -179,7 +177,6 @@ namespace refract
 
             return ref;
         }
-
 
         template <typename T>
         IElement* ExpandReference(const T& e, const Registry& registry)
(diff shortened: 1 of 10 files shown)


src/refract/ExpandVisitor.cc
```diff
@@ -168,35 +162,52 @@ namespace refract
...
+
+            IElement* base = registry.find(e.element());
+            TypeQueryVisitor t;
+            if (base) {
+                base->content(t);
+            }
 
-                return expanded;
+            if (t.get() == TypeQueryVisitor::Array) {
+                return CreateReference<ArrayElement>(e, href->value, registry);
```
pksunkara: I didn't know you would do this. I thought all the `ref` objects would be `ObjectElement`.

pksunkara: But, this is okay with me.

klokane: Oh, you are right. It is not correct. It stays there from original implemententation.
It is working because of same interface ArrayElement and ObjectElement
Will fix

klokane: It is working correcty, because of original object "ref" is every time of type ObjectElement.
So it every time call `else` branch.

PR #66, opened 2015-07-27
Fix refract mson

klokane:
- fix support for samples and defaults on Named Types && Array
 - refactoring Mson(Property|Value)ToRefract


src/RefractAST.cc
```diff
@@ -194,47 +194,75 @@ namespace drafter
...
     struct ExtractTypeSection
     { // This will handle primitive elements
-        const mson::TypeSection& ts;
+        std::vector<typename T::ValueType> values;
+        TypeSectionData& data;
 
-        ExtractTypeSection(const mson::TypeSection& v) : ts(v)
-        {
-        }
+        template <typename U, bool dummy = true>
```
klokane: This `dummy` param is required as workaround for inner template specialization
will add explaining comment

src/RefractAST.cc
```diff
@@ -194,47 +194,75 @@ namespace drafter
...
+
+    struct TypeSectionData {
+        RefractElements defaults;
+        RefractElements samples;
+        std::vector<std::string> descriptions;
+    };
+
+    template <typename T>
     struct ExtractTypeSection
     { // This will handle primitive elements
```
pksunkara: The previous PR had a comment on this line saying that this comment needs to be removed. I haven't looked into it yet but wanted to mention it.

klokane: thx for remind

klokane:
Fixed

klokane:
@pksunkara what is status of review?

pksunkara:
Reviewed and lgtm. I just have one concern which we need to write test cases for after these PRs are merged.

PR #67, opened 2015-07-27
Boutique samples and defaults

klokane:


src/refract/RenderJSONVisitor.cc
```diff
@@ -11,6 +11,8 @@
 #include "sosJSON.h"
 #include <sstream>
 
+#include <iostream>
```
pksunkara: Need to be removed.

src/refract/RenderJSONVisitor.cc
```diff
@@ -117,13 +174,18 @@ namespace refract
 
     void RenderJSONVisitor::visit(const ArrayElement& e) {
         RenderJSONVisitor renderer(sos::Base::ArrayType);
-        std::vector<refract::IElement*>::const_iterator it;
+
+        const ArrayElement::ValueType* val = getValue<ArrayElement>(e);
+        if (!val) {
```
pksunkara: Need empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -117,13 +174,18 @@ namespace refract
...
+            return;
+        }
 
         if (e.element() == "extend") {
             renderer.isExtend = true;
         }
 
-        for (it = e.value.begin(); it != e.value.end(); ++it) {
+        ArrayElement::ValueType::const_iterator it;
+        for (it = val->begin(); it != val->end(); ++it) {
```
pksunkara: Please add empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -100,13 +152,18 @@ namespace refract
...
+            return;
+        }
 
         if (e.element() == "extend") {
             renderer.isExtend = true;
         }
 
-        for (it = e.value.begin(); it != e.value.end(); ++it) {
+        std::vector<refract::IElement*>::const_iterator it;
+        for (it = val->begin(); it != val->end(); ++it) {
```
pksunkara: Please add empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -134,16 +196,26 @@ namespace refract
 
     void RenderJSONVisitor::visit(const NullElement& e) {}
 
+
```
pksunkara: Why the extra line?

src/refract/RenderJSONVisitor.cc
```diff
@@ -134,16 +196,26 @@ namespace refract
...
     void RenderJSONVisitor::visit(const StringElement& e) {
-        assign(sos::String(e.value));
+        const StringElement::ValueType* v = getValue<StringElement>(e);
+        if (v) {
```
pksunkara: Please add empty line before this. And similarly below.

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
...
+        operator const R*() {
+
+            if (const T* d = getDefault(element)) {
+                return &d->value;
+            } 
+
+            if (const T* s = getSample(element)) {
+                return &s->value;
+            }
+
```
pksunkara: Why the extra line?

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
         }
     }
 
+    template<typename T>
+    const T* getDefault(const T& e) {
+        IElement::MemberElementCollection::const_iterator i = e.attributes.find("default");
+        if (i == e.attributes.end()) {
+            return NULL;
+        }
+        TypeQueryVisitor v;
```
pksunkara: Please add empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
         }
     }
 
+    template<typename T>
+    const T* getDefault(const T& e) {
+        IElement::MemberElementCollection::const_iterator i = e.attributes.find("default");
+        if (i == e.attributes.end()) {
```
pksunkara: Please add empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
...
+            return NULL;
+        }
+        TypeQueryVisitor v;
+        return TypeQueryVisitor::as<T>((*i)->value.second);
+    }
+
+    template<typename T>
+    const T* getSample(const T& e) {
+        IElement::MemberElementCollection::const_iterator i = e.attributes.find("samples");
+        if (i == e.attributes.end()) {
```
pksunkara: Please add empty line before this.

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
...
+
+        TypeQueryVisitor v;
+        ArrayElement* a = TypeQueryVisitor::as<ArrayElement>((*i)->value.second);
+
+        if (!a || a->value.empty()) {
+            return NULL;
+        }
+
+        return TypeQueryVisitor::as<T>(*(a->value.begin()));
+
```
pksunkara: Why the extra line?

src/refract/RenderJSONVisitor.cc
```diff
@@ -77,6 +79,56 @@ namespace refract
...
+    template<typename T, typename R = typename T::ValueType>
+    struct getValue {
+        const T& element;
+
+        getValue(const T& e) : element(e) {}
+
+        operator const R*() {
+
+            if (const T* d = getDefault(element)) {
```
pksunkara: @klokane Why are we giving more importance to `default` values? I think sample values should get more importance in rendering since .... well they are samples.

klokane: @pksunkara Yes this is for discussion

@zdne @danielgtaylor @kylef  what do you mean about priority? What order use to fill JSON?

Currently it is:
- default
- sample
- value

pksunkara: Since `value` is `sample`, I think the priority should be:

* sample
* default

pksunkara:
Review done. Side note: Please don't forget to rebase the subsequent PRs when the top level PR is force-pushed. Example, first two commits in this PR are from the old #66 commit list.

src/refract/RenderJSONVisitor.cc
klokane: As defined in MSON:
Value
```
- a: value
```

Sample
```
-a: *value*
```
```
-a: value (sample)
```
```
-a
    - sample
        - value
```
Default
```
-a: value (default)
```
```
-a
    - default
          - value
```

pksunkara: Then, my list stands.

klokane:
Typo addresed, rebased, priority of sources still not changed I would like to hear someone other opinion about it

pksunkara:
Reviewed until here.

Just to continue the previous discussion from [here](https://github.com/apiaryio/drafter/pull/67#discussion_r35628455), I think concept of **default** implies that it will be used only when there is no **value**. That nails down half the order. We just need to decide where sample is going to reside.

src/refract/RenderJSONVisitor.cc
zdne: Default is a value that is used when no other value is present. So it should be used only when no value or sample is specified.

pksunkara:
So, from @zdne's comment at [here](https://github.com/apiaryio/drafter/pull/67#discussion_r35670651), the final priority order is:

* value
* sample
* default

klokane:
ok will change

klokane:
@pksunkara @zdne
Hmm there is one problematic point in this concept while we fill firstly from **value** in objects.. 

eg. what is expected result form MSON?
```
- o1
    - s1
    - s2

    - Sample
        - s1: a
        - s2: b
```

I mean correct is 
```
{ "o" : {
  "s1": "a",
  "s2": "b",
}}
```

But if we fill as first instance from value, result will be:
```
{ "o" : {
  "s1": "",
  "s2": "",
}}
```
It is Because value of `o` is set as two empty *string elements* `s1` and `s2`

pksunkara:
> It is Because value of o is set as two empty string elements s1 and s2

For primitive elements, we are checking if the value is empty or not before using samples. Why can't we do that for this too?

pksunkara:
That said, I don't think it's a big issue since the user could have used the sample values during the definition itself. As Z said earlier, the **value** itself is a sample value.

klokane:
Reordered with comment about empty members

pksunkara:
Reviewed. Ready to be merged once #66 is marked as ready.

PR #69, opened 2015-07-29
Refract MSON - double newline between descrpition and multiline description

klokane:
…iption

### DIFF
--- a/test/fixtures/mson/mson-multiline-description.json
+++ b/test/fixtures/mson/mson-multiline-description.json
@@ -34,7 +34,7 @@
             {
               "element": "member",
               "meta": {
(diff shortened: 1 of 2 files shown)


test/fixtures/mson/mson-multiline-description.json
```diff
@@ -66,7 +66,7 @@
             {
               "element": "member",
               "meta": {
-                "description": "this should ...\nbe member of user\n"
+                "description": "this should ...\n\nbe member of user\n"
```
pksunkara: This should not have an extra newline AFAIK atleast looking at the markdown.

klokane: It is exactly same case as above

PR #70, opened 2015-07-30
Hotfix to handle throwed exceptions

klokane:
This is hotfix to handle exception.
Now program return safe from exception, but still there can be some non released memory.
Working on fix

### DIFF
--- a/src/cdrafter.cc
+++ b/src/cdrafter.cc
@@ -35,7 +35,13 @@ SC_API int drafter_c_parse(const char* source,
 
     if (result) {
         std::stringstream resultStream;
-        serializer.process(drafter::WrapResult(blueprint, options), resultStream);
+        try {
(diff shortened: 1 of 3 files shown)


src/cdrafter.cc
```diff
@@ -35,7 +35,13 @@ SC_API int drafter_c_parse(const char* source,
 
     if (result) {
         std::stringstream resultStream;
-        serializer.process(drafter::WrapResult(blueprint, options), resultStream);
+        try {
+            serializer.process(drafter::WrapResult(blueprint, options), resultStream);
+        } 
+        catch (std::exception e) {
+            std::cerr << "Error: " << e.what() << std::endl;
```
pksunkara: Instead of printing the error why are we not adding it to `blueprint.report.error.message`? Similarly below.

klokane: updated

PR #72, opened 2015-08-04
Added CLI interface for APIB Refract

pksunkara:
This PR adds a CLI option for selecting between APIB Refract and APIB AST. It also renames some files and prepares internal infrastructure ready for the conversion of AST to Refract. It also does some file header comment maintenance.


pksunkara:
@zdne Please review. Github doesn't handle diff properly when files are renamed. So, you might want to review commits individually. /cc @kylef @danielgtaylor

src/RefractAST.h
```diff
@@ -2,27 +2,19 @@
 //  RefractAST.h
 //  drafter
 //
-//  Created by Jiri Kratochvil on 18/05/15.
+//  Created by Jiri Kratochvil on 31/07/15.
```
danielgtaylor: Is there a reason this creation date is changing? Also, why is this in the file when you can find the info via git?

pksunkara: We generally use creation dates in C++ projects. I don't mind them since Xcode automatically makes them. This is a case where the date was wrong since Jirka copied it from another file and I forgot to comment on this in the previous PR.

danielgtaylor:
Just the minor comment above. Looks good otherwise :+1:

src/RefractAST.h
danielgtaylor: Ah, okay that makes sense.

zdne:
@pksunkara please can you elaborate on naming – looking at the PR renaming "refract AST" to "refract MSON" does not make any sense (neither name does).

In addition, and we have discussed this before, this PR is mixing too many things together – if you need to rename files why not to do it in separate PR from a PR that introduces **UNRELATED** changes to CLI interface.

test/test-RefractASTTest.cc
```diff
@@ -0,0 +1,8 @@
+#include "draftertest.h"
+
+using namespace draftertest;
+
+TEST_CASE("Dummy test to perpare infrastructure", "[refract][ast]")
```
zdne: This is definitely not a good practice best to avoid.

pksunkara:
Basically, the code has `RefractAST` files which were converting MSON AST to MSON Refract. Since we are going to converting APIB AST to APIB Refract, @klokane renamed `RefractAST` to `RefractMSON` and added new files with the name `RefractAST`.

zdne:
@pksunkara note there is no "APIB refract" and (now) "MSON refract" the later has been renamed to "Data Structure Namespace" the former is "API Namespace". Therefore "RefractMSON" is not really correct.

pksunkara:
Do you want me to rename them to `RefractAPI` and `RefractDataStructure`?

zdne:
"RefractAPI" is about "API namespace" no? Not sure how it is related to "Data Structure namespace"

pksunkara:
> "RefractAPI" is about "API namespace" no? Not sure how it is related to "Data Structure namespace"

I think we are having a misunderstanding. We need two separate files to convert the AST into refract. One for MSON and the other for Blueprint. Are you okay with the current naming? `RefractAST` and `RefractMSON`.

If not, please propose something.

zdne:
@pksunkara I am just saying there is no longer such a thing as MSON refract

zdne:
it has been renamed to "Data Structure refract"

pksunkara:
And I agree with that. Which is why I was proposing to rename the above two files to `RefractAPI` and `RefractDataStructure`. One for API namespace and the other for Data Structure namespace.

zdne:
> rename RefractAPI and RefractDataStructure

Go for it.

My bad, I have read "rename them to RefractAPI and RefractDataStructure?" as "rename them to RefractAPI to RefractDataStructure?"

pksunkara:
Updated.

zdne:
@pksunkara OK, and please next time make sure to split the PRs ... I understand this is easier for you but think about others and also possible reverts in future.

PR #73, opened 2015-08-05
Fix build on Windows

danielgtaylor:
This change explicitly includes the `iterator` header in the
files using `std::back_iterator`. In some versions of Microsoft
Visual Studio, this header is not automatically included and
it leads to compile errors. See:

http://stackoverflow.com/questions/2959234/vs2008-vs2010-leads-to-cryptic-stl-errors

https://ci.appveyor.com/project/Apiary/protagonist/build/1.0.107

### DIFF
--- a/src/RefractAST.cc
+++ b/src/RefractAST.cc
@@ -6,6 +6,8 @@
 //  Copyright (c) 2015 Apiary Inc. All rights reserved.
 //
 
+#include <iterator>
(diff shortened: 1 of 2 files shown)


danielgtaylor:
Proof that this works:

https://ci.appveyor.com/project/Apiary/protagonist/build/1.0.109

PR #75, opened 2015-08-10
Started on refract conversion

pksunkara:
This PR starts the refract conversion and expands more on the refract option. It also streamlines a bit of visitor application code.

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -98,45 +98,47 @@ namespace drafter {
         return element;
     }
 
+    template<typename T>
+    refract::IElement* ExtractParameter(const snowcrash::Parameter& parameter)
+    {
+        refract::IElement* element = NULL;
+
+        if (parameter.values.empty()) {
+            element = refract::IElement::Create(LiteralTo<T>(parameter.exampleValue));
+        }
+        else {
+            element = ParameterValuesToRefract<T>(parameter);
+        }
+
+        if (!parameter.defaultValue.empty()) {
+            element->attributes["default"] = refract::IElement::Create(LiteralTo<T>(parameter.defaultValue));
+        }
+
+        return element;
+    }
+
     refract::IElement* ParameterToRefract(const snowcrash::Parameter& parameter)
     {
         refract::MemberElement* element = new refract::MemberElement;
-        refract::IElement *value, *defaultValue;
+        refract::IElement *value = NULL;
 
-        // Parameter type
+        // Parameter type, exampleValue, defaultValue, values
         if (parameter.type == "boolean") {
-            value = refract::IElement::Create(LiteralTo<bool>(parameter.exampleValue));
-            defaultValue = refract::IElement::Create(LiteralTo<bool>(parameter.defaultValue));
+            value = ExtractParameter<bool>(parameter);
         }
         else if (parameter.type == "number") {
-            value = refract::IElement::Create(LiteralTo<double>(parameter.exampleValue));
-            defaultValue = refract::IElement::Create(LiteralTo<double>(parameter.defaultValue));
+            value = ExtractParameter<double>(parameter);
         }
         else {
-            value = refract::IElement::Create(parameter.exampleValue);
-            defaultValue = refract::IElement::Create(parameter.defaultValue);
-        }
-
-        // Values
-        if (!parameter.values.empty()) {
-            if (parameter.type == "boolean") {
-                value = ParameterValuesToRefract<bool>(parameter);
-            }
-            else if (parameter.type == "number") {
-                value = ParameterValuesToRefract<double>(parameter);
-            }
-            else {
-                value = ParameterValuesToRefract<std::string>(parameter);
-            }
-        }
-
-        // Default Value
-        if (!parameter.defaultValue.empty()) {
-            value->attributes["default"] = defaultValue;
+            value = ExtractParameter<std::string>(parameter);
         }
 
         element->set(refract::IElement::Create(parameter.name), value);
-        element->meta["description"] = refract::IElement::Create(parameter.description);
+
+        // Description
+        if (!parameter.description.empty()) {
+            element->meta["description"] = refract::IElement::Create(parameter.description);
+        }
 
         // Parameter use
         if (parameter.use == snowcrash::RequiredParameterUse || parameter.use == snowcrash::OptionalParameterUse) {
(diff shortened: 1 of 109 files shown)


src/RefractAPI.cc
```diff
@@ -13,11 +13,154 @@
...
+
+        ast->element("category");
+        ast->meta["classes"] = CreateArrayElement("api");
+        ast->meta["title"] = refract::IElement::Create(blueprint.name);
+
+        if (!blueprint.description.empty()) {
+            elements.push_back(CopyToRefract(blueprint.description));
+        }
+
+        // FIXME: Correct this according to spec
```
klokane: remove fixme

klokane: will do :)

src/RefractAPI.cc
```diff
@@ -13,11 +13,154 @@
...
+        if (!blueprint.metadata.empty()) {
+            ast->attributes["meta"] = MetadataCollectionToRefract(blueprint.metadata);
+        }
+
+        // Append set of elements to content
+        const snowcrash::Elements& scElements = blueprint.content.elements();
+        std::transform(scElements.begin(), scElements.end(), std::back_inserter(elements), ElementToRefract);
+
+        // Remove NULL elements
```
klokane: Refactoring - (repeated on line no. 115) -> extract to function

src/RefractAPI.cc
```diff
@@ -13,11 +13,154 @@
...
+            return ptr == NULL;
+        }
+
+    }
+
+    refract::IElement* MetadataToRefract(const snowcrash::Metadata metadata)
+    {
+        refract::MemberElement* element = new refract::MemberElement;
+
+        element->meta["classes"] = CreateArrayElement("user");
```
klokane: Move all const text into `namespace key` like in ` RefractDataStructure.cc`

pksunkara:
Pushed new commit.

src/RefractDataStructure.cc
```diff
@@ -719,6 +719,10 @@ namespace drafter
 
     refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure)
     {
+        if (dataStructure.empty()) {
+            return NULL;
+        }
+
```
pksunkara: @klokane The issue which we need to fix here is that we do not have an [API namespace DataStructure](https://github.com/refractproject/refract-spec/blob/968cf6042757a365115ce88a5ddab6912e0e4a1e/namespaces/api-description-namespace.md#data-structure-element) here. Our AST directly embeds the MSON Refract.

src/RefractAPI.cc
pksunkara: I will do this in a separate PR.

pksunkara:
Updated the PR. Please review & merge this first.

src/RefractAPI.cc
klokane: ok

src/RefractAPI.cc
```diff
@@ -7,17 +7,442 @@
...
+        }
+        else {
+            element->element("httpRequest");
+            element->attributes["method"] = refract::IElement::Create(method);
+
+            if (payload != NULL) {
+                element->attributes["title"] = refract::IElement::Create(payload->name);
+            }
+
+            // FIXME: Expand href
```
klokane: What does mean this?
It is for this release?

pksunkara: I don't think API Blueprint uses this. But I left a remark here so that I can ask the guys later.

klokane:
@pksunkara pls check memory, `valgrind` report some memory leaks, all comes from  `ParameterToRefract`.

Probably some unassigned allocation

src/RefractAPI.cc
klokane: Ok so please add some little bit descriptive info.

pksunkara:
Updated.

src/RefractAPI.cc
```diff
@@ -7,17 +7,450 @@
...
+        }
+        else {
+            value = refract::IElement::Create(parameter.exampleValue);
+            defaultValue = refract::IElement::Create(parameter.defaultValue);
+        }
+
+        // Values
+        if (!parameter.values.empty()) {
+            if (parameter.type == "boolean") {
+                value = ParameterValuesToRefract<bool>(parameter);
```
klokane: probably memory leak - there will be content of `value` lost if it was already set on line  108. dtto for all other assigns below.

src/RefractAPI.cc
```diff
@@ -7,17 +7,450 @@
...
+
+        element->set(content);
+
+        return element;
+    }
+
+    refract::IElement* ParameterToRefract(const snowcrash::Parameter& parameter)
+    {
+        refract::MemberElement* element = new refract::MemberElement;
+        refract::IElement *value, *defaultValue;
```
klokane: if declare pointee variables to be assigned later every time init them to NULL

src/RefractAPI.cc
```diff
@@ -7,17 +7,450 @@
...
+                value = ParameterValuesToRefract<std::string>(parameter);
+            }
+        }
+
+        // Default Value
+        if (!parameter.defaultValue.empty()) {
+            value->attributes["default"] = defaultValue;
+        }
+        else {
+            delete defaultValue;
```
klokane: it is better to do not allocate memory if it is required to delete it later in same function

pksunkara:
Fixed.

src/RefractAPI.cc
```diff
@@ -7,17 +7,452 @@
...
+            defaultValue = refract::IElement::Create(LiteralTo<double>(parameter.defaultValue));
+        }
+        else {
+            value = refract::IElement::Create(parameter.exampleValue);
+            defaultValue = refract::IElement::Create(parameter.defaultValue);
+        }
+
+        // Values
+        if (!parameter.values.empty()) {
+            delete value;
```
klokane: What if value is not set? You will delete NULL

pksunkara: That won't happen because of the if-else loop above.

klokane: In both cases you are checking `parameter.type` against literal value
What about to join them together?

Something like:
```
     if (parameter.type == "boolean") { 
            if (!parameter.values.empty()) { 
                value = ParameterValuesToRefract<bool>(parameter);
            } 
            else { 
               value = refract::IElement::Create(LiteralTo<bool>(parameter.exampleValue));
            } 
            if (!parameter.defaultValue.empty()) { 
                value->attributes["default"] = refract::IElement::Create(LiteralTo<bool>(parameter.defaultValue));
            } 

        } 
  else if (...) {...}
```

and to DRY you can move this part to template function

pksunkara:
Updated.

pksunkara:
Updated.

klokane:
:+1:

PR #84, opened 2015-08-18
Render attributes when converting AST to Refract

pksunkara:


src/Serialize.h
```diff
@@ -25,6 +31,15 @@ namespace drafter {
         UnknownASTType = -1
     };
 
+#ifdef _WITH_REFRACT_
+    /**
+     * Hold all **Named Types** converted to Refract Element
+     * Later use this registry for expanding element and rendering before serialization
+     */
+
+    static refract::Registry NamedTypesRegistry;
```
klokane: This is not good, 
- do not use global variables
- `static`  in header is typical reason of "duplicit symbols" while linking.

Do we really need use it in more files?

pksunkara: Yes, we need it in more than one file since we need to do boutique rendering for both APIB AST and APIB Refract. Not having static is giving me "duplicate symbols".

klokane: So we need another solution.
Will look at it and propose something

pksunkara:
@klokane Updated.

pksunkara:
Updated.

PR #85, opened 2015-08-19
Added parse result envelope

pksunkara:
I have looked into few other methods and decided on this finally. I have also renamed the only blueprint fixture in test fixtures folder.


src/SerializeResult.cc
```diff
@@ -40,29 +40,43 @@ static sos::Object WrapAnnotation(const snowcrash::SourceAnnotation& annotation)
     return object;
 }
 
+sos::Object drafter::WrapParseResult(const snowcrash::ParseResult<snowcrash::Blueprint>& blueprint,
```
pksunkara: I might want to reconsider this name to be `WrapResultRefract`. What do you think @klokane ?

PR #87, opened 2015-08-25
Make sure references in attributes gets expanded in APIB Refract

pksunkara:
The attributes were getting expanded in MSON Refract in APIB AST but not in APIB Refract. This PR fixes that bug.

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -42,14 +42,15 @@ namespace drafter {
     refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure)
     {
         refract::IElement* msonElement = MSONToRefract(dataStructure);
+        refract::IElement* msonExpanded = ExpandRefract(msonElement, GetNamedTypesRegistry());
 
-        if (!msonElement) {
+        if (!msonExpanded) {
             return NULL;
         }
 
         refract::ObjectElement* element = new refract::ObjectElement;
         element->element(SerializeKey::DataStructure);
-        element->push_back(msonElement);
+        element->push_back(msonExpanded);
 
         return element;
     }
--- a/features/fixtures/refract.json
+++ b/features/fixtures/refract.json
@@ -679,7 +679,37 @@
                           "element": "dataStructure",
                           "content": [
                             {
-                              "element": "<data structure name>"
+                              "element": "extend",
+                              "content": [
+                                {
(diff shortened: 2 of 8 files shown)


src/RefractAPI.cc
```diff
@@ -41,15 +41,18 @@ namespace drafter {
 
     refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure)
     {
-        refract::IElement* msonElement = MSONToRefract(dataStructure);
+        refract::IElement *msonElement = NULL, *msonExpanded = NULL;
```
zdne: This line is needed. Furthermore please do declare variables on two lines without using the comma operator.

src/RefractAPI.cc
```diff
@@ -41,15 +41,18 @@ namespace drafter {
...
 
-        if (!msonElement) {
+        msonElement = MSONToRefract(dataStructure);
```
zdne: See my comment above

zdne:
@pksunkara 

> The attributes were getting expanded in MSON Refract in APIB AST but not in APIB Refract

Were expanded when? Please describe the issue more thoroughly, also is there a GH issue for this?  I understand this seems "more work" for you but explaining the problem is helpful even for you.

I do not like how did you moved `ExpandRefract` into `DataStructureToRefract` essentially changing the output of this function everywhere it was used. Maybe it should be called `DataStructureToExpandedRefract` then? 

What you have done is that `ElementToRefract` now always return expanded `Element::DataStructureElement` elements. Is this intended? I would much rather see this clearly separated as it was

"Side note": Also I absolutely dislike the fact that you do not use comments on the functions what so ever. This is a very bad practice not helping others to understand your code.

zdne:
Note I am also going to discuss the lack of comments with @klokane once he is back.

pksunkara:
> What you have done is that ElementToRefract now always return expanded Element::DataStructureElement elements. Is this intended? I would much rather see this clearly separated as it was

Apparently, this is a miscommunication and we are currently resolving this in Slack since github is down. Will update here later.

pksunkara:
Updated.

pksunkara:
Appveyor is failing due to github issues :disappointed:

zdne:
For the record: We have discussed this offline – the decision is to remove expansion from the result. This will be implemented after this PR is merged.

PR #88, opened 2015-08-25
Removed Data Structure expansion from APIB Refract

pksunkara:
This PR does a few things:

1. Removes Data Structure expansion from APIB Refract
2. Adds an option for `drafter::WrapBlueprint` for data structure expansion so that the MSON refract tests can use it to check the intermediate step of expanding refract for rendering.
3. Change the above mentioned tests to use the above mentioned option.

### DIFF
--- a/features/fixtures/ast.json
+++ b/features/fixtures/ast.json
@@ -243,37 +243,7 @@
                           "element": "dataStructure",
                           "content": [
                             {
-                              "element": "extend",
-                              "content": [
-                                {
-                                  "element": "object",
-                                  "meta": {
-                                    "description": "<data structure description>\n\n",
-                                    "ref": "<data structure name>"
-                                  },
-                                  "content": [
-                                    {
-                                      "element": "member",
-                                      "meta": {
-                                        "description": "<data structure property description>"
-                                      },
-                                      "content": {
-                                        "key": {
-                                          "element": "string",
-                                          "content": "<data structure property name>"
-                                        },
-                                        "value": {
-                                          "element": "string",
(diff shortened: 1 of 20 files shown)


zdne:
Looks good. Before merging I just want to make sure there are test for the expansion itself. @pksunkara can you please point me to it? 

Also since you are introducing the `ExpandMSON` option and we have the tests for situation where the expansion is enabled I would like to retain that (test). 

Can you please keep at least one of the test with expanded refract as it was using the  `ExpandMSON` (yes, that means you have to expose it). 

The reason I am asking is because you are keeping it in the code as a hypothetical execution branch and hence it should be tested. Or removed.

Also some comments around that in the code would not hurt...

pksunkara:
Basically, the folder `test/fixtures/mson` test the expansion, `test/fixtures/render` test the rendering and `test/fixtures/api` test the API Refract.

In this PR, I remove the MSON expansion from `test/fixtures/render` and `test/fixtures/api`. Where as `test/fixtures/mson` does not have any changes.

> Also since you are introducing the ExpandMSON option and we have the tests for situation where the expansion is enabled I would like to retain that (test).

> Can you please keep at least one of the test with expanded refract as it was using the ExpandMSON (yes, that means you have to expose it).

They are being retained by making changes to [this file](https://github.com/apiaryio/drafter/pull/88/files#diff-188b46cfb00d67bc72d934f63cf1924a). You can't see it in this PR because there are no changes to those tests since they are being retained.

zdne:
@pksunkara thanks!

PR #92, opened 2015-09-02
Don't exact match for content-type in rendering

pksunkara:
This PR makes sure that when rendering we don't try to match the content-types exactly. This PR also fixes a small case where a value is not passed by reference.

### DIFF
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -9,14 +9,18 @@
 #include "Render.h"
 #include "RefractDataStructure.h"
 #include "BlueprintUtility.h"
+#include "RegexMatch.h"
 
 using namespace snowcrash;
 
 namespace drafter {
 
+    // JSON content-type regex
+    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json";
+
     RenderFormat findRenderFormat(const std::string& contentType) {
 
-        if (contentType.find("application/json") != std::string::npos) {
+        if (RegexMatch(contentType, JSONRegex)) {
             return JSONRenderFormat;
         }
 
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -14,9 +14,9 @@ using namespace snowcrash;
 
 namespace drafter {
 
-    RenderFormat findRenderFormat(std::string contentType) {
+    RenderFormat findRenderFormat(const std::string& contentType) {
 
-        if (contentType == "application/json") {
+        if (contentType.find("application/json") != std::string::npos) {
             return JSONRenderFormat;
         }
 
--- a/test/fixtures/render/content-type.apib
+++ b/test/fixtures/render/content-type.apib
@@ -2,3 +2,13 @@
 + Response 200 (application/json; charset=utf-8)
     + Attributes
         + user: pksunkara
+
+# GET /hal
++ Response 200 (application/hal+json)
+    + Attributes
+        + user: pksunkara
+
+# GET /typo
++ Response 200 (application/haljson)
+    + Attributes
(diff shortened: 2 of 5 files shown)


src/Render.cc
```diff
@@ -14,9 +14,9 @@ using namespace snowcrash;
 
 namespace drafter {
 
-    RenderFormat findRenderFormat(std::string contentType) {
+    RenderFormat findRenderFormat(std::string& contentType) {
```
zdne: Please read http://duramecho.com/ComputerInformation/WhyHowCppConst.html

src/Render.cc
```diff
@@ -14,9 +14,9 @@ using namespace snowcrash;
...
 
-        if (contentType == "application/json") {
+        if (contentType.find("application/json") != std::string::npos) {
```
zdne: since you have got into this I would consider supporting - JSON-based formats (e.g. `application/hal+json` or `application/vnd.github.v3+json` etc.)

pksunkara: I don't think we should do this right now.

Anyway, we might want to do some custom rendering for `application/hal+json`

pksunkara:
Updated.

src/Render.cc
zdne: > I don't think we should do this right now. 

Please elaborate.

also 

> contentType.find("application/json") != std::string::npos 

isn't quite what you want – if I am not mistaken it will match things like `whatchildnot application/json` – you want to `application/json` only at the start of the string.

zdne:
@pksunkara added comment https://github.com/apiaryio/drafter/pull/92#discussion_r38537543

pksunkara:
Updated.

test/fixtures/render/content-type.apib
```diff
@@ -0,0 +1,14 @@
...
+    + Attributes
+        + user: pksunkara
+
+# GET /hal
++ Response 200 (application/hal+json)
+    + Attributes
+        + user: pksunkara
+
+# GET /typo
++ Response 200 (application/haljson)
```
zdne: Is this a real media type? Or do we know of any media type that is `application/whateverjson` ? My point is shouldn't we be checking for either `application/json` or `application/whatever+json` (I know the regex would be more complex 😜 )

pksunkara: As you see the regex, I am only checking for the aforementioned patterns. This weird example is there to show that this content type will not generate a json render. Please check the output below

zdne: Ok. Can you please fix the regex so it works with `application/json` or `+json` correctly?

pksunkara: Do you mean we should support anything ending with `+json`? Currently I think the regex supports `application/*+json`

zdne: `application/haljson` should not be supported

pksunkara: Yes, `application/haljson` is not supported by the regex.

zdne: Never mind. The regex is OK – I was confused with the example.

test/fixtures/render/content-type.json
```diff
@@ -0,0 +1,430 @@
...
+                  "description": "",
+                  "requests": [],
+                  "responses": [
+                    {
+                      "name": "200",
+                      "description": "",
+                      "headers": [
+                        {
+                          "name": "Content-Type",
+                          "value": "application/haljson"
```
kylef: HAL mime-type is actually `application/hal+json`.

PR #93, opened 2015-09-03
Add Drafter-php as binding for the PHP language

hendrikmaus:
I recently created [Drafter-php](https://github.com/hendrikmaus/drafter-php) as wrapper around the Drafter command line tool.

The package can be installed using composer.
Drafter command line tool has to be installed aside to it. The readme contains some basic instructions on how to do that with composer, but first of all references the Drafter repositories installation guides.

If you consider this a binding, this pull request would add it to the readme.

I am looking forward to your feedback.

Cheers,
Hendrik

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,7 @@ Drafter bindings in other languages:
 
 - [Protagonist](https://github.com/apiaryio/protagonist) (Node.js)
 - [RedSnow](https://github.com/apiaryio/redsnow) (Ruby)
+- [Drafter-php](https://github.com/hendrikmaus/drafter-php) (PHP)
 
 
 ## Contribute
--- a/README.md
+++ b/README.md
@@ -135,6 +135,8 @@ Drafter bindings in other languages:
 
 - [Protagonist](https://github.com/apiaryio/protagonist) (Node.js)
 - [RedSnow](https://github.com/apiaryio/redsnow) (Ruby)
+
+### CLI Wrapper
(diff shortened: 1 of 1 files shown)


pksunkara:
@hendrikmaus We love this, but this is just a wrapper around the CLI tool. This does not use the native c++/c-interface like the other bindings (protagonist, redsnow) publicised in README does.

I am a little hesitant to merge this. Maybe this should go to some kind of wiki page but not the README.

zdne:
Thanks @hendrikmaus – this is awesome! 

@pksunkara 

> c++/c-interface like the other bindings (protagonist, redsnow) publicised in README does

I believe that is OK. I would still add this to the README. If you really want to differentiate that I would add it to a new section of README called something like "CLI Binding"  but I think that just makes it harder to think about. 

So for me it is OK as it is now.

hendrikmaus:
Glad you like it, guys.
I investigated a bit, but I did not find a way to natively bind the c or c++ interface in a way I would be comfortable to put out and maintain. It would always involve building a php extension either in plain vanilla c, php-cpp, long broken ffi (pecl) or swig. a simple wrapper around the cli tool made the most sense for me.

pksunkara:
@hendrikmaus Can you make a `### CLI Wrapper` subheader under the `Bindings` header and list this under there?

I would like to differentiate between native bindings and CLI wrappers as some people would like to use different things based on their use cases. Thanks.

hendrikmaus:
@pksunkara added the subheader, cheers

pksunkara:
Thanks :smile:

PR #108, opened 2015-09-10
Addition of `nullable` type attribute

obihann:
Based on the pull request apiaryio/mson/pull/35 adding the `null` primitive type I have additionally added support for this in drafter.

#### 2.1.1 Primitive Types
Applies to basic data structure and type definitions. _[Types][]_ that are sub-typed from _Primitive Types_ MUST NOT
contain a _[Member Type Group][]_ or _[Nested Member Types][]_.

- `null`

    Specifies a null type.

- `boolean`

    Specifies a type with allowed values of `true` or `false`.

- `string`

    Specifies any string.

- `number`

    Specifies any number.


obihann:
When I run `make test` locally everything passes, I assume that for whatever reason this is not being built with the latest version of snowcrash.

pksunkara:
1. The commit history is very ugly with all kinds of commits. Please branch off `shared/refract` and use that new branch without including the master branch
2. Please remove `nullable` from parameters
3. You need to add rendering support for `nullable`.
  * Please create a fixture test first in `test/fixtures/render` which tests for all kind of cases using `nullable`

obihann:
I'll close this and do a clean one with the requested changes.

PR #116, opened 2015-09-11
Adding `nullable` attribute

### LINKED ISSUE #106: Add nullable type attribute
Based on apiaryio/mson#38 the nullable type attribute should be added. Examples of used and rendering can be found in the pull request.

If I understand this project correctly *drafter* is essentially a wrapper for *[apiaryio/snowcrash](https://github.com/apiaryio/snowcrash)*, which means this addition will not actually be part of drafter but *apiaryio/snowcrash* itself. Once the change is made in *[apiaryio/snowcrash](https://github.com/apiaryio/snowcrash)* then *drafter* will simply be updated to include the latest version of *[apiaryio/snowcrash](https://github.com/apiaryio/snowcrash)*... Am I close here?

if so then apiary/snowcrash#347 will be related to this as that issue is for the implementation of `nullable` in *[apiaryio/snowcrash](https://github.com/apiaryio/snowcrash)*.

obihann:
+ adding `nullable` type attribute based on apiaryio/mson#38 this should resolve #106
+ updating `ext/snowcrash` to `f9942fefa531626d3406effdad5d1b35ed4f0120` which was the addition of `nullable` to snowcrash in apiaryio/snowcrash#349

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -41,6 +41,9 @@ namespace drafter {
         if (ta & mson::FixedTypeAttribute) {
             attr->push_back(refract::IElement::Create(SerializeKey::Fixed));
         }
+        if (ta & mson::NullableTypeAttribute) {
+            attr->push_back(refract::IElement::Create(SerializeKey::Nullable));
+        }
 
         if (attr->value.empty()) {
             delete attr;
--- a/src/Serialize.h
+++ b/src/Serialize.h
@@ -71,6 +71,7 @@ namespace drafter {
         static const std::string Type;
         static const std::string Required;
         static const std::string Default;
+        static const std::string Nullable;
         static const std::string Example;
         static const std::string Values;
 
--- a/src/SerializeAST.cc
+++ b/src/SerializeAST.cc
@@ -161,6 +161,9 @@ sos::Array WrapTypeAttributes(const mson::TypeAttributes& typeAttributes)
     else if (typeAttributes & mson::DefaultTypeAttribute) {
         typeAttributesArray.push(sos::String("default"));
     }
+    else if (typeAttributes & mson::NullableTypeAttribute) {
+        typeAttributesArray.push(sos::String("nullable"));
+    }
     else if (typeAttributes & mson::SampleTypeAttribute) {
         typeAttributesArray.push(sos::String("sample"));
     }
--- a/src/refract/RenderJSONVisitor.cc
+++ b/src/refract/RenderJSONVisitor.cc
@@ -63,11 +63,42 @@ namespace refract
         e.content(*this);
     }
 
+    template<typename T>
+    bool isNullable(const T& e) {
+        IElement::MemberElementCollection::const_iterator ta = e.attributes.find("typeAttributes");
+        
+        if (ta == e.attributes.end()) {
+            return false;
+        }
+
+        ArrayElement* attrs = TypeQueryVisitor::as<ArrayElement>((*ta)->value.second);
+
+        if (!attrs) {
+            return false;
+        }
+
+        for (ArrayElement::ValueType::const_iterator it = attrs->value.begin() ; it != attrs->value.end() ; ++it ) {
+            StringElement* attr = TypeQueryVisitor::as<StringElement>(*it);
+            if (!attr) {
+                continue;
+            }
+            if (attr->value == "nullable") {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
     void RenderJSONVisitor::visit(const MemberElement& e) {
         RenderJSONVisitor renderer;
 
         if (e.value.second) {
-            renderer.visit(*e.value.second);
+            if (isNullable(e) && e.value.second->empty()) {
+                renderer.result = sos::Null();
+            } else {
+                renderer.visit(*e.value.second);
+            }
         }
 
         if (StringElement* str = TypeQueryVisitor::as<StringElement>(e.value.first)) {
@@ -136,6 +167,10 @@ namespace refract
                 return &d->value;
             }
 
+            if (element.empty() && isNullable(element)) {
+                return NULL;
+            }
+            
             return &element.value;
         }
     };
--- a/test/fixtures/render/nullable.apib
+++ b/test/fixtures/render/nullable.apib
@@ -1,0 +1,11 @@
+# GET /
++ Response 200 (application/json)
+    + Attributes
+        + s1: abc
+        + s2: def (required)
+        + s3 (optional)
+        + s4 (nullable)
+        + s5 (required, nullable)
+        + s6 (optional, nullable)
+        + s7: xyz (required, nullable)
+        + s8: xyz (optional, nullable)
--- a/test/fixtures/render/nullable.json
+++ b/test/fixtures/render/nullable.json
@@ -1,0 +1,410 @@
+{
+  "_version": "4.0",
+  "metadata": [],
+  "name": "",
+  "description": "",
+  "element": "category",
+  "resourceGroups": [
+    {
+      "name": "",
+      "description": "",
+      "resources": [
+        {
+          "element": "resource",
+          "name": "",
+          "description": "",
+          "uriTemplate": "/",
+          "model": {},
+          "parameters": [],
+          "actions": [
+            {
+              "name": "",
+              "description": "",
+              "method": "GET",
+              "parameters": [],
+              "attributes": {
+                "relation": "",
+                "uriTemplate": ""
+              },
+              "content": [],
+              "examples": [
+                {
+                  "name": "",
+                  "description": "",
+                  "requests": [],
+                  "responses": [
+                    {
+                      "name": "200",
+                      "description": "",
+                      "headers": [
+                        {
+                          "name": "Content-Type",
+                          "value": "application/json"
+                        }
+                      ],
+                      "body": "{\n  \"s1\": \"abc\",\n  \"s2\": \"def\",\n  \"s3\": \"\",\n  \"s4\": null,\n  \"s5\": null,\n  \"s6\": null,\n  \"s7\": \"xyz\",\n  \"s8\": \"xyz\"\n}",
+                      "schema": "",
+                      "content": [
+                        {
+                          "element": "dataStructure",
+                          "content": [
+                            {
+                              "element": "object",
+                              "content": [
+                                {
+                                  "element": "member",
+                                  "content": {
+                                    "key": {
+                                      "element": "string",
+                                      "content": "s1"
+                                    },
+                                    "value": {
+                                      "element": "string",
+                                      "content": "abc"
+                                    }
+                                  }
+                                },
+                                {
+                                  "element": "member",
+                                  "attributes": {
+                                    "typeAttributes": [
+                                      "required"
+                                    ]
+                                  },
+                                  "content": {
+                                    "key": {
+                                      "element": "string",
+                                      "content": "s2"
+                                    },
+                                    "value": {
+                                      "element": "string",
+                                      "content": "def"
+                                    }
+                                  }
+                                },
+                                {
+                                  "element": "member",
+                                  "attributes": {
+                                    "typeAttributes": [
+                                      "optional"
+                                    ]
+                                  },
+                                  "content": {
+                                    "key": {
+                                      "element": "string",
+                                      "content": "s3"
+                                    },
+                                    "value": {
+                                      "element": "string"
+                                    }
+                                  }
+                                },
+                                {
+                                  "element": "member",
+                                  "attributes": {
+                                    "typeAttributes": [
+                                      "nullable"
+                                    ]
+                                  },
+                                  "content": {
+                                    "key": {
+                                      "element": "string",
+                                      "content": "s4"
+                                    },
+                                    "value": {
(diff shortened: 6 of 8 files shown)


obihann:
Please note I'm creating this now to ensure the rebase went smooth, however I need to add the tests before it can be merged.

src/RefractDataStructure.cc
```diff
@@ -158,7 +158,7 @@ namespace drafter {
         //std::transform(elements.begin(), elements.end(),
         //               std::back_inserter(result),
         //               std::bind2nd(std::ptr_fun(MsonElementToRefract), nestedTypeName));
-
+
```
kylef: @obihann Would you be able to remove all the trailing white-space you've added please.

obihann: Defiantly, usually I'm good about those, sorry.

obihann:
@pksunkara I'm having some difficulty with my test cases, would you be able to take a look and point me in the correct direction? Please and thank you.

src/RefractDataStructure.cc
```diff
@@ -86,9 +89,9 @@ namespace drafter {
             }
 
             if (sample) {
-                refract::ArrayElement* samples = new refract::ArrayElement;
-                samples->push_back(refract::IElement::Create(LiteralTo<typename E::ValueType>(literal)));
-                element->attributes[SerializeKey::Samples] = samples;
+                refract::ArrayElement* a = new refract::ArrayElement;
+                a->push_back(refract::IElement::Create(LiteralTo<typename E::ValueType>(literal)));
+                element->attributes[SerializeKey::Samples] = a;
```
pksunkara: This is not your change.

src/RefractDataStructure.cc
```diff
@@ -233,9 +236,9 @@ namespace drafter {
         template<typename V>
         struct Store {
             void operator()(RefractElements& elements, const V& v) {
-                T* element = new T;
-                element->set(v);
-                elements.push_back(element);
+                T* e = new T;
+                e->set(v);
+                elements.push_back(e);
```
pksunkara: This is not your change.

src/RefractDataStructure.cc
```diff
@@ -41,6 +41,9 @@ namespace drafter {
         if (ta & mson::FixedTypeAttribute) {
             attr->push_back(refract::IElement::Create(SerializeKey::Fixed));
         }
+        if (ta & mson::NullableTypeAttribute) {
+            attr->push_back(refract::IElement::Create(SerializeKey::Nullable));
+        }
```
pksunkara: Please remove the changes from the whole file except this.

src/Serialize.h
```diff
@@ -150,7 +151,7 @@ namespace drafter {
     };
 
     /**
-     * \brief functor pattern to translate _collection_ into sos::Array on serialization
+     * \brief functor pattern to translate _collection_ into sos::Array on serialization
```
pksunkara: Please remove this.

src/SerializeAST.cc
```diff
@@ -789,8 +792,7 @@ bool IsElementResourceGroup(const Element& element)
 #if _WITH_REFRACT_
 typedef std::vector<const snowcrash::DataStructure*> DataStructures;
 
-void findNamedTypes(const snowcrash::Elements& elements, DataStructures& found)
-{
+void findNamedTypes(const snowcrash::Elements& elements, DataStructures& found) {
```
pksunkara: Please revert this.

pksunkara:
Please revert the changes I mentioned which are probably because of the rebase.

1. You don't have the correct test case.
2. You didn't add the render logic. You just added the serialization stuff.

test/fixtures/render/nullable.apib
```diff
@@ -0,0 +1,6 @@
+# GET /
+
++ Response 200 (application/json)
+    + Attributes
+        + s1: obihann (string, required) 
+        + s2: null (string, nullable)
```
pksunkara: Please change the Attributes in the test case to this:

```
+ s1: abc
+ s2: def (required)
+ s3 (optional)
+ s4 (nullable)
+ s5 (required, nullable)
+ s6 (optional, nullable)
```

Which should render the following JSON for body

```
{
  "s1": "abc",
  "s1": "def",
  "s4": null,
  "s5": null,
  "s6": null
}
```

pksunkara:
After changing the test case as I mentioned above, please add the rendering logic in `src/refract/RenderJSONVisitor.cc`

src/RefractDataStructure.cc
obihann: @pksunkara this is likely because I was branched off shared/refract then rebased to develop, I'll clean this all up now though.

obihann:
@pksunkara @kylef I believe I have all the bad code cleaned up, carrying on with the rest of the issues now :+1:

src/refract/RenderJSONVisitor.cc
```diff
@@ -106,6 +107,16 @@ namespace refract
         return TypeQueryVisitor::as<T>(*(a->value.begin()));
     }
 
+    template<typename T>
+    const T* getNullable(const T& e) {
+        IElement::MemberElementCollection::const_iterator ta = e.attributes.find("typeAttributes");
+        
+        // from within ta we have to look for the 'nullable' attribute
```
obihann: @pksunkara this is where I was trying to implement something like ```ta->find("nullable")```

src/refract/RenderJSONVisitor.cc
```diff
@@ -106,6 +107,16 @@ namespace refract
...
+        
+        // this should return the new value if nullable
+        return NULL;
```
obihann: This might be tricky since the other functions return null or the value, however for us the value is null. Perhaps I will return true and then in getValue return null when this returns true?

src/RefractDataStructure.cc
```diff
@@ -592,6 +595,7 @@ namespace drafter {
         return !value.valueDefinition.typeDefinition.typeSpecification.name.symbol.literal.empty();
     }
 
+
```
pksunkara: Please revert this.

src/RefractDataStructure.cc
```diff
@@ -631,7 +635,7 @@ namespace drafter {
                 return Trait::template Invoke<refract::ObjectElement>(input, defaultNestedType);
 
             case mson::UndefinedTypeName:
-            {
+
```
pksunkara: Please revert this.

src/RefractDataStructure.cc
```diff
@@ -652,7 +656,7 @@ namespace drafter {
                    default:
                        throw std::logic_error("Nested complex types are not Implemented");
                 }
-            }
+
```
pksunkara: Please revert this.

src/RefractDataStructure.cc
pksunkara: The changes below this section in this file haven't been cleaned.

src/refract/SerializeVisitor.cc
```diff
@@ -22,6 +23,17 @@ namespace refract
             for (iterator it = collection.begin(); it != collection.end(); ++it) {
                 SerializeCompactVisitor s;
                 s.visit(*(*it));
+
```
obihann: just some console logs so I can identify the structure, this should be removed from pull request

src/refract/RenderJSONVisitor.cc
```diff
@@ -102,10 +105,38 @@ namespace refract
...
+        {
+            std::cerr << s.key() << std::endl;
+            
+            for(int x = 0; x <= s.value().array().size(); x++)
+            {
+                std::cerr << s.value().array()[0].str << std::endl;
+            }
+        }
+        
+        return NULL;
```
obihann: @pksunkara what do you suggest we return when we find `nullable`, NULL almost makes sense but obviously could be confusing.

src/refract/RenderJSONVisitor.cc
```diff
@@ -102,10 +105,38 @@ namespace refract
...
     }
 
+    template<typename T>
+    const T* getNullable(const T& e) {
+        IElement::MemberElementCollection::const_iterator ta = e.attributes.find("typeAttributes");
+        
+        if (ta == e.attributes.end()) {
+            return NULL;
+        }
+
```
obihann: Still trying to work through the serialized attributes to find "nullable" when it exists.

src/refract/RenderJSONVisitor.cc
pksunkara: This will be `sos::Null()`

obihann: Thanks!

src/refract/RenderJSONVisitor.cc
pksunkara: `ta` is a `MemberElement`. But the issue here is that apparently the code is not coming here. I think `snowcrash` still has a problem since it keeps giving an error saying that ``unable to find `nullable` type``

obihann: @pksunkara I'm still pretty stuck here, I have a list of notes in gitter but the short version is that I believe when `getNullable` hits the string element representing our nullable element, it does not contain any attributes whatsoever. I came to this conclusion by adding a breakpoint and reading the value of e every time `getNullable` is called. I have some screenshots of the data in gitter that should help show what I found.

Do you think we can still find the attributes for string elements or will we need to update the serialization of string elements?

src/RefractDataStructure.cc
pksunkara: Please remove all the changes below this line in this file. They are here because of a bad rebase.

test/fixtures/render/nullable.json
```diff
@@ -0,0 +1,410 @@
...
+                    {
+                      "name": "200",
+                      "description": "",
+                      "headers": [
+                        {
+                          "name": "Content-Type",
+                          "value": "application/json"
+                        }
+                      ],
+                      "body": "{\n  \"s1\": \"abc\",\n  \"s2\": \"def\",\n  \"s3\": \"\",\n  \"s4\": null,\n  \"s5\": null,\n  \"s6\": null,\n  \"s7\": \"xyz\",\n  \"s8\": \"xyz\"\n}",
```
pksunkara: There should be no `s3` key in here. It should not exist in the JSON body at all since it is optional.

src/refract/RenderJSONVisitor.cc
```diff
@@ -63,11 +64,42 @@ namespace refract
...
+        return false;
+    }
+
     void RenderJSONVisitor::visit(const MemberElement& e) {
         RenderJSONVisitor renderer;
 
         if (e.value.second) {
-            renderer.visit(*e.value.second);
+            if (isNullable(e) && e.value.second->empty() == true) {
```
pksunkara: `e.value.second->empty()` should be enough. No need to do `e.value.second->empty() == true`

pksunkara:
Just a few minor issues left.

src/refract/RenderJSONVisitor.cc
obihann: without it, if we have something like this:

```
+ s7: xyz (required, nullable)
```

The json outputted is

```
"s7": null
```

Without that check the tests will fail because it expects this:

```
"s7": "xyz"
```

test/fixtures/render/nullable.json
obihann: If I run `nullable.apib` through drafter currently I get the following body/content:

```
"content": "\"{\n  \\"s1\\": \\"abc\\",\n  \\"s2\\": \\"def\\",\n  \\"s3\\": \\"\\",\n  \\"s4\\": null,\n  \\"s5\\": null,\n  \\"s6\\": null,\n  \\"s7\\": \\"xyz\\",\n
\\"s8\\": \\"xyz\\"\n}\","
```

That includes `"s3":""`. Is that not correct?

obihann: If that is incorrect should their be an issue created, since that behaviour is not part of my changes. I could probably look into fixing if it is actually a bug.

pksunkara: Yeah, please create an issue for that. You need not fix it.

src/refract/RenderJSONVisitor.cc
pksunkara: I am not asking you to remove the whole check. I am asking you to change the line to the following:

```cpp
if (isNullable(e) && e.value.second->empty()) {
```

obihann: I need more coffee, makes perfect sense now :+1:

test/fixtures/render/nullable.json
obihann: #121 has been created for this

obihann:
@pksunkara I believe the issues have now been resolved. I will rebase my latest commits so that their is only 1 for the final merge.

pksunkara:
@obihann Have you looked at this? https://github.com/apiaryio/drafter/pull/116#discussion_r40329382. It still contains bad diffs.

obihann:
@pksunkara sorry about that, its cleaned up now though.

pksunkara:
Thanks for the help @obihann :smile:

obihann:
No problem, thanks to all of you for the help you gave me. Hopefully I can help out more going forward.

PR #126, opened 2015-09-28
Added reporting and improved parse result serialization

pksunkara:
This PR does the following:

* Improves `drafter::WrapResult` and `drafter::WrapParseResult`
* Does parse result element building before serialising it
* Adds parse result reporting
* Moves a few functions around
* Adds a new render strategy

@klokane I have other PRs but I want this PR to be merged before they are


klokane:
looks gook. But i do not understand meaning of refract::IElement::rElement``, can you explain please?

pksunkara:
`refract::IElement::rElement` denotes that the particular element should use `rFull` strategy while the elements in the `content` should use `rCompact`. It is basically a mix of `rFull` and `rCompact`.

klokane:
Can you, please, use more descriptive name? eg. `rCompactContent`.
Do we have case where is it used? It is for `location` in parse result?

pksunkara:
It is used for `location` in parse result.

klokane:
Not sure if it is good idea to change behavior of "library" for one specific case
You can do it by set "compact" on parent element.
Is there defined behavior for "deeply nested children"?

pksunkara:
> You can do it by set "compact" on parent element. Is there defined behavior for "deeply nested children"?

I tried that. I tried all kinds of combinations of render strategies. But finally I decided that we really need one.

src/RefractParseResult.cc
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    refract::IElement* SourceMapToRefract(const mdp::CharactersRangeSet& sourceMap)
+    {
+        refract::ArrayElement* element = new refract::ArrayElement;
+        RefractElements content;
+
+        element->element(SerializeKey::SourceMap);
+
+        for (mdp::CharactersRangeSet::const_iterator it = sourceMap.begin();
```
klokane: Please use `std::transform` instead of `for` loop.
```
std::transform(sourceMap.begin(), sourceMap.end(), std::back_inserter(content), SourceMapRowToRefract);
```
It is easier to understand to one line of code than to six lines.

pksunkara:
Updated.

klokane:
Ok, I checked rendering strategy. It seems there is no way how to implement it inside drafter.
All other is ok. Merging

PR #129, opened 2015-09-29
Removed WITH_REFRACT from most places

### LINKED ISSUE #103: Remove all conditional parts around macro _WITH_REFRACT_
Refract is used in both types of AST (refract/ast) we should remove all around this define

pksunkara:
This PR along with #126 removes most of the WITH_REFRACT macros except the main one. This PR fixes #103

### DIFF
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -48,7 +48,6 @@ namespace drafter {
             return body;
         }
 
-#if _WITH_REFRACT_
         switch (renderFormat) {
             case JSONRenderFormat:
             {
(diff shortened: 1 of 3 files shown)


klokane:
What about SerializeAST.cc? By my opinion, we need just only `#ifdef` in `WrapDataStructure`.
Additionaly, please, add `#else` part into `WrapDataStructure`

pksunkara:
The `_WITH_REFRACT_` in SerializeAST.cc has been first removed in #126. This PR removes the left over ones. Since this PR can be merged directly to master, I made this. Otherwise, I would have waited for #126 to be merged first.

PR #131, opened 2015-09-30
Key 'warnings' in WrapResult should be an empty array

pksunkara:
This PR makes sure that the `warnings` key in `drafter::WrapResult` should be an empty array when no warnings are present. I am doing this because that is how protagonist has been serialising until now. And since `WrapResult` is synonymous with the output from protagonist, we need to make sure they are similar.

This PR also does the following minor things:

* Expose `drafter::WrapAnnotation` so that protagonist can use it.
* Updates the sos submodule.

Blocked by #126 and need to be rebased once that is merged.

### DIFF
--- a/src/SerializeResult.cc
+++ b/src/SerializeResult.cc
@@ -16,7 +16,7 @@
 
 using namespace drafter;
 
-static sos::Object WrapLocation(const mdp::BytesRange& range)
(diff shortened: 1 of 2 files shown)


pksunkara:
@zdne Need your go ahead on this.

klokane:
code is OK.

danielgtaylor:
LGTM :+1:

PR #147, opened 2015-10-16
Do not render empty elements within an array inside JSON visitor

### LINKED ISSUE #142: Regression: array[string] is not getting rendered correctly
# Example

```apib
FORMAT: 1A

# Example

# GET /
    + Attributes (object)
        + a (array[string])
            + b
            + c
```

# Old behaviour
Response body is rendered as

```json
{"a": ["b", "c"]}
```

# New behaviour
Response body is rendered as

```json
{"a": ["", "b", "c"]}
```

# Expected
Old behaviour is correct.

kylef:
Fixes #142

### DIFF
--- a/src/refract/RenderJSONVisitor.cc
+++ b/src/refract/RenderJSONVisitor.cc
@@ -240,7 +240,7 @@ namespace refract
              it != val->end();
              ++it) {
 
-            if (*it) {
+            if (*it && !(*it)->empty()) {
                 renderer.visit(*(*it));
             }
         }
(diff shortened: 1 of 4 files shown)


pksunkara:
I don't this is the right fix. The MSON refract itself is wrong in the first place.

> If nested type is primitive type, don't add it to content.
If nested type is non-primitive, add it to content.

klokane:
@pksunkara I mean it is correct solution. 
There are "empty elements" in refract as "type hint". Eg. for `array[string, number]` there is no other way to sent type hints into refract. So way how to corectly show it in JSON is do not render empty elements.

If you do not agree with this, can you please more detail explain your comment?

@kyle please just use `empty()` instead of `hasValue`

pksunkara:
Okay, please rebase.

kylef:
This has been updated.

PR #148, opened 2015-10-16
Source map support for Refract

klokane:
Source Map support for refract


klokane:
Hmm seems there are some problems on clang and msvc working on fix

src/RefractSourceMap.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include "Serialize.h"
+
+namespace drafter {
+
+
+    refract::IElement* BytesRangeToRefract(const mdp::BytesRange& bytesRange);
+
+    template<typename T>
+    refract::IElement* SourceMapToRefract(const T& sourceMap)
```
pksunkara: Why are all these function bodies in header file? They should be in `.cc` file. Header file only contains the signatures if you look at `RefractAPI.h` and `RefractParseResult.h`.

src/Render.cc
```diff
@@ -57,13 +64,26 @@ namespace drafter {
...
             {
                 refract::RenderJSONVisitor renderer;
-                refract::IElement* element = MSONToRefract(attributes);
+
+                refract::IElement* element = MSONToRefract(*attributes);
+
+                if (!element) {
+                    return body;
+                }
+
```
pksunkara: We don't need these lines at all since `ExpandRefract` checks them.

src/Render.cc
```diff
@@ -57,13 +64,26 @@ namespace drafter {
...
+                if (!expanded) {
+                    return body;
+                }
+
                 renderer.visit(*expanded);
+
                 delete expanded;
 
                 return renderer.getString();
+
```
pksunkara: Why the extra line?

src/Render.cc
```diff
@@ -57,13 +64,26 @@ namespace drafter {
...
+
                 refract::IElement* expanded = ExpandRefract(element, registry);
 
+                if (!expanded) {
+                    return body;
+                }
+
                 renderer.visit(*expanded);
+
```
pksunkara: Why the extra line?

src/Render.cc
```diff
@@ -73,14 +93,14 @@ namespace drafter {
...
-    Asset renderPayloadSchema(const Payload& payload) {
-        Asset schema = payload.schema;
+    Asset renderPayloadSchema(const SectionInfo<Payload>& payload) {
+        Asset schema = payload.section.schema;
         RenderFormat renderFormat = JSONSchemaRenderFormat;
 
-        if (!payload.schema.empty() || payload.attributes.empty() || renderFormat == UndefinedRenderFormat) {
-            return schema;
+        if (!payload.section.schema.empty() || payload.section.attributes.empty() || renderFormat == UndefinedRenderFormat) {
+            return payload.section.schema;
```
pksunkara: This should be `return schema;`

src/Render.cc
```diff
@@ -73,14 +93,14 @@ namespace drafter {
...
         }
 
-        return schema;
+        return payload.section.schema;
```
pksunkara: This should be `return schema;`

src/Render.cc
klokane: In genereal, `MSONToRefract()` can return NULL

pksunkara: Yes, that is checked by `ExpandRefract` below and transferred the same NULL to expanded and which gets caught by the `if (!expanded)` condition. We need to have the above condition.

src/Render.cc
```diff
@@ -40,16 +40,23 @@ namespace drafter {
...
 
-        if (attributes.empty() && action != NULL && !action->attributes.empty()) {
-            attributes = action->attributes;
+        //Attributes attributes = payload.section.attributes;
+        SectionInfo<Attributes> payloadAttributes = MAKE_SECTION_INFO(payload, attributes);
+        SectionInfo<Attributes> actionAttributes = MAKE_SECTION_INFO(action, attributes);
+
+        // hold attributes via pointer - because problems with assignment in SectionInfo<>
+        SectionInfo<Attributes>* attributes = &payloadAttributes;
+
```
pksunkara: Why the extra line?

src/Render.cc
```diff
@@ -40,16 +40,23 @@ namespace drafter {
...
-        Asset body = payload.body;
-        RenderFormat renderFormat = findRenderFormat(getContentTypeFromHeaders(payload.headers));
-        Attributes attributes = payload.attributes;
+    Asset renderPayloadBody(const SectionInfo<Payload>& payload, const SectionInfo<Action>& action, const refract::Registry& registry) {
+        Asset body = payload.section.body;
+        RenderFormat renderFormat = findRenderFormat(getContentTypeFromHeaders(payload.section.headers));
 
-        if (attributes.empty() && action != NULL && !action->attributes.empty()) {
-            attributes = action->attributes;
+        //Attributes attributes = payload.section.attributes;
```
pksunkara: Please remove this line.

test/draftertest.h
```diff
@@ -56,11 +56,18 @@ namespace draftertest {
 
     struct FixtureHelper {
 
-        static bool handleBlueprintJSON(const std::string& basepath, drafter::ASTType astType = drafter::NormalASTType, bool expand = false, bool mustBeOk = true) {
+        //static bool handleBlueprintJSON(const std::string& basepath, drafter::ASTType astType = drafter::NormalASTType, bool expand = false, bool mustBeOk = true) {
```
pksunkara: Please remove this line.

src/SerializeAST.h
```diff
@@ -10,10 +10,21 @@
 #define DRAFTER_SERIALIZEAST_H
 
 #include "Serialize.h"
+#include "SectionProcessor.h" // from Blueprint
```
pksunkara: Why the comment at the end of the line?

test/draftertest.h
```diff
@@ -69,7 +76,9 @@ namespace draftertest {
             std::stringstream outStream;
             sos::SerializeJSON serializer;
 
-            serializer.process(drafter::WrapBlueprint(blueprint.node, astType, expand), outStream);
+            snowcrash::SourceMap<snowcrash::Blueprint>* sourceMap = NULL;
```
pksunkara: Why is this line here? We are not using the variable anywhere?

src/SerializeAST.h
```diff
@@ -10,10 +10,21 @@
...
-    sos::Object WrapBlueprint(const snowcrash::Blueprint& blueprint, const ASTType astType, bool expand = false);
+    struct WrappingContext {
+        const ASTType astType;
+        const bool expandMSON;
+        const bool exportSourceMap;
+
+        WrappingContext(const ASTType astType, const bool expandMSON, const bool exportSourceMap) 
+            : astType(astType), expandMSON(expandMSON), exportSourceMap(exportSourceMap) {
+        }
+    };
```
pksunkara: I am okay with the `WrappingContext` idea, but I am wondering if we should use `snowcrash::ParserOptions`.

test/fixtures/sourcemap/full.apib
```diff
@@ -0,0 +1,121 @@
...
+
+            <response schema>
+
++ Response 201
+
+    [<resource name>][]
+
++ Response 201 (application/json)
+
+    + Attributes (<data structure name>)
```
pksunkara: This PR is missing integration tests. Instead of just copying the blueprint file from `features/fixtures`, please add an integration test similar to how I added for ParseResult refract.

Please remove the above lines and just use a small sample in here.

src/SerializeAST.cc
```diff
@@ -813,20 +855,22 @@ sos::Object WrapBlueprintAST(const Blueprint& blueprint)
...
     sos::Object blueprintObject;
     snowcrash::Error error;
 
-    try {
-        RegisterNamedTypes(blueprint.content.elements());
-        ExpandMSON = expand;
+#if _WITH_REFRACT_
+    registerNamedTypes(blueprint.node.content.elements());
+    ExpandMSON = context.expandMSON;
+#endif
```
pksunkara: This is a rebase issue. I will fix this.

src/SerializeAST.cc
```diff
@@ -770,16 +775,53 @@ bool IsElementResourceGroup(const Element& element)
     return element.element == Element::CategoryElement && element.category == Element::ResourceGroupCategory;
 }
 
-sos::Object WrapBlueprintRefract(const Blueprint& blueprint)
+#if _WITH_REFRACT_
```
pksunkara: This is a rebase issue. I will fix this.

src/SerializeAST.cc
```diff
@@ -512,8 +512,13 @@ sos::Object WrapPayload(const Payload& payload, const Action* action = NULL)
                       WrapCollection<Header>()(payload.headers, WrapHeader));
 
     // Render using boutique
-    snowcrash::Asset payloadBody = renderPayloadBody(payload, action, GetNamedTypesRegistry());
-    snowcrash::Asset payloadSchema = renderPayloadSchema(payload);
+//<<<<<<< HEAD
+//    snowcrash::Asset payloadBody = renderPayloadBody(payload, action, GetNamedTypesRegistry());
+//    snowcrash::Asset payloadSchema = renderPayloadSchema(payload);
+//=======
```
pksunkara: This is a rebase issue. I will fix this.

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+
+            if (collection.size() == sourceMaps.collection.size()) {
+                sections = Zip<CollectionType>(collection, sourceMaps.collection, SectionInfoCollection::MakeSectionInfo<typename T::value_type>);
+            }
+            else {
+                std::transform(collection.begin(), collection.end(), 
+                               std::back_inserter(sections), 
+                               MakeSectionInfoWithoutSourceMap<typename T::value_type>);
+            }
+
```
pksunkara: Why the extra line?

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+                sections = Zip<CollectionType>(collection, sourceMaps.collection, SectionInfoCollection::MakeSectionInfo<typename T::value_type>);
+            }
+            else {
+                std::transform(collection.begin(), collection.end(), 
+                               std::back_inserter(sections), 
+                               MakeSectionInfoWithoutSourceMap<typename T::value_type>);
+            }
+            
+        }
+
```
pksunkara: Why the extra line?

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+        static SectionInfo<U> MakeSectionInfo(const U& section, const snowcrash::SourceMap<U>& sourceMap)
+        {
+            return SectionInfo<U>(section, sourceMap);
+        }
+
+        SectionInfoCollection(const T& collection, const snowcrash::SourceMap<T>& sourceMaps)
+        //SectionInfoCollection(const SectionInfo<T>& sectionInfo)
+        {
+            //const T& collection = sectionInfo.section;
+            //const snowcrash::SourceMap<T>& sourceMaps = sectionInfo.sourceMap;
```
pksunkara: Please remove the commented lines above.

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+
+    template<typename ResultType, typename Collection1, typename Collection2, typename BinOp>
+    ResultType Zip(const Collection1& collection1, const Collection2& collection2, const BinOp& Combinator) {
+        ResultType result;
+        std::transform(collection1.begin(), collection1.end(), collection2.begin(), std::back_inserter(result), Combinator);
+        return result;
+    }
+
+    template<typename T>
+    // FIXME: inherit directly from `std::vector<SectionInfo<typename T::value_type> >` instead of hold it as member
```
pksunkara: Can we please move this comment above the `template<typename T>` line?

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+    }
+
+
+    template <typename T>
+    SectionInfo<T> MakeSectionInfoWithoutSourceMap(const T& section)
+    {
+        return SectionInfo<T>(section, SectionInfo<T>::NullSourceMap());
+    }
+
+#define MAKE_SECTION_INFO(from, member) MakeSectionInfo(from.section.member, from.sourceMap.member, from.hasSourceMap())
```
pksunkara: Couldn't we have moved this to the top of the file since this is a macro? We don't need this to be here.

src/Serialize.h
```diff
@@ -33,6 +33,98 @@ namespace drafter {
...
+            return &sourceMap != &null;
+        }
+    };
+
+    template <typename T>
+    SectionInfo<T> MakeSectionInfo(const T& section, const snowcrash::SourceMap<T>& sourceMap, const bool hasSourceMap)
+    {
+        return SectionInfo<T>(section, hasSourceMap ? sourceMap : SectionInfo<T>::NullSourceMap());
+    }
+
```
pksunkara: Why the extra line?

src/RefractSourceMap.h
```diff
@@ -0,0 +1,73 @@
...
+        refract::IElement* element = refract::IElement::Create(LiteralTo<T>(sectionInfo.section));
+        AttachSourceMap(element, sectionInfo);
+
+        return element;
+    }
+
+}
+
+#endif // #ifndef DRAFTER_REFRACTSOURCEMAP_H
+
```
pksunkara: Why the extra line?

src/RefractSourceMap.h
```diff
@@ -0,0 +1,73 @@
...
+        return element;
+    }
+
+    template<typename T>
+    void AttachSourceMap(refract::IElement* element, const T& sectionInfo) 
+    {
+        if (sectionInfo.hasSourceMap() && !sectionInfo.sourceMap.sourceMap.empty()) {
+
+            element->attributes[SerializeKey::SourceMap] = SourceMapToRefract(sectionInfo.sourceMap.sourceMap);
+
```
pksunkara: Why the extra line?

src/RefractParseResult.cc
```diff
@@ -67,7 +67,10 @@ namespace drafter {
 
         element->element(SerializeKey::ParseResult);
 
-        content.push_back(BlueprintToRefract(blueprint.node));
+        snowcrash::SourceMap<snowcrash::Blueprint> nullSourceMap = SectionInfo<snowcrash::Blueprint>::NullSourceMap() ;
+        //SectionInfo<Blueprint> section(blueprint.node, blueprint.sourceMap);
```
pksunkara: Please remove the comment.

src/RefractDataStructure.cc
```diff
@@ -819,7 +996,7 @@ namespace drafter {
                 break;
 
             default:
-                throw snowcrash::Error("unknown type of data structure", snowcrash::MSONError);
+                throw std::runtime_error("Unhandled type of DataStructure");
```
pksunkara: This is a rebase issue. I will fix it.

src/RefractDataStructure.cc
```diff
@@ -514,51 +663,69 @@ namespace drafter {
...
-            SetElementType(element->value.first, property.name.variable.typeDefinition);
+            snowcrash::SourceMap<mson::Literal> sourceMap;
+            sourceMap.sourceMap.append(property.sourceMap.name.sourceMap);
+
+            refract::IElement* key = PrimitiveToRefract(MakeSectionInfo(property.section.name.variable.values.begin()->literal, sourceMap, property.hasSourceMap()));
+
+            key->attributes[SerializeKey::Variable] = refract::IElement::Create(true);
+
+            element->set(key, value);
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -514,51 +663,69 @@ namespace drafter {
...
             }
 
-            element->set(property.name.variable.values.begin()->literal, value);
-            element->value.first->attributes[SerializeKey::Variable] = refract::IElement::Create(true);
-            SetElementType(element->value.first, property.name.variable.typeDefinition);
+            snowcrash::SourceMap<mson::Literal> sourceMap;
+            sourceMap.sourceMap.append(property.sourceMap.name.sourceMap);
+
+            refract::IElement* key = PrimitiveToRefract(MakeSectionInfo(property.section.name.variable.values.begin()->literal, sourceMap, property.hasSourceMap()));
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -500,8 +620,37 @@ namespace drafter {
         }
     }
 
+    template<typename T>
+    refract::IElement* DescriptionToRefract(const ElementData<T>& data) 
+    {
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -388,35 +486,40 @@ namespace drafter {
...
-            if (valueMember.valueDefinition.values.empty() || (valueMember.valueDefinition.typeDefinition.typeSpecification.nestedTypes.size() > 1)) {
-                InjectNestedTypeInfo<typename T::ValueType>()(valueMember, data.values);
+            if (vm.section.valueDefinition.values.empty() || (vm.section.valueDefinition.typeDefinition.typeSpecification.nestedTypes.size() > 1)) {
+                InjectNestedTypeInfo<typename T::ValueType>()(vm, data.values);
+                InjectNestedTypeInfoSourceMaps<typename T::ValueType>()(vm, data.valuesSourceMap);
             }
         }
     };
 
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -388,35 +486,40 @@ namespace drafter {
...
-        void operator ()(const mson::ValueMember& valueMember)
+        void operator ()(const SectionInfo<mson::ValueMember>& vm)
         {
             Fetch<typename T::ValueType> fetch;
+            FetchSourceMap<typename T::ValueType> fetchSourceMap;
 
-            if (!valueMember.valueDefinition.values.empty()) {
-                mson::TypeAttributes attrs = valueMember.valueDefinition.typeDefinition.attributes;
-                const mson::Value& value = *valueMember.valueDefinition.values.begin();
+            if (!vm.section.valueDefinition.values.empty()) {
```
pksunkara: I am pretty sure this is a rebase issue. I will fix it.

src/RefractDataStructure.cc
```diff
@@ -347,19 +407,34 @@ namespace drafter {
...
+                // do nothing
             }
         };
 
         template<bool dummy>
         struct InjectNestedTypeInfo<RefractElements, dummy> {
-            void operator()(const mson::ValueMember& valueMember, std::vector<typename T::ValueType>& values) {
+            void operator()(const SectionInfo<mson::ValueMember>& vm, ValueCollectionType& values) {
                 // inject type info into arrays [ "type", {}, {}, null ]
-                const mson::TypeNames& nestedTypes = valueMember.valueDefinition.typeDefinition.typeSpecification.nestedTypes;
```
pksunkara: This is a rebase issue. I will fix it.

src/RefractDataStructure.cc
```diff
@@ -347,19 +407,34 @@ namespace drafter {
             }
         };
 
+        template <typename U, bool dummy = true>
+        struct FetchSourceMap {  // primitive values
+
+            template <typename S>
+            void operator()(S& storage, const SectionInfo<mson::ValueMember>& vm) {
```
pksunkara: Please use `valueMember` instead of `vm`.

src/RefractDataStructure.cc
```diff
@@ -330,11 +390,11 @@ namespace drafter {
         struct Fetch<RefractElements, dummy> { // Array|Object
 
             template <typename S>
-            void operator()(S& storage, const mson::ValueMember& valueMember) {
-                const mson::BaseTypeName type = SelectNestedTypeSpecification(valueMember.valueDefinition.typeDefinition.typeSpecification.nestedTypes);
+            void operator()(S& storage, const SectionInfo<mson::ValueMember>& vm) {
```
pksunkara: This is a rebase issue, I will fix it.

src/RefractDataStructure.cc
```diff
@@ -316,11 +375,12 @@ namespace drafter {
         struct Fetch {  // primitive values
 
             template <typename S>
-            void operator()(S& storage, const mson::ValueMember& valueMember) {
-                if (valueMember.valueDefinition.values.size() > 1) {
-                    throw snowcrash::Error("only one value is supported for primitive types", snowcrash::MSONError);
+            void operator()(S& storage, const SectionInfo<mson::ValueMember>& vm) {
+                if (vm.section.valueDefinition.values.size() > 1) {
+                    throw std::logic_error("For primitive types is just one value supported");
```
pksunkara: This is a rebase issue. I will fix it.

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+        {
+        }
+
+        void operator()(const SectionInfo<ValueType>& value)
+        {
+            for_each(value.section.begin(), value.section.end(), std::bind1st(std::mem_fun(&ElementType::push_back), element));
+        }
+    };
+
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+        typedef T ElementType;
+        typedef V ValueType;
+        ElementType*& element;
+        Append(ElementType*& e) : element(e)
+        {
+        }
+
+        void operator()(const SectionInfo<ValueType>& value)
+        {
+            //throw std::logic_error("Can not append to primitive type");
```
pksunkara: Please remove this line.

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+#include "RefractSourceMap.h"
+
 namespace drafter {
 
+    template <typename T, typename V = typename T::ValueType>
+    struct Append {
+        typedef T ElementType;
+        typedef V ValueType;
+        ElementType*& element;
+        Append(ElementType*& e) : element(e)
```
pksunkara: Please add empty line before this.

src/RefractAPI.h
```diff
@@ -18,8 +18,9 @@ namespace drafter {
 
     void RegisterNamedTypes(const snowcrash::Elements& elements);
 
-    refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure, bool expand = false);
-    refract::IElement* BlueprintToRefract(const snowcrash::Blueprint& blueprint);
+    //refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure, bool expand = false);
```
pksunkara: Please remove this line.

src/RefractAPI.h
```diff
@@ -18,8 +18,9 @@ namespace drafter {
...
+    refract::IElement* DataStructureToRefract(const SectionInfo<snowcrash::DataStructure>& dataStructure, bool expand = false);
+    refract::IElement* BlueprintToRefract(const drafter::SectionInfo<snowcrash::Blueprint>& blueprint);
```
pksunkara: We don't need `drafter::` in this line.

src/RefractAPI.cc
```diff
@@ -39,6 +41,18 @@ namespace drafter {
             }
         }
 
+        refract::ArrayElement* CreateArrayElement(refract::IElement* value, bool rFull = false)
```
pksunkara: We already have a `CreateArrayElement` in this file. Please use that by adding a parameter.

src/RefractDataStructure.cc
```diff
@@ -489,10 +592,27 @@ namespace drafter {
             }
         }
 
+        template <typename T>
+        struct MakeSectionInfo_ {
```
pksunkara: Is this used anywhere?

src/RefractDataStructure.cc
```diff
@@ -489,10 +592,27 @@ namespace drafter {
...
+                return MakeSectionInfo<T>(v, sm, hasSourceMap);
+            }
+        };
+
         template<typename T>
-        void TransformElementData(T* element, ElementData<T>& data) {
+        void TransformElementData(T* element, ElementData<T>& data, bool hasSourceMap) {
+
+            if (data.values.size() != data.valuesSourceMap.size()) {
+                throw std::logic_error("Internal: count of source maps is not equal to count of elements");
```
pksunkara: Please throw `snowcrash::Error` as used in this file at other places. And don't use capital letters and the word `internal`.

src/Serialize.h
klokane: I'm going to fix this by implementation

src/Serialize.h
```diff
@@ -33,6 +33,101 @@ namespace drafter {
 
     refract::Registry& GetNamedTypesRegistry();
 
+    template<typename T>
+    struct SectionInfo {
+        typedef SectionInfo<T> Type;
+        typedef T SectionType;
+        typedef snowcrash::SourceMap<SectionType> SourceMapType;
```
pksunkara: `typedef snowcrash::SourceMap<T> SourceMapType;`

I know it's the same but this would be more easier to read and understand for other people.

src/Serialize.h
```diff
@@ -33,6 +33,101 @@ namespace drafter {
...
+
+        const SectionType& section;
```
pksunkara: My main concern with the whole thing. Why are we not simply using `ParseResult` (maybe after extending it a bit) instead of this? If we are using `SectionInfo`, why is this not named `node`? We already were using `node` and `sourceMap` in snowcrash, why not simply continue the naming convention?

src/SerializeAST.h
klokane: I think about it, but ParseOptions contains set of instruction "how to parse" while WrappingContext contains diferent set of rules (yes some of them may be same).

pksunkara: I don't agree. Extending the `enum snowcrash::BlueprintParserOption` as `enum drafter::BlueprintWrapOption` would be better instead of introducing a different way of doing this with `WrappingContext`.

src/Serialize.h
klokane: I start with ParseResult, but after few tests I create new structure. 
Mainly ParseResult contains values instead of reference (it mean - copy value while creating ParseResult instance), I use just reference to avoid copy.  Additionaly ParseResult contain Annotation.

pksunkara: There is a `struct ParseResultRef` just for that in snowcrash. As I start looking at the code, I am more in favour of using `ParseResult` related stuff.

klokane: I do not see to problem to rename "section" to "node"

klokane: > There is a struct ParseResultRef just for that in snowcrash. As I start looking at the code, I am more in favour of using ParseResult related stuff.

In drafter we are working with "const reference". Do you mean we should create another structure"ParseResultConstRef"?

src/Serialize.h
```diff
@@ -33,6 +33,99 @@ namespace drafter {
...
+
+    template<typename ResultType, typename Collection1, typename Collection2, typename BinOp>
+    ResultType Zip(const Collection1& collection1, const Collection2& collection2, const BinOp& Combinator) {
+        ResultType result;
+        std::transform(collection1.begin(), collection1.end(), collection2.begin(), std::back_inserter(result), Combinator);
+        return result;
+    }
+
+    template<typename T>
+    struct SectionInfoCollection  : public std::vector<SectionInfo<typename T::value_type> > {
```
pksunkara: Can we please move all this `SectionInfo` into a separate `NodeInfo.h` header file?

src/Serialize.h
```diff
@@ -33,6 +33,99 @@ namespace drafter {
...
+        static SectionInfo<U> MakeSectionInfo(const U& section, const snowcrash::SourceMap<U>& sourceMap)
+        {
+            return SectionInfo<U>(section, sourceMap);
+        }
+
+        SectionInfoCollection(const T& collection, const snowcrash::SourceMap<T>& sourceMaps)
+        {
+
+            if (collection.size() == sourceMaps.collection.size()) {
+                CollectionType sections = Zip<CollectionType>(collection, sourceMaps.collection, SectionInfoCollection::MakeSectionInfo<typename T::value_type>);
```
pksunkara: Do we need the `SectionInfoCollection::` in this line?

src/RefractAPI.cc
```diff
@@ -60,7 +74,9 @@ namespace drafter {
                 element->element(key);
             }
 
-            std::transform(collection.begin(), collection.end(), std::back_inserter(content), transformFunctor);
+            SectionInfoCollection<C> sectionInfoCollection(collection.section, collection.sourceMap);
+
```
pksunkara: Why the extra line?

src/RefractAPI.cc
```diff
@@ -100,6 +116,26 @@ namespace drafter {
 
     refract::IElement* DataStructureToRefract(const snowcrash::DataStructure& dataStructure, bool expand)
     {
+       refract::IElement* msonElement = MSONToRefract(MakeSectionInfoWithoutSourceMap(dataStructure));
```
pksunkara: Looks like you are not using this function anywhere. Please remove this whole function.

src/RefractAPI.cc
```diff
@@ -126,50 +162,50 @@ namespace drafter {
...
 
         element->element(SerializeKey::Enum);
 
+        // FIXME: nearly duplicit code in ExtractParameter()
         // Add sample value
-        if (!parameter.exampleValue.empty()) {
+        if (!parameter.section.exampleValue.empty()) {
             refract::ArrayElement* samples = new refract::ArrayElement;
-            samples->push_back(CreateArrayElement(LiteralTo<T>(parameter.exampleValue), true));
+            // FIXME: sourcemap of exampleValue does is not equal to
```
pksunkara: I don't understand what this comment is saying.

src/RefractAPI.cc
```diff
@@ -177,15 +213,18 @@ namespace drafter {
...
-    refract::IElement* ExtractParameter(const snowcrash::Parameter& parameter)
+    refract::IElement* ExtractParameter(const SectionInfo<snowcrash::Parameter>& parameter)
     {
         refract::IElement* element = NULL;
 
-        if (parameter.values.empty()) {
-            element = refract::IElement::Create(LiteralTo<T>(parameter.exampleValue));
+        if (parameter.section.values.empty()) {
+            element = refract::IElement::Create(LiteralTo<T>(parameter.section.exampleValue));
+            AttachSourceMap(element, MAKE_SECTION_INFO(parameter, exampleValue));
```
pksunkara: Couldn't you have used `LiteralToRefract` for the above two lines?

src/RefractAPI.cc
```diff
@@ -177,15 +213,18 @@ namespace drafter {
...
 
-            if (!parameter.defaultValue.empty()) {
-                element->attributes[SerializeKey::Default] = refract::IElement::Create(LiteralTo<T>(parameter.defaultValue));
+            if (!parameter.section.defaultValue.empty()) {
+                refract::IElement* defaultElement = refract::IElement::Create(LiteralTo<T>(parameter.section.defaultValue));
+                AttachSourceMap(defaultElement, MAKE_SECTION_INFO(parameter, defaultValue));
+                element->attributes[SerializeKey::Default] = defaultElement;
```
pksunkara: Could have used `LiteralToRefract` in just one line instead of the above 3 lines.

src/RefractAPI.cc
```diff
@@ -195,116 +234,125 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* ParameterToRefract(const snowcrash::Parameter& parameter)
+    refract::IElement* ParameterToRefract(const SectionInfo<snowcrash::Parameter>& parameter)
     {
         refract::MemberElement* element = new refract::MemberElement;
+
```
pksunkara: Why the extra line?

src/RefractAPI.cc
```diff
@@ -195,116 +234,125 @@ namespace drafter {
...
     }
 
-    refract::IElement* HeaderToRefract(const snowcrash::Header& header)
+    refract::IElement* HeaderToRefract(const SectionInfo<snowcrash::Header>& header)
     {
         refract::MemberElement* element = new refract::MemberElement;
-        element->set(refract::IElement::Create(header.first), refract::IElement::Create(header.second));
+
+        element->set(refract::IElement::Create(header.section.first), refract::IElement::Create(header.section.second));
+
```
pksunkara: Why the extra line?

src/RefractAPI.cc
```diff
@@ -195,116 +234,125 @@ namespace drafter {
...
         RefractElements content;
 
         // Use HTTP method to recognize if request or response
-        if (action == NULL || action->method.empty()) {
+        if (action.isNull() || action.section.method.empty()) {
             element->element(SerializeKey::HTTPResponse);
 
-            if (payload != NULL && !payload->name.empty()) {
-                element->attributes[SerializeKey::StatusCode] = refract::IElement::Create(payload->name);
+            if (!payload.isNull()) {
```
pksunkara: You also need to check for `!payload.section.name.empty()`

src/RefractAPI.cc
```diff
@@ -195,116 +234,125 @@ namespace drafter {
...
             }
         }
         else {
             element->element(SerializeKey::HTTPRequest);
-            element->attributes[SerializeKey::Method] = refract::IElement::Create(action->method);
+            element->attributes[SerializeKey::Method] = PrimitiveToRefract(MAKE_SECTION_INFO(action, method));
 
-            if (payload != NULL) {
-                element->attributes[SerializeKey::Title] = refract::IElement::Create(payload->name);
+            if (!payload.isNull()) {
```
pksunkara: You also need to check for `!payload.section.name.empty()`

src/RefractAPI.cc
```diff
@@ -195,116 +234,125 @@ namespace drafter {
...
             return element;
         }
 
-        if (!payload->headers.empty()) {
-            element->attributes[SerializeKey::Headers] = CollectionToRefract<refract::ArrayElement>(payload->headers, HeaderToRefract, SerializeKey::HTTPHeaders, refract::IElement::rFull);
+        if (!payload.section.headers.empty()) {
+            element->attributes[SerializeKey::Headers] = CollectionToRefract<refract::ArrayElement>(MAKE_SECTION_INFO(payload, headers),
+                                                                                                    HeaderToRefract,
+                                                                                                    SerializeKey::HTTPHeaders,
+                                                                                                    refract::IElement::rFull);
```
pksunkara: Too big. Let's divide into two lines. :smile:

src/RefractAPI.cc
```diff
@@ -332,63 +380,77 @@ namespace drafter {
...
-            element->attributes[SerializeKey::Relation] = refract::IElement::Create(action.relation.str);
+        if (!action.section.relation.str.empty()) {
+            // FIXME: add SourceMap
+            element->attributes[SerializeKey::Relation] = refract::IElement::Create(action.section.relation.str);
         }
 
-        if (!action.uriTemplate.empty()) {
-            element->attributes[SerializeKey::Href] = refract::IElement::Create(action.uriTemplate);
+        if (!action.section.uriTemplate.empty()) {
+            // FIXME: This doues not chnge rendering, do we have exammple with uriTemplate?
```
pksunkara: What rendering are you talking about?

src/RefractAPI.cc
```diff
@@ -332,63 +380,77 @@ namespace drafter {
...
         RefractElements content;
 
         element->element(SerializeKey::Transition);
-        element->meta[SerializeKey::Title] = refract::IElement::Create(action.name);
+        element->meta[SerializeKey::Title] = PrimitiveToRefract(MAKE_SECTION_INFO(action, name));
 
-        if (!action.relation.str.empty()) {
-            element->attributes[SerializeKey::Relation] = refract::IElement::Create(action.relation.str);
+        if (!action.section.relation.str.empty()) {
+            // FIXME: add SourceMap
```
pksunkara: Why is the source map not attached here?

src/RefractAPI.cc
```diff
@@ -332,63 +380,77 @@ namespace drafter {
...
+        if (!action.section.attributes.empty()) {
+            refract::IElement* dataStructure = DataStructureToRefract(MAKE_SECTION_INFO(action, attributes));
             dataStructure->renderType(refract::IElement::rFull);
             element->attributes[SerializeKey::Data] = dataStructure;
         }
 
-        content.push_back(CopyToRefract(action.description));
+        content.push_back(CopyToRefract(MAKE_SECTION_INFO(action, description)));
+
+        typedef SectionInfoCollection<snowcrash::TransactionExamples> ExamplesType;
```
pksunkara: Please rename `ExamplesType` to just `Examples`

pksunkara: Similarly below for `ResponsesType` and `RequestsType`. We don't need to explicitly say `Type`.

src/RefractAPI.cc
```diff
@@ -399,23 +461,26 @@ namespace drafter {
...
         RefractElements content;
 
         element->element(SerializeKey::Resource);
-        element->meta[SerializeKey::Title] = refract::IElement::Create(resource.name);
-        element->attributes[SerializeKey::Href] = refract::IElement::Create(resource.uriTemplate);
 
-        if (!resource.parameters.empty()) {
-            element->attributes[SerializeKey::HrefVariables] = ParametersToRefract(resource.parameters);
+        element->meta[SerializeKey::Title] = PrimitiveToRefract(MAKE_SECTION_INFO(resource, name));
+
```
pksunkara: Why the extra line?

src/RefractAPI.cc
```diff
@@ -399,23 +461,26 @@ namespace drafter {
...
         }
 
-        content.push_back(CopyToRefract(resource.description));
-        content.push_back(DataStructureToRefract(resource.attributes));
+        content.push_back(CopyToRefract(MAKE_SECTION_INFO(resource, description)));
+        content.push_back(DataStructureToRefract(MAKE_SECTION_INFO(resource, attributes)));
 
-        std::transform(resource.actions.begin(), resource.actions.end(), std::back_inserter(content), ActionToRefract);
+        SectionInfoCollection<snowcrash::Actions> actions(resource.section.actions, resource.sourceMap.actions);
+        std::transform(actions.begin(), actions.end(), std::back_inserter(content), ActionToRefract);
```
pksunkara: Couldn't you have combined these 2 lines into a macro? I see this occurring many times.

src/RefractAPI.cc
```diff
@@ -423,23 +488,29 @@ namespace drafter {
...
+        else if (element.section.category == snowcrash::Element::DataStructureGroupCategory) {
             category->meta[SerializeKey::Classes] = CreateArrayElement(SerializeKey::DataStructures);
         }
 
-        snowcrash::Elements elements = element.content.elements();
-        std::transform(elements.begin(), elements.end(), std::back_inserter(content), ElementToRefract);
+        if (!element.section.content.elements().empty()) {
+            const snowcrash::SourceMap<snowcrash::Elements>& sourceMap = element.sourceMap.content.elements().collection.empty()
+                ? SectionInfo<snowcrash::Elements>::NullSourceMap()
+                : element.sourceMap.content.elements();
```
pksunkara: Let's use a proper `if...else` loop for this since this statement is spanning 3 lines anyway.

src/RefractAPI.cc
```diff
@@ -463,24 +534,24 @@ namespace drafter {
...
 
-    refract::IElement* BlueprintToRefract(const snowcrash::Blueprint& blueprint)
+    refract::IElement* BlueprintToRefract(const SectionInfo<snowcrash::Blueprint>& blueprint)
     {
         refract::ArrayElement* ast = new refract::ArrayElement;
         RefractElements content;
 
         ast->element(SerializeKey::Category);
         ast->meta[SerializeKey::Classes] = CreateArrayElement(SerializeKey::API);
-        ast->meta[SerializeKey::Title] = refract::IElement::Create(blueprint.name);
```
pksunkara: Why the extra line?

src/RefractParseResult.cc
```diff
@@ -67,7 +67,10 @@ namespace drafter {
...
+        SectionInfo<snowcrash::Blueprint> sectionBlueprint(blueprint.node, nullSourceMap);
+        content.push_back(BlueprintToRefract((sectionBlueprint)));
```
pksunkara: I think we could have made the above 3 code lines into a single line.

```cpp
content.push_back(BlueprintToRefract(MakeSectionInfoWithoutSourceMap(blueprint.node)));
```

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+        typedef typename T::ValueType ValueType;
+        ElementType*& element;
+
+        Append(ElementType*& e) : element(e)
+        {
+        }
+
+        void operator()(const SectionInfo<ValueType>& value)
+        {
+            for_each(value.section.begin(), value.section.end(), std::bind1st(std::mem_fun(&ElementType::push_back), element));
```
pksunkara: Please use `std::` before `for_each`

src/RefractDataStructure.cc
```diff
@@ -162,8 +186,11 @@ namespace drafter {
         //std::transform(elements.begin(), elements.end(),
         //               std::back_inserter(result),
         //               std::bind2nd(std::ptr_fun(MsonElementToRefract), nestedTypeName));
+        //
```
pksunkara: Why is this line here? Infact, can we remove the whole comment from here?

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+                element->set(value.section);
+                AttachSourceMap(element, value);
+            }
+        }
+    };
+
+    template <typename T>
+    struct Append<T, std::vector<refract::IElement*> > {
+        typedef T ElementType;
+        typedef typename T::ValueType ValueType;
```
pksunkara: `typedef ElementType::ValueType ValueType`

src/RefractDataStructure.cc
```diff
@@ -182,15 +209,30 @@ namespace drafter {
 
     template <typename T>
     struct ElementData {
-        std::vector<typename T::ValueType>  values;
+        typedef T ElementType;
+
+        typedef typename T::ValueType ValueType;
```
pksunkara: `typedef ElementType::ValueType ValueType`

src/RefractDataStructure.cc
```diff
@@ -204,20 +246,33 @@ namespace drafter {
...
         struct Fetch<RefractElements, dummy> {
-            RefractElements operator()(const mson::TypeSection& t, const mson::BaseTypeName& defaultNestedType) {
-                return MsonElementsToRefract(t.content.elements(), defaultNestedType);
+            RefractElements operator()(const SectionInfo<mson::TypeSection>& t, const mson::BaseTypeName& defaultNestedType) {
+                return MsonElementsToRefract(MakeSectionInfo(t.section.content.elements(), t.sourceMap.elements(), t.hasSourceMap()), defaultNestedType);
             }
         };
 
         template <typename U, bool dummy = true>
-        struct FetchTypeDefinition {};
```
pksunkara: What happened to the empty `FetchTypeDefinition`?

src/RefractDataStructure.cc
```diff
@@ -204,20 +246,33 @@ namespace drafter {
...
+                    return SectionInfo<U>::NullSourceMap();
+                }
+
+                snowcrash::SourceMap<U> sourceMap;
+                sourceMap.sourceMap = t.sourceMap.value.sourceMap;
+                return sourceMap;
+            }
+        };
+
+        template <typename U, bool dummy = true> struct FetchTypeDefinition;
```
pksunkara: Please move `struct FetchTypeDefinition` to a new line and leave the first line just for the template definition.

src/RefractDataStructure.cc
```diff
@@ -265,8 +320,9 @@ namespace drafter {
                     //
                     // FIXME: handle this by specialization for **Primitives**
                     // rewrite it to similar way to ExtractValueMember
-                    if (!ts.content.elements().empty()) { 
-                        data.values.push_back(fetch(ts, defaultNestedType));
+                    if (!ts.section.content.elements().empty()) { 
+                      data.values.push_back(fetch(ts, defaultNestedType));
+                      data.valuesSourceMap.push_back(fetchSourceMap(ts, defaultNestedType));
```
pksunkara: Indentation issue for the above 2 lines.

src/RefractDataStructure.cc
```diff
@@ -347,19 +407,34 @@ namespace drafter {
             }
         };
 
+        template <typename U, bool dummy = true>
+        struct FetchSourceMap {  // primitive values
+
+            template <typename S>
+            void operator()(S& storage, const SectionInfo<mson::ValueMember>& valueMember) {
+
```
pksunkara: Why the extra line?

src/RefractDataStructure.cc
```diff
@@ -347,19 +407,34 @@ namespace drafter {
...
+                snowcrash::SourceMap<typename T::ValueType> sourceMap = SectionInfo<typename T::ValueType>::NullSourceMap();
+                if (valueMember.hasSourceMap()) {
```
pksunkara: Please add empty line before this line.

src/RefractDataStructure.cc
```diff
@@ -372,6 +447,29 @@ namespace drafter {
             }
         };
 
+        template<typename X, bool dummy = true>
+        struct InjectNestedTypeInfoSourceMaps {
+            void operator()(const SectionInfo<mson::ValueMember>&, std::vector<snowcrash::SourceMap<typename T::ValueType> >&) {
```
pksunkara: You have `ValueSourceMapCollectionType` in `ElementData`.

src/RefractDataStructure.cc
```diff
@@ -372,6 +447,29 @@ namespace drafter {
...
+            }
+        };
+
+        template<bool dummy>
+        struct InjectNestedTypeInfoSourceMaps<RefractElements, dummy> {
+            void operator()(const SectionInfo<mson::ValueMember>& valueMember, std::vector<snowcrash::SourceMap<typename T::ValueType> >& values) {
```
pksunkara: You have `ValueSourceMapCollectionType` in `ElementData` which can be used here.

src/RefractDataStructure.cc
```diff
@@ -372,6 +447,29 @@ namespace drafter {
...
+                // inject type info into arrays [ "type", {}, {}, null ]
+                const mson::TypeNames& nestedTypes = valueMember.section.valueDefinition.typeDefinition.typeSpecification.nestedTypes;
+
+                if (!nestedTypes.empty() && GetType(valueMember.section.valueDefinition) != mson::EnumTypeName) {
+                    snowcrash::SourceMap<typename T::ValueType> sourceMap = SectionInfo<typename T::ValueType>::NullSourceMap();
+                    if (valueMember.hasSourceMap()) {
```
pksunkara: Please add an empty line before this.

src/RefractDataStructure.cc
```diff
@@ -489,10 +592,27 @@ namespace drafter {
...
-        void TransformElementData(T* element, ElementData<T>& data) {
+        void TransformElementData(T* element, ElementData<T>& data, bool hasSourceMap) {
+
+            if (data.values.size() != data.valuesSourceMap.size()) {
+                throw std::logic_error("Internal: count of source maps is not equal to count of elements");
+            }
 
-            std::for_each(data.values.begin(), data.values.end(), refract::AppendDecorator<T>(element));
+            typedef std::vector<SectionInfo< typename T::ValueType> > ValueSectionInfoCollection;
+            ValueSectionInfoCollection valuesSectionInfo = Zip<ValueSectionInfoCollection>(data.values, data.valuesSourceMap, MakeSectionInfo_<typename T::ValueType>(hasSourceMap));
```
pksunkara: Did you mean to use `MakeSectionInfo` instead of `MakeSectionInfo_` here?

src/RefractDataStructure.cc
pksunkara: If this is used someplace, I think this should have a better name and please move it to be with other section info stuff.

src/RefractDataStructure.cc
```diff
@@ -500,8 +620,37 @@ namespace drafter {
...
+        refract::StringElement* element = new refract::StringElement;
+        std::string description;
+        Join join(description);
+
+        for_each(data.descriptions.begin(), data.descriptions.end(), join);
+
+        element->set(description);
+
+        snowcrash::SourceMap<std::string> sourceMap;
+        typedef typename std::vector<snowcrash::SourceMap<std::string> >::const_iterator iterator;
```
pksunkara: Use capital for new type `iterator`

src/RefractDataStructure.cc
```diff
@@ -514,51 +663,69 @@ namespace drafter {
...
+            element->meta[SerializeKey::Description] = DescriptionToRefract(data);
         }
 
-        SetElementType(element, value.valueDefinition.typeDefinition);
+        SetElementType(element, value.section.valueDefinition.typeDefinition);
 
-        std::for_each(value.sections.begin(), value.sections.end(), ExtractTypeSection<T>(data, value));
+        SectionInfoCollection<mson::TypeSections> typeSections(value.section.sections, value.sourceMap.sections);
+
+        std::for_each(typeSections.begin(), typeSections.end(), ExtractTypeSection<T>(data, value));
```
pksunkara: As I said above, these 2 lines can possibly replaced by a macro

src/RefractDataStructure.cc
```diff
@@ -514,51 +663,69 @@ namespace drafter {
...
         return element;
     }
 
     template <typename T>
-    refract::MemberElement* RefractElementFromProperty(const mson::PropertyMember& property, const mson::BaseTypeName defaultNestedType)
+    refract::MemberElement* RefractElementFromProperty(const SectionInfo<mson::PropertyMember>& property, const mson::BaseTypeName defaultNestedType)
     {
         refract::MemberElement* element = new refract::MemberElement;
-        refract::IElement* value = RefractElementFromValue<T>(property, defaultNestedType);
+        refract::IElement* value = RefractElementFromValue<T>(SectionInfo<mson::ValueMember>(property.section, property.sourceMap), defaultNestedType);
```
pksunkara: Could you have used `MakeSectionInfo` here?

src/RefractDataStructure.cc
```diff
@@ -514,51 +663,69 @@ namespace drafter {
...
+
+            SetElementType(element->value.first, property.section.name.variable.typeDefinition);
         }
         else {
             throw snowcrash::Error("no property name", snowcrash::MSONError);
         }
 
-        mson::TypeAttributes attrs = property.valueDefinition.typeDefinition.attributes;
+        mson::TypeAttributes attrs = property.section.valueDefinition.typeDefinition.attributes;
+        // there is no source map for attributes
```
pksunkara: Are we sure about this?

src/RefractDataStructure.cc
```diff
@@ -587,18 +757,21 @@ namespace drafter {
...
-            }
+        SectionInfoCollection<mson::TypeSections> typeSections(property.section.sections, property.sourceMap.sections);
+
+        for (SectionInfoCollection<mson::TypeSections>::const_iterator it = typeSections.begin() ; it != typeSections.end(); ++it) {
+           if (it->section.klass == mson::TypeSection::BlockDescriptionClass) {
+               if (addNewLine) {
+                   descriptionRef.append("\n");
+                   addNewLine = false;
+               }
+               join(it->section.content.description);
```
pksunkara: Please add empty line before this.

src/RefractDataStructure.cc
```diff
@@ -756,45 +931,47 @@ namespace drafter {
...
+            element->meta[SerializeKey::Id] = PrimitiveToRefract(MakeSectionInfo(ds.section.name.symbol.literal, sourceMap, ds.hasSourceMap()));
         }
 
         ElementData<T> data;
 
-        std::for_each(ds.sections.begin(), ds.sections.end(), ExtractTypeSection<T>(data, ds));
+        SectionInfoCollection<mson::TypeSections> typeSections(ds.section.sections, ds.sourceMap.sections);
 
-        TransformElementData<T>(element, data);
+        std::for_each(typeSections.begin(), typeSections.end(), ExtractTypeSection<T>(data, ds));
```
pksunkara: As I said before, could have used macro here. Please check other places where the same macro could be used since I am going to stop pointing this part out.

src/RefractSourceMap.cc
```diff
@@ -0,0 +1,11 @@
+#include "RefractSourceMap.h"
+
+refract::IElement* drafter::BytesRangeToRefract(const mdp::BytesRange& bytesRange)
```
pksunkara: What is the difference between this and `SourceMapRowToRefract` from `RefractParseResult.cc`? Infact, this is wrong since we need to use `CharactersRange`. Maybe we should move that function to here?

src/RefractSourceMap.h
```diff
@@ -0,0 +1,73 @@
...
+
+    refract::IElement* BytesRangeToRefract(const mdp::BytesRange& bytesRange);
+
+    template<typename T>
+    refract::IElement* SourceMapToRefract(const T& sourceMap) 
+    {
+        refract::ArrayElement* element = new refract::ArrayElement;
+        element->element(SerializeKey::SourceMap);
+        element->renderType(refract::IElement::rFull);
```
pksunkara: This already gets overridden by line no 28.

src/RefractSourceMap.h
pksunkara: I don't understand what's the difference between this and `SourceMapToRefract` from `RefractParseResult.h`. Why do we need `template` here? I think this function is unnecessarily more complex. Please move the other function here and use it.

pksunkara:
@klokane I have finished my review. A major issue I found is that you have duplicated `SourceMapToRefract` stuff. It's probably that way because you were working on this while I finished that function. Please use my functions since they use `CharactersRange` and move them if necessary.

src/RefractSourceMap.h
klokane: All non-template functions (in this caseBytesRangeToRefract()). 
All others are template functions. If you want them get into `.cc` file so please write there prefered solution. You can not "just" put them into `.cc` until C++11 

Try it yourself!

src/Render.cc
klokane: OMG `ExpandRefract()` delete input element, but just in case while expansion happens?! It is really ugly behavior!!!

src/Render.cc
klokane: That is strange:
```
if (something) return schema;
return "same schema"?
```

The condition is useless...

But you are right it is in original code.

src/RefractSourceMap.h
pksunkara: What? Did you reply to the wrong comment? This is almost the same code as `SourceMap` functions in `RefractParseResult.*` files. So, I was asking you to move that code here and remove this part.

pksunkara: And we don't need this function to be a template. Since, there is only 1 type.

src/Render.cc
pksunkara: > The condition is useless...

I left it there, because we will be using it when we are rendering JSON Schema. Check how `renderPayloadBody` is.

src/Render.cc
pksunkara: > delete input element

It instead of modifying the input element and returning it (which it was doing previously), it now deletes the input element and returning a new element. That way, it is a bit clear.

If you really think this is wrong, please make a separate PR for this later. Let's not mix things up.

src/RefractSourceMap.h
klokane: Ah sorry, I did not read carefully I understand, you want to be visible just declarations of functions.

About "there is just 1 type", no every one `SourceMap<>` is indvidual type, but you are right the one comon ancestor.

Will check what can I use from this, while I start work on SourceMaps It was not accessible in drafter

src/Render.cc
klokane: No it is not clear. **if you did not allocate memory (or create instance of object), you never should not release it.** 
It is responsibility of context where it was created. This is typical one of reason why C programs failure on memory access.

src/Render.cc
klokane: That is funny, when I have in code similar "hook to future" or comment. You want to remove it :)

src/SerializeAST.h
klokane: To be clear it is not part or drafter itself.

src/Render.cc
klokane: Modify element is OK, because you "say it" by non-const pointer, If you wan to say, "don't modify this parameter" you will use const pointer/reference/value.

src/RefractDataStructure.cc
klokane: It is similar "hook to future" like "return schema"

src/RefractAPI.cc
klokane: I don't understand to context. It probably comes while merge. I do not use "CreateArrayElement" anewhere.

src/RefractDataStructure.cc
klokane: Probably not, It stay there while I testing something and I forgot to remove it.

src/RefractDataStructure.cc
klokane: Will change, snowcrash::Error was not there when i start work on

src/Serialize.h
klokane: Sure. I start in `Serialize.h` while it was small class, whule develop it grow up

src/Serialize.h
klokane: I mean yes, because it can collide to global `MakeSectionInfo`

src/RefractAPI.cc
klokane: Yes, it is "residuum" from conversion i forgot to remove it

src/RefractAPI.cc
klokane: I found it, you are right I create this function. It is override to template function for special case when it receive refract element. 

But I still don't understand what do you mean by:
> Please use that by adding a parameter.

klokane:
Damn, too many comments about white spacing are really disturbing. 
Found real problems in that ballast is really exhausting...

src/RefractDataStructure.cc
klokane: Where? line 720? It is followed by another line of code. Set Element Value, and Set Type of Element are different operations.
Can you please explain why there is "extra line"

src/RefractDataStructure.cc
klokane: Same as above, diferent context.

src/RefractAPI.cc
klokane: I do not see problem to call it Type. It is better to know it is type later in code in first look.

src/RefractAPI.cc
kylef: @klokane I think Pavan was saying you could add the `bFull = false` parameter to the existing `CreateArrayElement` function.

src/RefractAPI.cc
klokane: Diferent context, setting value/meta/attributes

src/RefractDataStructure.cc
klokane: No this really does not work, if you remove `typename` compiler will return error:
```
error: missing 'typename' prior to dependent type name 'T::ValueType'
```

Please check things before you write them into comments.

src/RefractDataStructure.cc
klokane: Same as above

src/RefractAPI.cc
klokane: @kyle I know about it, but this is special case where `CreateArrayElement` will receive as parameter `refract::IElement*` original templatized function is not able to handle this case.

src/Render.cc
klokane: different context, receive result vs release allocated memory.

src/RefractDataStructure.cc
klokane: There is explaining comment line below, I can swap explaining comment and this line. But this commented lines should stay in code

src/RefractAPI.cc
klokane: Yes, I will replaceit.
It happens because I start with doing it "by hand". Later I created function `LiteralToRefract` because repeating code and I forget to change it there.

src/RefractAPI.cc
klokane: Tests pass without additional part of condition which you suggest . Have you example for this case?

src/RefractAPI.cc
klokane: Why?
- original code does not contain additional check. 
- current set of tests pass without part you require.

Are you able to deliver test to confirm your comment?

src/RefractAPI.cc
klokane: Why?  There are much longer lines than this one. eg:
https://github.com/apiaryio/drafter/pull/148/files#diff-2ab1b5da6ee8cb5b29287e4ebfd30e94R243

FYI: linked line was longer than this one, before i made change on it.

src/RefractAPI.cc
klokane: There is no difference between output before and after new code is applied.
Will clear comment.

src/RefractAPI.cc
klokane: Because I did not implement it. Will fix

src/RefractAPI.cc
klokane: I mean it is not good idea

Take a look at original code:

``` 
SectionInfoCollection<snowcrash::Actions> actions(resource.section.actions, resource.sourceMap.actions);
std::transform(actions.begin(), actions.end(), std::back_inserter(content), ActionToRefract);
```
As input param to macro you need:
- type of collection `snowcrash::Actions`
- input collection `resource.section.actions`
- input sourcemap `resource.sourceMap.actions`
- output colection `content`
- transformation function `ActionToRefract`

So macro will looks like:
```
TRANSFORM_MACRO(snowcrash::Actions, resource.section.actions, resource.sourceMap.actions, content, ActionToRefract);
```

We can simplify duplicity fetching parts `section` and `sourceMap` like in `MAKE_SECTION_INFO()`

But it will convert macro into form:
```
TRANSFORM_MACRO(snowcrash::Actions, resource, actions, content, ActionToRefract);
``` 

But it absolutely does not explain what happens there. Two lines of original code are better readable than this macro.

Possible solutions I see:
- add macro like `MAKE_SECTION_INFO()` for collections, or
- replace macro `MAKE_SECTION_INFO()` by macro usable for both `SectionInfo` and `SectionInfoCollection`

```
#define SECTION_INFO(from, part) from.section.part, from.sourceMap.part
```

src/RefractAPI.cc
klokane: It is not so simple as you propose.

- reference must be set in place where it is defined
-`sourceMap` is defined as const reference because:
    - `SectionInfo<snowcrash::Elements>::NullSourceMap()` returns const reference.
    - element.sourceMap.content.elements() is too const reference (because input `const SectionInfo<snowcrash::Element>& element `

There are possible solution: 
- use const pointer to set one of reference via `if/else`and later it dereference in `SectionInfoCollection` constructor
- extract if/else part into function

Which solution do you prefer? 

Please test changes that you propose before you write them into PR review!

src/RefractDataStructure.cc
klokane: There is explaining comment above

src/RefractDataStructure.cc
klokane: No I can't because there is downcast from `mson::PropertyMember` to `mson::ValueMember`

For same reason I can not use `property' itself.

src/RefractDataStructure.cc
klokane: What I found i wrote in comment. There is sourceMap for full `mson::ValueDefinition` for `typeSection` part i did not found.
If you know where is sourceMap for typeAttributes link it please.

src/RefractAPI.cc
klokane: same as comment above, will fix

src/RefractSourceMap.cc
klokane: Hmm so something is really wrong because 
```
    struct SourceMapBase {
        mdp::BytesRangeSet sourceMap;
    };
```

- there is wrong defined `SourceMapBase` or 
- conversion from `ByteRange` to `CharacterRange` is missing everywhere in drafter

src/Render.cc
pksunkara: I understand. Can we please fix this in a later PR?

src/SerializeAST.h
pksunkara: We are not exactly following this convention in other places from Drafter.

src/RefractSourceMap.cc
pksunkara: We only use CharactersRange for source maps. We do not use BytesRange.

src/RefractDataStructure.cc
```diff
@@ -10,8 +10,50 @@
...
+        typedef V ValueType;
+        ElementType*& element;
+
+        Append(ElementType*& e) : element(e)
+        {
+        }
+
+        void operator()(const NodeInfo<ValueType>& value)
+        {
+            //throw std::logic_error("Can not append to primitive type");
```
pksunkara: I thought this comment was from old code somewhere. Let's use the error format we are following `snowcrash::Error` instead of `logic_error` then.

src/RefractDataStructure.cc
pksunkara: I actually meant `typedef typename ElementType::ValueType ValueType` since I thought once we typedef'd `T`, we don't use it anymore as a convention. I will leave this to you.

src/RefractAPI.cc
pksunkara: When I said adding a parameter, I meant something like the following function signature

```cpp
    template <typename T>
    refract::ArrayElement* CreateArrayElement(const T& content, bool rFull, bool isContentElement)
```

But that's not important. Important thing is duplication of code. Can we have something like the following?

```cpp
    namespace {
        refract::ArrayElement* CreateArrayElement(refract::IElement* value, bool rFull = false)
        {
            refract::ArrayElement* array = new refract::ArrayElement;

            if (rFull) {
                value->renderType(refract::IElement::rFull);
            }

            array->push_back(value);
            return array;
        }
    }

    template <typename T>
    refract::ArrayElement* CreateArrayElement(const T& content, bool rFull)
    {
        refract::IElement* value = refract::IElement::Create(content);
        return CreateArrayElement(value, rFull);
    }
```

src/RefractDataStructure.cc
pksunkara: I am not sure about this. I will leave the decision to you.

src/RefractAPI.cc
pksunkara: If you had changed the other line, then that's okay with me. It's just it felt weird that line was on a single line but this line which is smaller than that is on multiple lines.

src/RefractAPI.cc
pksunkara: How about something like this?

```cpp
#define NODE_INFO(from, member) from.section.member, from.section.member

template<typename T>
void SomeFnName(const T& node, const SourceMap<T>& sourceMap, RefractElements content, Functor &wrapper) {
    NodeInfoCollection<T> collection(node, sourceMap);
    std::transform(collection.begin(), collection.end(), std::back_inserter(content), wrapper);
}
```

Use it as

```cpp
SomeFnName(NODE_INFO(resource, actions), content, ActionToRefract);
```

You can also modify the `MAKE_NODE_INFO` macro to use `NODE_INFO`

src/RefractAPI.cc
pksunkara: Ah, I did not see the reference. Sorry.

src/RefractAPI.cc
klokane: Ok good idea, will change as you suggest

src/RefractAPI.cc
klokane: Yes it is what i tried to describe in comment above :) Will change

klokane: Ahh, i didn't realize you suggest additional function, will check if it is possible  (but probably yes like `RemoveElements`)

src/RefractDataStructure.cc
klokane: Few comments above you suggest reverse replacement
https://github.com/apiaryio/drafter/pull/148#discussion-diff-42248516

We should select one version and stay consistent

src/Render.cc
klokane: Sure, will you create fix, or should I have to add issue for it?

klokane: Sure. Will you make this fix? Or I have to create issue for this?

src/RefractDataStructure.cc
klokane: Ok will change as you suggest. And swap position explaining comment with commented line. We discuss more times same topic.

src/RefractSourceMap.cc
klokane: In this case it is wrong definition in `snowcrash` because code snippet above, comes from `MSONSourceMap.h` and should be fixed.

src/RefractAPI.cc
klokane: Seems good, will implement

src/RefractAPI.cc
klokane: That is reason why C++ allows override function.
Good idea, to use call overriden inside template, will implement

src/RefractAPI.cc
klokane: Ehm, this function is already there:
https://github.com/apiaryio/drafter/blob/klokane/source-maps-rebase/src/RefractAPI.cc#L68

klokane: Hmm but nut fully usable, because sometime is required to push some content to container before  "element collection".

src/SerializeAST.h
klokane: I have look at your propose about `drafter::BlueprintWrapOption` , but I do not like idea mix together independent options (ast type, expansion, sourcemap) to one enum.

src/config.cc
```diff
@@ -33,6 +35,7 @@ void PrepareCommanLineParser(cmdline::parser& parser)
     parser.add<std::string>(config::Format,    'f', "output format of the AST (yaml|json)", false, "yaml", cmdline::oneof<std::string>("yaml", "json"));
     parser.add<std::string>(config::Type,      't', "type of the AST (refract|ast)", false, "refract", cmdline::oneof<std::string>("refract", "ast"));
     parser.add<std::string>(config::Sourcemap, 's', "export sourcemap AST into file", false);
+    parser.add(config::RefractSourceMap,       'r', "produce sourcemap for refract AST");
```
pksunkara: Couldn't we merge `r` and `s` option? I don't like two options for basically the same thing.

src/config.cc
```diff
@@ -73,16 +81,34 @@ void ParseCommadLineOptions(int argc, const char *argv[], /* out */Config& conf)
...
+    conf.output           = parser.get<std::string>(config::Output);
+    conf.sourceMap        = parser.get<std::string>(config::Sourcemap);
+    conf.refractSourceMap = parser.exist(config::RefractSourceMap);
+
+
+    //  source map generating logic 
+    //  (s - sourcemap, r - refract sourcemap)
+    //  ast type: | AST | REFRACT
+    //  ----------+-----+---------
+    //  opt: -s   | s,r | r
```
pksunkara: I don't understand this line. According to the logic below, I thought Refract supports `s,r` and AST supports `s` only.

src/config.cc
klokane: we disscus it with @zdne and decide, there should be new commandline option.
while you produce sourcemap for refract it is generated directly into refract ouput, there is no extra file with sourcemap. So it is strange to have option `-s` with `sourcemap`. 
Yes, there is possibility to do not generate source map, but it is non consistent behavior. 
Because behavior of one option change behavior of other option. 

So we decide for additional switch.

src/config.cc
klokane: Hmm, try to better explain in comment, but:
- Refract supports just `-r`
- AST should supports `-s`, but expansion of MSON part should happen in same way as in refract, so expansion of refract part is added "automatic" into internal config in this case

klokane:
@pksunkara last two commits reviewed. it is Ok

pksunkara:
@klokane All that is left is to make small snippets of tests for source maps testing each and every part of the blueprint feature. And any fixes if needed.

klokane:
@pksunkara postponed into next week

PR #150, opened 2015-10-19
Port over tests from Boutique.js

danielgtaylor:
This ports the tests over from [Boutique.js](https://github.com/apiaryio/boutique.js) by doing the following:

* Create an `.apib` file for each MSON sample with a single action whose attributes equal the MSON sample.
* Create a `.json` refract parse result output that includes the element tree, rendered JSON body (if available) and rendered JSON schema (if available).
* Creates a test file similar to the ones already present to test each of the files.

This approach doesn't require any new logic in the test utilities, since you are just comparing the parsed result to the input blueprint, like all the other tests. Thoughts? Is there a better approach?

Edit: obviously the tests will fail, as these features are not yet implemented. Whoever implements them can use this branch as a starting point!


klokane:
:+1:

pksunkara:
Can you please rebase this over master? Thanks.

klokane:
@pksunkara it is not intended to merge! It is "just" portion of tests ported from `boutique.js`. 
If I understand correctly anyone can use thas as "test DB" to feel free to pick them while implementing json schema.

danielgtaylor:
This is now updated to the latest master and the new test format using `TEST_REFRACT` macros.

w-vi:
I suggest to close this one as it doesn't make sense to merge it and  all test are now ported to the JSON Schema PR https://github.com/apiaryio/drafter/pull/191

danielgtaylor:
Closing in favor of #191.

Issue #134, opened 2015-10-06
Better output from failing integration tests


klokane:
Now is reported just fixture file who which failed. eg. 
```
  REQUIRE( FixtureHelper::handleBlueprintJSON("test/fixtures/api/resource", drafter::RefractASTType) )
with expansion:
  false
```

We need add output which shows diff between fixture and current output instead of "diffing by hand". 

There is a few C/C++ libs which allow this improvement eg. DTL https://github.com/cubicdaiya/dtl

klokane:
Closed by #152

PR #152, opened 2015-10-20
Refactor tests for consistency

danielgtaylor:
This is my attempt to refactor the Drafter test fixtures and test code. Most of it was left alone or minimally modified, so the diff below is mostly from files being moved around a little. What this does:

* Reorganize test case filenames like the following:
  * `.apib` - the input API Blueprint
  * `.json` - the output Refract parse result
  * `.sourcemap.json` - the output Refract parse result with source maps (not used yet)
  * `.ast.json` - the output AST parse result
  * `.ast.sourcemap.json` - the output AST parse result with source maps
* Move the top level `blueprint.apib` test fixtures into their own folder, just like the other fixtures.
* I did **not** touch the expanded MSON tests, as those would not be useful in an external test fixture repository and they are very implementation-specific.
* I also did **not** do anything special for the C-interface tests, as those are specific to the Drafter library.
* Create some new macros to remove the need for boilerplate code
  * Test descriptions are almost identical to what they were before, but may contain `-` characters due to the difficulty of working with strings in C++ and macros.
  * Tags are still created as before. The only change is that the `parse_result` tag is now `parse-result`.
* Test failures now output the parsed result, so you can take that and run a diff against the fixture.

In the end, this means that many of the test files look like this:

```cpp
#include "draftertest.h"

using namespace draftertest;

TEST_REFRACT("api", "description");
TEST_REFRACT("api", "metadata");
TEST_REFRACT("api", "resource-group");

// ...

TEST_REFRACT("parse-result", "simple");
TEST_REFRACT("parse-result", "warning");

// ...

TEST_AST("render", "simple-object");
TEST_AST("render", "content-type");
TEST_AST("render", "nested-object");

// ...

TEST_AST_SOURCE_MAP("result", "blueprint");
TEST_REFRACT("result", "blueprint");
```

These macros expand like so:

```cpp
TEST_REFRACT("api", "description");

// Becomes:

TEST_CASE("Testing refract serialization for api description", "[refract][api]") {
  FixtureHelper::handleResultJSON("test/fixtures/api/description");
}
```

With these changes I believe that:

1. Test consistency will mean easier maintenance
2. Test organization means the fixtures could go into a stand-alone git repository to be shared
3. Test fixtures could now be easily re-used by the pure JS Drafter
4. Test fixtures could now be easily re-used by other parsers or tools, e.g. taken as a good set of test inputs to run through your tool when testing.


klokane:
Like idea, but too big patch I have received message
"Sorry, we could not display the entire diff because it was too big." while try to see it :)

klokane:
Can you please split this patch into more commits? 
At least to part changing fixtures, and source code itself?

And if you want be really nice, you can integrate some "diffing" library, to see diffs directly in test output. eg. https://github.com/cubicdaiya/dtl related to #134

danielgtaylor:
@klokane I have updated the code to split it into multiple commits and have added diff support. It's best to ignore whitespace (by adding `?w=0` to the URL) when looking at the diff, e.g:

Third commit: https://github.com/apiaryio/drafter/commit/6c1010b02c8e9837da04118bdab66ea437049fa6?w=0

The output on a failure will look like this now (colorized in a terminal):

```
-------------------------------------------------------------------------------
Testing refract serialization for parse-result simple
-------------------------------------------------------------------------------
../test/test-RefractParseResultTest.cc:13
...............................................................................

../test/draftertest.h:160: FAILED:
explicitly with messages:
  Filename: test/fixtures/parse-result/simple.json
  @@ -14,7 +14,7 @@
             "element": "category",
             "meta": {
               "classes": [
  -              "resourceGroup"
  +              "resourceGroup2"
               ],
               "title": ""
             },
  @@ -38,7 +38,7 @@
                         "element": "httpTransaction",
                         "content": [
                           {
  -                          "element": "httpRequest",
  +                          "element": "httpReques",
                             "attributes": {
                               "method": "GET"
                             },
  

===============================================================================
test cases: 89 | 88 passed | 1 failed
assertions: 83 | 82 passed | 1 failed
```

screen shot 2015-10-21 at 13 05 12

klokane:
@danielgtaylor wow great, thank you. 
C/C++ code and macros looks good. There are a few "nitpicks" about it. 
On morning I'm going to check tests.

klokane:
@danielgtaylor please use "git submodule" instead of directly copy file.

danielgtaylor:
@klokane updated to move string extensions into constants. I'm not sure I can use the submodule... the original `dtl` module does no coloring of the output and is hard to read, so I modified a few constants to add the standard diff colored output. Sadly it isn't configurable so either I need to copy the code or write my own templated printer function (which I may be able to hack together but I'm not sure).

klokane:
@danielgtaylor it is BSD licensed so we can fork it and hold separated repository with our own changes.
I mean it is better than copy files directly. 

Especially in case while it is used "just" for test output and testing library is linked into project via submodule.

kylef:
@danielgtaylor Perhaps we could offer a pull request back upstream to add an option for these colour outputs in `dtl`?

klokane:
@kylef good idea. if will be accepted, we can later link submodule back to original repository.

klokane:
@danielgtaylor reviewed. Last thing to resolved is `dtl` as submodule. Will discuss with @zdne today.

klokane:
@danielgtaylor so I discuss it with Z. Result is 
- make fork of `dtl` 
- commit your changes (color output) on top, 
- link as submodule into drafter
- optionaly pull request your change back on dtl

danielgtaylor:
@klokane this is now using a submodule from https://github.com/apiaryio/dtl

I have also opened an issue on the original repo about colorizing the output: https://github.com/cubicdaiya/dtl/issues/5

zdne:
👍

klokane:
@danielgtaylor great.
I'm sorry on yesterday I miss it, but I found last "problematic" place. 
BIG_LETTERS_WITH_UDERSCORES are in C/C++ by convention used for macros. 
Please change string for file extensions to CamelCaseNames (maybe enclosed into namespace eg. `extension` or `suffix` or something you like)

danielgtaylor:
@klokane updated to use CamelCasing. Wasn't quite sure what you mean with the namespaces so I just changed the casing and left everything else as-is!

klokane:
@danielgtaylor I mean something like
```
namespace ext {
  const std::string json = ".json";
...
}
```

and later
```
filename += ext::json;
```

But it is not important for now. Because it is just in one file. But it is good to do it while it is thru multiple files. As example you can look on to `SerializeKey`

klokane:
@danielgtaylor can you please fix comment above? https://github.com/apiaryio/drafter/commit/2608322969c505339d39ede6b1d0934802328e56#commitcomment-13945886

PR #156, opened 2015-10-29
Make a prerelease

pksunkara:
This PR marks a prerelease for drafter so that Pure JS drafter can be released for people to experiment with.

This PR also adds a PRE_RELEASE_VERSION so that we can have multiple pre releases and follow semver rules.


kylef:
Would `npm install drafter.js` install a pre-release or is there some flag we need to add to the instructions in the README too? I know for RubyGems you'd need to add the `--pre` flag to allow prerelease.

danielgtaylor:
@kylef @pksunkara if you `npm publish` without a specific tag (which is what the automated release process does) then it will be installed when you `npm install drafter.js`. If you want to control the release manually then you can `npm publish` yourself (e.g. with `--tag pre` or something) before pushing the tag to GitHub, which will cause the publish from CI to fail. I think just letting CI do the release should be fine since this is the first release and we don't have to worry about breaking anybody.

pksunkara:
@danielgtaylor I think kyle is asking about what happens when the npm module has only 1 prerelease tag. Will an user install prerelease version? Or will npm complain?

danielgtaylor:
@pksunkara npm doesn't care what the version number is, only what `--tag` option you publish with. If you don't use a `--tag` option then the version (no matter what it is) will be installed with `npm install drafter.js`. Doesn't matter if it has `-pre` in the name as far as I know.

pksunkara:
Thanks @danielgtaylor. @kylef Can we merge now?

kylef:
:+1:

PR #157, opened 2015-10-29
Fix upload asset name

danielgtaylor:
Fixes the release issue from the latest pre-release. There was a bug in the script where it tried to upload two files to the same name.

Both files are now uploaded to the latest release; this fix will apply to the next release:

https://github.com/apiaryio/drafter/releases/tag/v2.0.0-pre.0

Thanks!


emcc/release.sh
```diff
@@ -9,7 +9,7 @@ tar -xjf $OS-amd64-github-release.tar.bz2
 
 ./bin/$OS/amd64/github-release release -u apiaryio -r drafter --tag $TAG
```
pksunkara: Do we need to check for `pre` in TAG and mark the github release as pre release?

pksunkara:
Just one non-blocking comment.

pksunkara:
I have realized that the number of pre-releases are going to be very rare.

PR #160, opened 2015-11-02
Fix memory allocation - emscripten

IvanGoncharov:
Lib crashed when I try to parse specs with Unicode symbols in them.
I debug issue and took solution from here https://github.com/kripken/emscripten/blob/42fb486c53d627b203b77af6e5d0c088c0ad03ad/src/preamble.js#L158
Please add Unicode tests to your testsuite.

### DIFF
--- a/emcc/post.js
+++ b/emcc/post.js
@@ -16,7 +16,8 @@ Module['parse'] = function(blueprint, options) {
   }
 
   var chptr = _malloc(4);
-  var buffer = _malloc(blueprint.length+1);
+  // at most 4 bytes per UTF-8 code point, +1 for the trailing '\0'
+  var buffer = _malloc(4*blueprint.length+1);
   var parseOptions = 0;
   var astType = 1;
 


pksunkara:
There seems to be an issue with CircleCI. I am currently looking into it.

IvanGoncharov:
@pksunkara it failed because it can't execute `npm install`. I ran into the same problem when i tested it locally. It because of [this](https://github.com/apiaryio/drafter/blob/master/.npmrc) file. NPM can find appropriate variable so it result in following error: 
> Error: Failed to replace env in config: ${NPM_TOKEN}

I resolved it locally by simply removing `.npmrc` file.

pksunkara:
@IvanGoncharov Can you rebase this on master?

IvanGoncharov:
@pksunkara Done

IvanGoncharov:
@pksunkara Can you please release new version of NPM package?

pksunkara:
@IvanGoncharov We just need to update this line to https://github.com/apiaryio/drafter/blob/master/src/Version.h#L21from `0` to `1`. Would you please make a PR?

IvanGoncharov:
@pksunkara Done

PR #163, opened 2015-11-03
Port mson tests

klokane:
Small refactoring for tests:
- include tests for MSON into new testing convetions, 
- remove duplicit code for different wrappers

@danielgtaylor  can you, please, review, it is base on your work


test/test-RefractDataStructureTest.cc
```diff
@@ -4,187 +4,47 @@ using namespace draftertest;
...
+TEST_MSON("array-typed-content");
+TEST_MSON("resource-nested-member");
+TEST_MSON("array-sample");
+TEST_MSON("object-sample");
+TEST_MSON("enum-members-description");
+TEST_MSON("array-reference");
+TEST_MSON("reference-override");
+TEST_MSON("enum-variants");
+
+#undef TEST_MSON
```
pksunkara: What about `#undef TEST_MSON_`. And can we have a better name for `TEST_MSON_`?

test/draftertest.h
```diff
@@ -12,20 +12,25 @@
 #include "SerializeResult.h"
 #include "sosJSON.h"
 
+
+#define TEST_DRAFTER(description, category, name, tag,  wrapper, options, mustBeOk) TEST_CASE(description " " category " " name, "[" tag "][" category "]") { \
+    REQUIRE(FixtureHelper::handleResultJSON(wrapper, "test/fixtures/" category "/" name, options, mustBeOk)); \
+}
```
pksunkara: We could have actually had `TEST_MSON` here itself. Why are we treating it specially when compared to other macros?

pksunkara:
Looks good. Just a few things.

test/test-RefractDataStructureTest.cc
klokane: Oh good catch, there were original just `TEST_MSON`. Macro `TEST_MSON_` I add later to allow failing parser output.

test/draftertest.h
klokane: I mean, we should do it in same way for other tests.

There is no reason to have specific macro accessible for other test files, it just filling "global namespace"

pksunkara: Okay, we can do that later then. Please fix the other comment and I can merge this.

test/test-RefractDataStructureTest.cc
klokane: If you don't like name, propose some better :)

pksunkara: `TEST_MSON_SUCCESS`? Just having an underscore at the end looks unreadable.

danielgtaylor: I think `TEST_MSON_SUCCESS` makes sense.

danielgtaylor:
Looks good to me, just that one comment :+1:

pksunkara:
@klokane Let's merge this tomorrow.

PR #175, opened 2015-11-12
Remove version matrix from appveyor

kylef:
With the version matrix, appveyor is not updating PR statuses correctly.


kylef:
Looks like this wasn't the cause.

PR #180, opened 2015-11-23
Fix memory leaks

klokane:
Fixed memory leaks on
- create description
- double registration of Named Types in AST branch

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ Other systems refer to [build notes](#build).
 
 ### C++ library
 ```c++
-#include "drafter.h"      // Blueprint Parser
-#include "SerializeAST.h" // Blueprint Result Wrapper for serialization
-#include "sosJSON.h"      // Serializer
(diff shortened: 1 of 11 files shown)


kylef:
I have the one question, but otherwise the changes look good to me. Would appreciate if someone else could give a :+1: too.

:+1:

w-vi:
The changes look good but I have to admit that I don't see where the memory leak actually was. @klokane

klokane:
@w-vi second one is more complex, But in general while was call `WrapResult()` with `AST` type, it come later to `WrapBlueprint()`  

They both made `RegisterNamedTypes()` and no where was checked if named object is already registrated. 
So I split it into two diferent "run branches" and clarify interface by removing unneeded functions

w-vi:
@klokane OK now I get it ... nice job

pksunkara:
:+1:

PR #181, opened 2015-11-26
Add convenience PrintVisitor

w-vi:
PrintVisitor is a helper visitor, it just prints the refract element and
its sybelements to any `ostream` provided. Default is `std::cout`

### DIFF
--- a/drafter.gyp
+++ b/drafter.gyp
@@ -57,6 +57,7 @@
 
         "src/refract/Visitor.h",
         "src/refract/Visitors.h",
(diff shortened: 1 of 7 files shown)


w-vi:
@klokane please have a look, if it is good enough for starters.

klokane:
Nice. Just please use `camelCase` naming convention for PrintVisitor methods

klokane:
And do not use underscore prefixed members. Just formality, but wide used

w-vi:
@klokane I'm done

PR #179, opened 2015-11-23
Fixes bug in `classes` key of asset element

pksunkara:
The current master branch serialises AssetElement as

```json
{
  "element": "asset",
  "classes": "messageBody"
}
```

when it should be according to the spec

```json
{
  "element": "asset",
  "classes": [
    "messageBody"
  ]
}
```

Issue #141, opened 2015-10-08
Regression: Member reference is not getting rendered


pksunkara:
# Example

```apib
FORMAT: 1A

# Example

# Data Structures

## B (object)
+ b: b

## A (object)
+ a (B)

# GET /
+ Response 200 (application/json)
    + Attributes (A)
```

# Old behaviour
Response body is getting rendered as

```json
{"a": {"b": "b"}}
```

# New behaviour
Response body is getting rendered as

```json
{"a": {}}
```

# Expected
Old behaviour is correct.

Issue #137, opened 2015-10-08
Regression: Mixins are not getting rendered


pksunkara:
# Example

```apib
FORMAT: 1A

# Example

# Data Structures
## A (object)
+ Include B

## B (object)
+ b: b

# GET /
+ Response 200 (application/json)
    + Attributes (A)
```

# Old behaviour
Response body is rendered as

```json
{"b": "b"}
```

# New behaviour
Response body is an empty JSON object.

# Expected
Old behaviour is correct.

Issue #136, opened 2015-10-08
Regression: Value for enum is not getting rendered


pksunkara:
# Example

```apib
FORMAT: 1A

# Example

# Data Structures
## A
+ a: TAC (B, required)

## B (enum)
- TAC

# GET /
+ Response 200 (application/json)
    + Attributes (A)
```

# Old behaviour

Response body is getting rendered as:

```json
{"a": "TAC"}
```

# New behaviour

```json
{"a": {}}
```

# Expected

Old behaviour is correct.

klokane:
@pksunkara @zdne I've question, is prefered value `TAC` from `A` or `TAC` from `B` in this case?
It sounds weird, but I need to know from which source I should render value

pksunkara:
It's from `A`. Because `B` might be something like this:

```apib
## B (enum)
- TAC
- CAT
```

klokane:
@pksunkara sure, but `A` can be similar defined as  `+ a: TAC, CAT (B, required)`

klokane:
/Ok lets assume this one definition
```## A
+ a: A,B (B, required)

## B (enum)
- C
- D
```

which one value to select?

pksunkara:
Your example would  render `+ Attributes (A)` as `{ "a": "A,B" }`

klokane:
`+ a: A,B (enum)` should be interreted as `[ `enum`, {}, {}, ["A", "B"] ]`
So i mean
```
## B (enum)
## A
+ a: A,B (B)
```
`A.a` should be interpreted in same way, because `B` is `enum`

pksunkara:
@zdne We really **really** need your comment on this.

zdne:
@pksunkara what is the question?

klokane:
@zdne which value to pick for asset 
```
## A
+ a: A,B (B, required)

## B (enum)
- C
- D

# GET /
+ Response 200 (application/json)
    + Attributes (A)
```

zdne:
@klokane for now the rule of the thumb should be:

1. value 
2. sample(s) (if value not present)
3. default (if value or sample(s) not present)

zdne:
@klokane so question is what is the value of A

zdne:
@klokane 

and I believe we have discussed the case 

```apib
- color: red, green (enum)
   - blue
   - cyan
```

zdne:
obviously the one above does not make much sense. What does, however is:

```apib
- color: red
   - red
   - green
   - blue
```

[...]

PR #182, opened 2015-11-26
Release 2.0.0

kylef:
This is a backward-incomatible release with the following changes.

##### Breaking

- C++ interface now returns Refract ParseResult instead of API Blueprint AST by default. You can get the previous behaviour using the `NormalASTType` option.
- Command line interface has changed to by default, return the whole ParseResult under both Refract and AST output types, along with source maps.

##### Enhancements

- Support for the `nullable` MSON type attribute.  
  [#106](https://github.com/apiaryio/drafter/issues/106)
- Source maps are now available when using the Refract output type with the `exportSourceMap` option.  
  [#148](https://github.com/apiaryio/drafter/pull/148)
- We now offer a JavaScript version of drafter using emscripten. 

##### Bug Fixes

- Drafter should no longer raise exceptions and instead attach all errors to the parse result.
- Various bugs while rendering JSON from MSON have been resolved.  
  [#117](https://github.com/apiaryio/drafter/issues/117)
  [#136](https://github.com/apiaryio/drafter/issues/136)
  [#137](https://github.com/apiaryio/drafter/issues/137)
  [#141](https://github.com/apiaryio/drafter/issues/141)
- Meta `classes` from an `asset` element are now an array as per the Refract specification instead of using the `class` key.  
  [#179](https://github.com/apiaryio/drafter/pull/179)


zdne:
@kylef 

> C++ interface now returns Refract ParseResult instead of API Blueprint AST. 

I would add a note by default (IIRC there is a switch to get the previous output?)

> Source maps are now generated when using the Refract output type.

Are they now generated all the time or on demand? (add it to the notes)

Otherwise looks good!

PR #186, opened 2015-12-01
Update snowcrash to use version with cached character map

klokane:
Better performance for C++ drafter - SourceMaps are generated form cached character map instead while SourceMaps are generated


klokane:
Do Not Merge: Some problems with integration test on windows.

pksunkara:
This is done by #189

PR #192, opened 2015-12-09
Do MSON validating and rendering when validate option is supplied

### LINKED ISSUE #188: `-l` or "validate" CLI option doesn't catch MSON errors
I believe this is due to the validate option doesn't render the JSON structures and this any errors raised from those code-paths won't be raised.

This can be reproduced with the following blueprint:

```apib
# R [/r]
## Retrieve [GET]
- response 200 (application/json)
    - attributes (array[User])
        - name: aa
        - nested
            - a
            - b
            - c
# Data Structures
## User
- name: John
```

Where if you try to validate the blueprint using the drafter CLI tool it succeeds:

```shell
$ drafter y.apib -l

OK.
```

However, without the `-l` option, drafter fails with an error:

```shell
$ drafter y.apib

error: (4)  nested complex types are not implemented
```

pksunkara:
This PR fixes #188 

Only drawback is there are 2 unnecessary objects created when `-l` option is given. `sos::Serialize* serializer` and `std::ostream* out`

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -65,25 +65,26 @@ int main(int argc, const char *argv[])
     sc::ParseResult<sc::Blueprint> blueprint;
     sc::parse(inputStream.str(), options, blueprint);
 
-    if (!config.validate) {  // not just validate -> we will serialize
+    sos::Serialize* serializer = CreateSerializer(config.format);
+    std::ostream *out = CreateStreamFromName<std::ostream>(config.output);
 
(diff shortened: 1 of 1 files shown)


pksunkara:
@klokane @w-vi Please review and merge.

kylef:
:ok_hand: thanks for fixing this @pksunkara.

w-vi:
@pksunkara as a quick fix it's good to go

PR #195, opened 2015-12-11
Add files to xcodeproj and some style fixes

pksunkara:
* We use a newline char before `{` on function signature lines in `.cc` file.
* `else` should be on a separate line
* If a condition in `for` or `if` or `while` spans more than a single line, we add an empty line at the beginning of the block.
* We always use `{` & `}` symbols even if the block is a single line


pksunkara:
@w-vi Please review and merge.

PR #198, opened 2015-12-12
Refactoring the JSONSchemaVisitor

w-vi:
- DRY primitive types
- Remove unnecessary heap allocations

### DIFF
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -15,11 +15,10 @@
 namespace refract
 {
 
-    JSONSchemaVisitor::JSONSchemaVisitor()
+    JSONSchemaVisitor::JSONSchemaVisitor(bool fixit /*= false*/) : fixed(fixit)
     {
         pObj = new ObjectElement;
         pObj->renderType(IElement::rCompact);
-        fixed = false;
     }
 
     JSONSchemaVisitor::~JSONSchemaVisitor()
@@ -32,6 +31,24 @@ namespace refract
     void JSONSchemaVisitor::setFixed(bool fixit)
     {
         fixed = fixit;
+    }
+
+    template<>
+    void JSONSchemaVisitor::setPrimitiveType(const BooleanElement& e)
+    {
+        setSchemaType("boolean");
+    }
+
+    template<>
+    void JSONSchemaVisitor::setPrimitiveType(const StringElement& e)
+    {
+        setSchemaType("string");
+    }
+
+    template<>
+    void JSONSchemaVisitor::setPrimitiveType(const NumberElement& e)
+    {
+        setSchemaType("number");
     }
 
     void JSONSchemaVisitor::setSchemaType(const std::string& type)
@@ -66,6 +83,22 @@ namespace refract
         return true;
     }
 
+    template<typename T>
+    void JSONSchemaVisitor::primitiveType(const T& e)
+    {
+        const typename T::ValueType *v = GetValue<T>(e);
+
+        if (v) {
+            setPrimitiveType(e);
+
+            if (fixed && !e.empty()) {
+                ArrayElement *a = new ArrayElement;
+                a->push_back(IElement::Create(*v));
+                addMember("enum", a);
+            }
+        }
+    }
+
     void JSONSchemaVisitor::visit(const IElement& e)
     {
         e.content(*this);
@@ -76,16 +109,10 @@ namespace refract
         JSONSchemaVisitor renderer;
 
         if (e.value.second) {
-            if (IsTypeAttribute(e, "nullable") && e.value.second->empty()) {
-                renderer.visit(*e.value.second);
-            }
-            else if (IsTypeAttribute(e, "fixed") || fixed) {
+           if (IsTypeAttribute(e, "fixed") || fixed) {
                 renderer.setFixed(true);
-                renderer.visit(*e.value.second);
             }
-            else {
-                renderer.visit(*e.value.second);
-            }
+           renderer.visit(*e.value.second);
         }
 
         if (StringElement* str = TypeQueryVisitor::as<StringElement>(e.value.first)) {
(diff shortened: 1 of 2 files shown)


src/refract/JSONSchemaVisitor.cc
```diff
@@ -340,62 +336,22 @@ namespace refract
...
-            if (fixed && !e.empty()) {
-                ArrayElement *a = new ArrayElement;
-                StringElement *s = new StringElement;
-
-                s->set(*v);
-                a->push_back(s);
-                addMember("enum", a);
-            }
-        }
+        primitiveType(e, GetValue<StringElement>(e));
```
klokane: you know type sent into template, `GetValue<T>(e)` can be moved into `primitiveType<>()` body

src/refract/JSONSchemaVisitor.h
```diff
@@ -36,8 +37,28 @@ namespace refract
...
         bool allItemsEmpty(const ArrayElement::ValueType* val);
 
+        template<typename T>
+        void setPrimitiveType(const T& e)
+        {
+            static_assert(sizeof(T) == 0, "Only String, Number, Boolean types allowed");
+        };
+
+        template<typename T>
+        void primitiveType(const T& e, const typename T::ValueType *v)
```
klokane: Not sure, but I mean body of private template member function can be just "declared" in header and definition can be moved in to `cc`. we need no to expose it.

src/refract/JSONSchemaVisitor.h
```diff
@@ -36,8 +37,28 @@ namespace refract
         void enumElement(const ArrayElement& e, const ArrayElement::ValueType *val);
         bool allItemsEmpty(const ArrayElement::ValueType* val);
 
+        template<typename T>
+        void setPrimitiveType(const T& e)
+        {
+            static_assert(sizeof(T) == 0, "Only String, Number, Boolean types allowed");
```
klokane: :+1:

src/refract/JSONSchemaVisitor.cc
```diff
@@ -340,62 +336,22 @@ namespace refract
 
     void JSONSchemaVisitor::visit(const NullElement& e)
     {
-        NullElement nil;
-        addMember("type", &nil);
+        addMember("type", new NullElement);
```
pksunkara: Is this heap allocation going to be cleaned up somewhere? I would prefer the original code.

src/refract/JSONSchemaVisitor.cc
w-vi: I was getting an error something like `expected primary-expression before ‘>’ token` which I ahve to admit I don't understand .... need to dive deeper into templates :)

w-vi:
@klokane  Are we doing C++03 or C++0x? As Circle CI is failing as there is nothing like `static_assert` which is part of of the C++0x standard.

pksunkara:
We are C++03 since there is a limitation with nodejs in protagonist.

src/refract/JSONSchemaVisitor.cc
w-vi: Original code was passing a pointer to a local variable allocated on stack ... and yes there is clean up, when the Visitor's desctructor is called.

klokane: original code is wrong, it return address of object allocated on stack.

pksunkara:
Everything looks good. Please look into if you can improve the `GetValue` thing.

klokane:
@w-vi 03, but I already thing about `static_assert<>` port into code base, there is a few point where it can be usefull instead of runtime check

klokane:
@w-vi please look at https://github.com/apiaryio/drafter/blob/wvi/json-schema-cleanup/src/refract/JSONSchemaVisitor.cc#L95

there is
```
           if (...) {
                renderer.visit(*e.value.second);
            }
            else if (...) {
                renderer.setFixed(true);
                renderer.visit(*e.value.second);
            }
            else {
                renderer.visit(*e.value.second);
            }
```

It probably can be reduced into:
```
if (/*second case*/) {
  renderer.setFixed(true);
}
renderer.visit(*e.value.second);
```

klokane:
https://github.com/apiaryio/drafter/blob/wvi/json-schema-cleanup/src/refract/JSONSchemaVisitor.cc#L156

please add explaining comment why you need no check availability of  `o1->value[0]`  it is not clear from where it come
``` 
                MemberElement *m1 = TypeQueryVisitor::as<MemberElement>(o1->value[0]->clone());
```

w-vi:
@klokane So what to do about `static_assert`? Should I just put a FIXME there for now and put it back once we put `static_assert` in our codebase?

src/refract/JSONSchemaVisitor.cc
klokane: `typename T::ValueType* v = GetValue<T>(e);`

klokane:
@w-vi your decision. If you are willing to port `static_assert<>` go straight :)

klokane:
But maybe leave it for now with FIXME. We need to discuss which one code base to integrate (drafter/snowcrash/refract).

w-vi:
@klokane leaving the `static_assert` for later discussion and using the runtime approach

klokane:
@w-vi this should be probably replaced by `TypeQueryVisitor::as<ObjectElement>` or is there reason to use check against string?
https://github.com/apiaryio/drafter/blob/bef8ac702e21337a97a0c2f762c6572e6d79f4a5/src/refract/JSONSchemaVisitor.cc#L208

klokane:
All other seems ok

w-vi:
@klokane no, `TypeQueryVisitor::as<ObjectElement>` is better, copy and paste error, as there is nothing like `EnumElement` and there I need to check the string `enum` .... let's do the `EnumElement` :)

klokane:
Yeah `EnumElement` is already proposed. Hope for delivery as soon a possible :)

Issue #47, opened 2015-06-17
Nested arrays are not parsed correctly


w-vi:
The nested array object is not correctly parsed using drafter, see the example API Blueprint below.
In  Apiary I can see render only 

```json
{
  "categories": [
    {
      "name": "Category One",
      "subcategories": []
    }
  ]
}
```
instead of 
```json
{
  "categories": [
    {
      "name": "Category One",
      "subcategories":
      [
        {
          "name": "Sub category One"
        }
      ]
    }
  ]
}
```

Example Blueprint
```apib
FORMAT: 1A

# Test nested arrays-in-object-arrays

A navigation has multiple categories. Each category can have multiple subcategories.

# GET /navigation

+ Response 200 (application/json)

    + Attributes

        + categories (array)
            + (object)
                + name: Category One (string) - Name of the category
                + subcategories (array)
                    + (object)
                        + name: Sub category One (string) - Name of the subcategory
```

**EDIT by @pksunkara**: The MSON AST which is generated from the blueprint is wrong.

JeroenVdb:
Related StackOverflow: http://stackoverflow.com/questions/30890933/nested-array-with-object-in-mson-is-empty-in-apiary-io-documentation/30893537

pksunkara:
Thanks for the report @JeroenVdb. I am going to look into this now.

pksunkara:
This is a bug in the parser itself. The generated MSON AST is not correct given the MSON.

JeroenVdb:
Super @pksunkara can you give a shout here when this is fixed?
How can I see if this is then used by the apiary.io editor?

pksunkara:
@JeroenVdb Apiary should have the bug fix as soon as we release the bug fix in here.

alexturek:
@pksunkara Is [this](https://github.com/apiaryio/snowcrash/issues/330) the right snowcrash bug? Or is one filed?

pksunkara:
@alexturek No, that is a different bug. This is a drafter or drafter.js bug which needs to be fixed.

huiyiqun:
Please refer this: apiaryio/drafter.js#38

Working apib:
```apib
FORMAT: 1A

# Test nested arrays-in-object-arrays

A navigation has multiple categories. Each category can have multiple subcategories.

# GET /navigation

+ Response 200 (application/json)

    + Attributes

        + categories (array)
            + (object, fixed)
                + name: Category One (string) - Name of the category
                + subcategories (array)
[...]

PR #199, opened 2015-12-16
Fix inheritance

### LINKED ISSUE #98: MSON - named type based on primitive is wrongly rendered
While we try to define new type of Element eg. **Data** who inherit from **StringElement** it returns wrong body example

e.g
```apib
# GET /sample
    + Attributes
        + message: (Message)

# Data Structures

# Message
+ created (Date)
+ text (string)

# Date (string)

## Sample
2012-12-01
```

returns:
```
...
"body": "{\n  \"message\": {\n    \"created\": {},\n    \"text\": \"\"\n  }\n}"
...
```

Expected output is:
```
....
"body": "{\n  \"message\": {\n    \"created\":  "2012-12-01",\n    \"text\": \"\"\n  }\n}"

...
```

### LINKED ISSUE #176: Regression - enum inline definition pick into JSON "array" instead of member
```
# Data Structures
## A
+ a: X,Y (B, required)

## B (enum)
- P
- Q

# GET /
    + Attributes (A)
```

## Current
```json
{"a": ["X", "Y"] }
```

## Expected
```json
{"a": "X"}
```

Releated #136

klokane:
Reworked inheritance:
- fix correct order in `extend` element for inheritance from older ancestor to newest
- allow inheritance for primitive members
- fixed inheritance for `RenderJSONVisitor`
- introduce ` ElementMerger` to handle correctly/standardized  way inheritance resolve
- refactoring `ExpandVisitor`
- move members out of `IElement` (it now provide just Interface and some helpers)
- little refactoring in `TypeQueryVisitor` (expose enum ast typedef)
- correct order "data structures" according to membership/inheritance (most dependent objects are registered latest)

To be resolved: 
- cripled some schema test after rebase (done)
- add test for primitive types inheritance (done)
- add test for complex membership/inheritance 
(test are created, just needs to added to repository and check if schema is generated correctly) 

Resolved issues:
- fix #98
- fix #176


src/RefractAPI.cc
```diff
@@ -6,7 +6,6 @@
 //  Copyright (c) 2015 Apiary Inc. All rights reserved.
 //
 
-#include <iterator>
```
kylef: Curious, any particular reason you've moved this down? In most languages it preferred to put system includes before local includes.

klokane: In short? In C/C++ is prefered system down :)

In long:
```
// file a.h
std::string fn();
```

```
// file a.cc
#include <string>
#include "a.h"

std::string fn() { return "hello"; }
```

`a.cc` will be compiled ok

but
```
// file: b.cc
#include "a.h"

something...
```

compilation will fail with something like `unknown type 'std::string'`

kylef: Ah that's a good point. It's no longer been a problem in Objective-C with the introduction of clang modules. I completely forgot it's not standard/default across all compilers.

:+1:

klokane:
Last commit (103c0bf) is for discussion, it fix problems with schema fixtures, but fixtures definition in schema are not correct. There should be selected one value for `enum` instead of array of possible values.

klokane:
Hmm interesting, sent patch (commented out part) fixes #176 :)

klokane:
Curent statust
- @w-vi is checking changes in JSON Schema rendering
- working on fix: https://github.com/apiaryio/drafter/blob/klokane/fix-inheritance/test/fixtures/mson/primitive-variables.apib

(member `c8` should be expanded as object)

klokane:
@w-vi rebased on top of your patch, added fix for for **variable property name**
Ready for  code review

klokane:
@w-vi it I'm going thru issues. It seems, this patch fix #47 too, can you please check it and confirm or deny?

w-vi:
@klokane I have check the latest master and the issue is fixed there as well, so I don't know when it was fixed but it is fixed.

klokane:
ok

klokane:
There is something wrong with emscripten build, have no idea what happen
```
FAIL ../test/fixtures/mson/primitive-variables.apib (ast)

/home/ubuntu/drafter/emcc/drafter.nomem.js:1
aughtException",(function(ex){if(!(ex instanceof ExitStatus)){throw ex}}));Mod
                                                                    ^
TypeError: Cannot call method 'split' of undefined
    at Object.lineDiff.tokenize (/home/ubuntu/drafter/node_modules/diff/lib/diff/line.js:20:32)
    at Object.diff (/home/ubuntu/drafter/node_modules/diff/lib/diff/base.js:36:39)
    at Object.diffJson (/home/ubuntu/drafter/node_modules/diff/lib/diff/json.js:33:19)
    at /home/ubuntu/drafter/emcc/test-drafter:125:25
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:721:13
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:52:16
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:269:32
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:44:16
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:718:17
    at /home/ubuntu/drafter/node_modules/async/lib/async.js:167:37
npm ERR! Test failed.  See above for more details.
```

w-vi:
@klokane I am looking into it right now

src/refract/Element.cc
```diff
@@ -129,4 +131,245 @@ namespace refract
...
+                }
+                return true;
+            }
+        };
+    }
+
+    void ExtendElement::set(const ExtendElement::ValueType& val)
+    {
+        if (!TypeChecker(val)) {
+            throw LogicError("ExtendElement require to be composed form Elements of Equal Type");
```
klokane: use same message like line 191

src/refract/Element.cc
```diff
@@ -129,4 +131,245 @@ namespace refract
...
+                                (*item)->value.second = (*it)->value.second->clone();
+                            }
+                        }
+                    }
+                }
+            };
+
+            /**
+             * precondition - target && append element MUST BE of same type
+             * we use static_cast<> without checking type t is responsibility if caller
```
klokane: s/if/of/

src/RefractAPI.cc
```diff
@@ -82,10 +84,145 @@ namespace drafter {
...
+            }
+        }
+
+        bool hasMember(const snowcrash::DataStructure* object, const snowcrash::DataStructure* member) {
+            MembersMap::const_iterator members = objectToMembers.find(name(object));
+            if (members == objectToMembers.end()) {
+                return false;
+            }
+
+            return members->second.find(name(member)) == members->second.end();
```
klokane: swap condition to `!=` to be conform to name of function

src/refract/IsExpandableVisitor.cc
```diff
@@ -37,6 +37,13 @@ namespace refract
             }
         };
 
+        //template <>
```
klokane: remove commented code

src/refract/SerializeVisitor.cc
```diff
@@ -35,6 +35,18 @@ namespace refract
             return s.get();
         }
 
+        template <typename T>
+        sos::Array SerializeValuelist(const T& e) {
```
klokane: typo `s/SerializeValuelist/SerializeValueList`

src/refract/Element.cc
```diff
@@ -129,4 +131,243 @@ namespace refract
...
+
+                    case TypeQueryVisitor::Object:
+                        doMerge<ObjectElement>(result, e);
+                        return;
+
+                    case TypeQueryVisitor::Member:
+                        doMerge<MemberElement>(result, e);
+                        return;
+
+                    case TypeQueryVisitor::Null:
```
w-vi: shouldn't these two do something, at least throw an error or something? Or put comment why we can do nothing

src/refract/Element.cc
```diff
@@ -129,4 +131,243 @@ namespace refract
...
+            };
+
+            template <typename T>
+            struct InfoMerge {
+                IElement::MemberElementCollection& info;
+                InfoMerge(IElement::MemberElementCollection& info) : info(info) {}
+
+                void operator()(const IElement::MemberElementCollection& append) {
+                    for (IElement::MemberElementCollection::const_iterator it = append.begin() ; it != append.end() ; ++it) {
+                        if (StringElement* key = TypeQueryVisitor::as<StringElement>((*it)->value.first)) {
```
w-vi: Is this correct? It looks like it is only overwriting existing keys but new keys are just skipped.

klokane: Probably not, will fix

klokane: It should append new one, and replace old one like for values. Additionally there are some condition for some special key in meta `id`, .`namespace` and `prefix`

src/refract/Element.cc
klokane: ok will add explaining comment

w-vi:
Not really part of this PR but when doing the review I came across this bit https://github.com/apiaryio/drafter/blob/klokane/fix-inheritance/src/refract/IsExpandableVisitor.cc#L24

It looks like this is condition is redundant isn't it?  couldn't we just do `if (type == "ref")` because basically only when type is `ref` we want it to return `true`. Am I correct? 

Also couple of lines down https://github.com/apiaryio/drafter/blob/klokane/fix-inheritance/src/refract/IsExpandableVisitor.cc#L32 it looks like it can be just `return checkElement(e)`

klokane:
good catch, I mean in older version of code was no `ref` part of `isReserved()` that is probably reason. Will remove.

w-vi:
@klokane  rest of the bugs I can't find  :)

klokane:
Yeaah, they are really good hidden!

klokane:
@w-vi i look at condition `!isReserved(type) || type == "ref";` it has different meaning there is `not reserved` (mean inherited type) or `it is reference`. So this condition is correct.

klokane:
@w-vi review addressed. There are no tests :( in current tests implementation it is not possible to prepare test with comparable output.

We need implement querying for refract to build tests for similar cases

w-vi:
My part is done, I think it's ready for merge ... so leaving it to @pksunkara if he has anything to comment on

klokane:
BTW why "continuous-integration/appveyor" CI is (again and again) yellow? I mean, we should remove it form CI proces it it does not work.

pksunkara:
We should be relying on the other 2 green things, not the `continuous-integration/appveyor` which is just an aggregator.

```
 continuous-integration/appveyor/branch
 continuous-integration/appveyor/pr
```

The yellow one is an issue but hopefully appveyor will fix it soon. We can try to forget about it until then since we need the other CI statuses.

klokane:
@pksunkara yes, I know about it. But if I remember it is not working for few months. It is confusing

pksunkara:
> yes, I know about it. But if I remember it is not working for few months. It is confusing

I know and I feel the same. :sob: But we really do need the windows CI. :disappointed:

test/fixtures/mson/inheritance-primitive.apib
```diff
@@ -0,0 +1,27 @@
...
++ text (string)
+
+# Date (string)
+
+## Sample
+2012-12-00
+
+# Date2 (Date1)
+## Sample
+2012-12-01
```
pksunkara: Did you mean to have `2012-12-02` here?

src/RefractAPI.cc
```diff
@@ -82,10 +84,145 @@ namespace drafter {
         }
     }
 
+    struct InheritanceComparator {
```
pksunkara: This looks good but I need to look into this again more carefully.

src/RefractDataStructure.cc
```diff
@@ -723,12 +760,22 @@ namespace drafter {
...
 
             snowcrash::SourceMap<mson::Literal> sourceMap;
             sourceMap.sourceMap.append(property.sourceMap.name.sourceMap);
 
+            // check if base variable type is StringElement
+            if (!property.node.name.variable.typeDefinition.empty()) {
+                const std::string& type = property.node.name.variable.typeDefinition.typeSpecification.name.symbol.literal;
+                if (!type.empty()) {
+                    if (!refract::TypeQueryVisitor::as<refract::StringElement>(FindRootAncestor(type, GetNamedTypesRegistry()))) {
+                        throw snowcrash::Error("variable named property must be string ot its sub-type", snowcrash::MSONError);
```
pksunkara: `ot` -> `or`

src/RefractDataStructure.cc
```diff
@@ -68,9 +68,46 @@ namespace drafter {
...
+                return mson::UndefinedTypeName;
+        }
+
+        return mson::UndefinedTypeName;
+    }
+
     template<typename T>
     static mson::BaseTypeName GetType(const T& type) {
-        return type.typeDefinition.typeSpecification.name.base;
+        mson::BaseTypeName nameType =  type.typeDefinition.typeSpecification.name.base;
```
pksunkara: Extra space after `=`. I don't know if our style rules will fix this or not when implemented.

src/RefractDataStructure.cc
```diff
@@ -723,12 +760,22 @@ namespace drafter {
...
             }
 
             snowcrash::SourceMap<mson::Literal> sourceMap;
             sourceMap.sourceMap.append(property.sourceMap.name.sourceMap);
 
+            // check if base variable type is StringElement
+            if (!property.node.name.variable.typeDefinition.empty()) {
+                const std::string& type = property.node.name.variable.typeDefinition.typeSpecification.name.symbol.literal;
+                if (!type.empty()) {
+                    if (!refract::TypeQueryVisitor::as<refract::StringElement>(FindRootAncestor(type, GetNamedTypesRegistry()))) {
```
pksunkara: I am sorry but I might be confused. Why are we hardcoding this to `StringElement`? All primitive element types are possible. Aren't they?

```
# ID (number)
## Sample
123

# ID2 (ID)
## Sample
456
```

klokane: This is different case, *VariablePropertyName*
```
## VariableProperty (string)
### Sample
this can variate

## ObjectWithVariableProperty
- *var (VariableProperty)* (number)
```

and in MSON definition is written:
> In the case of specifying a Variable Property Name, the Value Definition MAY reference a Named Type that MUST be sub-typed from a string Primitive Type.

pksunkara: Ah, thanks for clearing it up. I was confused :)

src/RefractDataStructure.cc
klokane: Yes, this will fix clang-format, but I will fix by hand for now

src/RefractDataStructure.cc
klokane: good catch

src/refract/Element.h
```diff
@@ -287,10 +278,26 @@ namespace refract
...
+        Element() : hasContent(false), renderStrategy(IElement::rDefault), value(TraitType::init())
         {
         }
 
+        virtual bool empty() const
+        {
+            return !hasContent;
+        }
+
+
```
pksunkara: Unnecessary extra line

src/RefractDataStructure.cc
klokane: better to ask, maybe will add explaining comment to be clear

src/refract/Element.h
klokane: will fix format-clang

src/refract/RenderJSONVisitor.cc
```diff
@@ -9,265 +9,254 @@
...
+                }
             }
         }
 
         if (StringElement* str = TypeQueryVisitor::as<StringElement>(e.value.first)) {
-            assign(str->value, renderer.get());
+            MemberElement *m = new MemberElement;
+            IElement* v= renderer.result ? renderer.getOwnership() : new StringElement;
+            m->set(str->value, v);
+            result = m;
```
pksunkara: Couldn't this be done something like this?

```cpp
result = new MemberElement;
...
result->set(str->value, v);
```

src/refract/RenderJSONVisitor.cc
```diff
@@ -9,265 +9,254 @@
...
-            // Imitate an extend object
-            isExtend = true;
-            assign(renderer.get());
-            isExtend = false;
+        ObjectElement::ValueType members;
+        FetchMembers(e, members);
+        ObjectElement* o = new ObjectElement;
+        o->set(members);
+        o->renderType(IElement::rCompact);
+        result = o;
```
pksunkara: Why the dancing around here? Am I missing something? Can't we do the following:

```cpp
result = new ObjectElement;
result->set(members);
result->renderType(IElement::rCompact);
```

src/refract/RenderJSONVisitor.cc
```diff
@@ -9,265 +9,254 @@
...
+            return;
         }
 
-        assign(renderer.get());
+        ArrayElement::ValueType members;
+        FetchMembers(e, members);
+        ArrayElement* a = new ArrayElement;
+        a->set(members);
+        a->renderType(IElement::rCompact);
+        result = a;
```
pksunkara: Similar comment to https://github.com/apiaryio/drafter/pull/199/files#r48477358

src/refract/RenderJSONVisitor.cc
klokane: No, `result` is downcasted onto `IElement` so you will be not able to call `set()` method

src/refract/RenderJSONVisitor.cc
klokane: same as above

src/refract/RenderJSONVisitor.cc
klokane: dtto

src/refract/Element.cc
```diff
@@ -129,4 +131,276 @@ namespace refract
         }
     }
 
+    namespace {
+        struct TypeChecker {
```
pksunkara: Can you add a small comment above this line briefly explaining this? Thanks.

src/refract/Element.cc
```diff
@@ -129,4 +131,276 @@ namespace refract
...
+                        return;
+
+                    case TypeQueryVisitor::Object:
+                        doMerge<ObjectElement>(result, e);
+                        return;
+
+                    case TypeQueryVisitor::Member:
+                    case TypeQueryVisitor::Extend:
+                    case TypeQueryVisitor::Null:
+                        throw LogicError("Unappropriate kind of element to merging");
```
pksunkara: `refract::LogicError`

src/refract/Element.cc
```diff
@@ -129,4 +131,276 @@ namespace refract
...
+                    default:
+                        throw LogicError("Element has no implemented merging");
```
pksunkara: `refract::LogicError`

klokane: we are in `refract` namespace, it is restfull

src/refract/RenderJSONVisitor.cc
pksunkara: Ah. Thought I was missing something. Thanks.

src/RefractAPI.cc
```diff
@@ -82,10 +84,145 @@ namespace drafter {
...
+        }
+
+        const std::string& name(const mson::PropertyMember& pm) {
+            return name(pm.valueDefinition);
+        }
+
+        Members collectMembers(const mson::Elements& elements) {
+            Members members;
+
+            for (mson::Elements::const_iterator ie = elements.begin() ; ie != elements.end() ; ++ie) {
```
pksunkara: Please rename `ie` to `it`.

pksunkara:
@klokane Finished review. Just minor stuff to fix.

src/RefractAPI.cc
pksunkara: Looked again more slowly. :+1:

src/refract/Element.cc
klokane: Will receive container of elements via ref. Later `bool` typecasting operator will check, if all elements in container are of same type. 

It is alternative way to check it in c-tor and save bool value as result, and later just return result. Or it can be implemnted just as simple function.

I mean that was more complicated while a start it write and in a few iterations I refactor it into this.

In this case it is not important which way is used because we call `bool` casting operator just once on line 164 `if (!TypeChecker(val)) { // ...`

pksunkara: I understood what this is about but I think it's better if we have a comment in the code explaining this.

klokane: But in this case will be probably better to transform it into simple function

pksunkara: Yeah. I will leave it to you.

test/fixtures/mson/inheritance-primitive.apib
klokane: Probably, yes. But it is not important, so it can stay as is.
In final inheritance is used sample from `Date3` so test is able to check if it is inherited in required way

klokane:
Review addressed, fixedup on to previous commit `Fix typo` && push --force

PR #201, opened 2015-12-21
Fix expanded JSON Schema

w-vi:

Issue #202, opened 2015-12-21
nullable attaribute is not implemented for anything else then ObjectElement properties


klokane:
Nullable does not work for following use cases
```apib
# Data Structures

# ArrayWithNullable (array)
- m (nullable)

# NullableString (string, nullable)
```

pksunkara:
This is fixed by #201

klokane:
This is not fixed #201 implements `nullable` just for properties (MemeberElement), other kinds of elements are still not imlemented

kylef:
I can confirm string is now implemented:

```apib
+ Response 200 (application/json)
    + Attributes (string, nullable)
```

Message Body Schema:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "anyOf": [
    {
      "type": "null"
    },
    {
      "type": "string"
    }
  ]
}
```

Message Body:

```json
null
```

Array doesn't appear to be implemented.

PR #203, opened 2015-12-21
nullable attribute in JSON Schema

w-vi:
- where possible, see issue #202


w-vi:
cheery picked to the JSON Schema Fix https://github.com/apiaryio/drafter/pull/201

PR #204, opened 2016-01-04
EnumElement implementation

klokane:
Introduced `EnumElement` instead of parasiting on `ArrayElement`
- cleanup code

### DIFF
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -15,6 +15,23 @@
 namespace refract
 {
 
+    template <typename T>
+    void CloneMembers(T* a, const RefractElements* val)
+    {
+            for (RefractElements::const_iterator it = val->begin();
+                 it != val->end();
+                 ++it) {
+
+                if ((*it)->empty()) {
+                    continue;
+                }
+
+                IElement *e = (*it)->clone();
+                e->renderType(IElement::rCompact);
+                a->push_back(e);
+            }
+    }
+
     JSONSchemaVisitor::JSONSchemaVisitor(bool fixit /*= false*/) : fixed(fixit)
     {
         pObj = new ObjectElement;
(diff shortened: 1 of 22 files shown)


src/refract/VisitorUtils.cc
```diff
@@ -40,4 +40,8 @@ namespace refract
         return TypeQueryVisitor::as<StringElement>((*i)->value.second);
     }
 
+    void SetRenderFlag(RefractElements& elements, const IElement::renderFlags flag) {
```
w-vi: :+1:

src/refract/JSONSchemaVisitor.cc
```diff
@@ -15,6 +15,23 @@
 namespace refract
 {
 
+    template <typename T>
```
w-vi: Hope that it works as expected, as I;m not sure now if in one case we don't want to `clone` even empty elements as is done on line 284 ... but otherwise good move :)

w-vi:
Not related but this code won't work if the "type" key is already an array, will it? https://github.com/apiaryio/drafter/blob/klokane/enum-element/src/refract/JSONSchemaVisitor.cc#L82

src/refract/JSONSchemaVisitor.cc
```diff
@@ -304,73 +293,75 @@ namespace refract
...
+    }
+
+    void JSONSchemaVisitor::visit(const EnumElement& e) {
+        const EnumElement::ValueType* val = GetValue<EnumElement>(e);
+
+        if (!val || val->empty()) {
+            return;
+        }
+
+        enumElement(e, val);
```
w-vi: Could not these two methods be merged, it is now unnecessary to have it as it was just to clarify the `visit(ArrayElement)` method ? It would make it simpler and avoid unnecessary double `GetDefault(e)` call.

klokane: @w-vi about https://github.com/apiaryio/drafter/pull/204#issuecomment-168695317
Yes it is problem, simmilar if there will be already "array" with more values, it will enclose array into array.

klokane: @w-vi merge is good idea

w-vi:
@klokane apart form few bits here and there, nice job! It makes the code so much easier to read and follow ... halelujah :)

w-vi:
@klokane are you going to do any more changes to this? I'd love to merge it.

klokane:
@w-vi will address your notices today. will ping

klokane:
Sorry I merge && close this patch accidentally instead of another one.
Will make separeted patch to address review comments

PR #215, opened 2016-01-18
Use apiaryio homebrew formulae tap

### LINKED ISSUE #214: Remove head-only requirement from homebrew package
I've already been bitten by this once when the library was updated to 1.0; my scripts were built against 0.x library and it took forever to figure out why they were broken on a new teammate's Mac.

Tried installing from a stable release but got the following error:

```
$ brew install https://raw.githubusercontent.com/apiaryio/drafter/v2.1.2/tools/homebrew/drafter.rb
######################################################################## 100.0%
Error: drafter is a head-only formula
Install with `brew install --HEAD drafter`
```

I agree with [these folks](https://github.com/Homebrew/homebrew-head-only#why-is-head-only-bad) on the matter—as long as this package is head-only, I can't get consistent results out of any scripts I write that use drafter.

kylef:
This change will also allow us to provide a non-HEAD homebrew formula that uses the release tarball which allows `brew upgrade` to actually work for drafter.

Fixes #214
##### Dependencies
- [x] https://github.com/apiaryio/homebrew-formulae/pull/2
##### To-be merged after this is merged
- [x] ~~https://github.com/apiaryio/apiblueprintorg/pull/102~~


pksunkara:
This is going to be an issue. Whenever there's a new release of drafter, we need to update the homebrew-forumlae repo with the new version tag of drafter. Until now, we only had to do it in 1 place.

w-vi:
Is that a bad thing? This decouples homebrew install and the drafter master. Adding one more step is good, there will be one more incetive to automate it :)

pksunkara:
Let's wait until we automate then. Because a lot of people rely on homebrew for the latest version currently.

kylef:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;oBwczmvp&#x2F;439-build-plutonium-dev-release-mechanism"> Build Plutonium-dev release mechanism</a>

SMillerDev:
Not sure if this is still an open issue but the current approach does not provide updates to homebrew users.

kylef:
I've updated this pull-request to just remove our formula and point to the official one. Since homebrew now includes drafter by default: https://github.com/Homebrew/homebrew-core/blob/master/Formula/drafter.rb

PR #216, opened 2016-01-19
DESTDIR shouldn't be the BINDIR in the makefile

kylef:
This change splits out BINDIR from DESTDIR, traditionally DESTDIR is not the "bin" dir on other makefiles and this makes our makefile work similar to others.

This change also makes sure the bin directory exists using mkdir in the install phase.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,8 @@ BUILDTYPE ?= Release
 BUILD_DIR ?= ./build
 PYTHON ?= python
 GYP ?= ./ext/snowcrash/tools/gyp/gyp
-DESTDIR ?= /usr/local/bin
+DESTDIR ?= /usr/local
+BINDIR ?= $(DESTDIR)/bin
 
 # Default to verbose builds
 V ?= 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -26,7 +27,8 @@ drafter: config.gypi $(BUILD_DIR)/Makefile
 	cp -f $(BUILD_DIR)/out/$(BUILDTYPE)/$@ ./bin/$@
 
 install: drafter
-	cp -f $(BUILD_DIR)/out/$(BUILDTYPE)/drafter $(DESTDIR)/drafter
+	mkdir -p $(BINDIR)
+	cp -f $(BUILD_DIR)/out/$(BUILDTYPE)/drafter $(BINDIR)/drafter
```
kylef: Also, this install phase doesn't install the shared library or headers. Is there some reason it hasn't done this in the past @pksunkara?

I'd expect it would install the library so that I can build C or C++ tools/libraries that depend on it.

pksunkara: The average guy do not want to install the libraries and headers. Only the people who develop using drafter want them. And they probably don't install them to /usr/local rather installing them to their own customised places.

w-vi: I disagree ... at least in the Linux world you expect the headers and libs go to `/usr/local/include` and `/usr/local/lib` respectively, I think that there should be a `install_lib` target doing exactly this.

w-vi:
Another convention is to use `prefix` for setting the prefix of the path defaulting to `/usr/local` instead of `DESTDIR` see for example http://www.gnu.org/software/make/manual/html_node/Directory-Variables.html

pksunkara:
@w-vi If we want to have an `install_lib` target, let's make a trello card in the inbox. I will leave it to you to decide.

PR #220, opened 2016-01-22
Changes drafter.js API to protagonist.

w-vi:
drafter.js should be a drop in replacemenet so that it can be used where
native code is not possible or implies big overhead and parsing times are
not critical, i.e. dredd in CI where 10s parsing time extra is much less
then building protagonist from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,27 @@ Supported options:
   missing a title.
 - `type`: Either `refract` (default) or `ast`.
 
+`drafter.js` can serve as drop in replacement for
+[protagonist](https://github.com/apiaryio/protagonist), it supports
+the same API. So stuff like
+```js
+try {
+  var protagonist = require('protagonist');
+}
+catch (e) {
+  console.log("protagonist not found, using drafter.js.");
+  var protagonist = require('drafter.js');
+}
+
+protagonist.parse(data, options, function(err, result) {
+  if (err) {
+    console.log(JOSN.stringify(err));
+  }
+  console.log(JSON.stringify(data));
+});
+```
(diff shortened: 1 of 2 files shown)


w-vi:
@danielgtaylor please can you have a look, is it something what would work for aglio?

README.md
```diff
@@ -41,7 +41,7 @@ Other systems refer to [build notes](#build).
 ```c++
 #include "drafter.h"         // Blueprint Parser
 #include "SerializeResult.h" // Result Wrapper for serialization
-#include "sosJSON.h"         // Serializer
+#include "sosJSON.h"         // Serializger
```
kylef: Serializger?! Accidental change?

w-vi: yes, good catch, cheers

danielgtaylor:
This looks good! I only have one concern regarding option names:

> `if (options.exportSourcemap) {`
https://github.com/apiaryio/drafter/pull/220/files#diff-689601c2796264057e956ee9afc2c395R31

That is the Drafter name of the option. The name in Protagonist has been changed to `generateSourceMap` (see https://github.com/apiaryio/protagonist#parsing-options). If we want a true drop-in replacement we should probably use the new option name (or at least support it in a back-compat way). Note that we can send either the new or old option to Protagonist, and so we should do the same with Drafter.js in my opinion.

I think it will work nicely for Aglio. In my experimental branch I currently have logic specific to `drafter.parse` and it will be nice to remove that extra code. :+1:

zdne:
Waiting for this to update the Atom linter plugin.

PR #225, opened 2016-01-31
Resolve mixins in JSONSchema generator

### LINKED ISSUE #223: JSONSchema generator does not expand mixins correctly
Given following example API Blueprint
```
# Resource [/example]

## Action [GET]

    + Attributes (Notification Token Full)

# Data Structures

## Notification Token Create (object)
- device_type: `ios`    (string) - The platform. Should be `ios` or `gcm`.
- token: `a-push-token` (string) - The token given by the platform.
- links                 (object, optional)
  - user: 123           (string)

## Notification Token Full (object)
- id: `122` (string)
- Include Notification Token Create
```
the JSON Schema produced is not correct.

JSON:
```
{
 "id":"122",
 "device_type": "ios",
 "token": "a-push-token",
 "links": {
   "user": "123"
 }
}
```

JSON Schema
```
{ 
"type": "object",   
"properties": {
    "id": {"type": "string"},
    "type": "object"
    }
}
```

Expected JSON Schema should look more like the following
```
{ 
"type": "object",   
"properties": {
    "id": {"type": "string"},
    "device_type": {"type": "string"},
    "token": {"type": "string"},
    "links": {"type": "object"},
    }
}
```

w-vi:
Fixes #223

### DIFF
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -33,6 +33,45 @@ namespace refract
             }
     }
 
+
+    template <typename T>
+    void IncludeMembers(const T& element, typename T::ValueType& members)
+    {
+        const typename T::ValueType* val = GetValue<T>(element);
+
+        if (!val) {
+            return;
+        }
+
+        for (typename T::ValueType::const_iterator it = val->begin();
+             it != val->end();
+             ++it) {
+
+            if (!(*it) || (*it)->empty()) {
+                continue;
+            }
+
+            if ((*it)->element() == "ref") {
+                IElement::MemberElementCollection::const_iterator found = (*it)->attributes.find("resolved");
+
+                if (found == (*it)->attributes.end()) {
+                    continue;
+                }
+
+                const T* resolved = TypeQueryVisitor::as<T>((*found)->value.second);
+
+                if (!resolved) {
+                    throw refract::LogicError("Mixin must refer to same type as parent");
+                }
+
+                IncludeMembers(*resolved, members);
+                continue;
+            }
+
+            members.push_back(*it);
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


src/refract/JSONSchemaVisitor.cc
```diff
@@ -33,6 +33,45 @@ namespace refract
...
+                IElement::MemberElementCollection::const_iterator found = (*it)->attributes.find("resolved");
+
+                if (found == (*it)->attributes.end()) {
+                    continue;
+                }
+
+                const T* resolved = TypeQueryVisitor::as<T>((*found)->value.second);
+
+                if (!resolved) {
+                    throw refract::LogicError("Mixin must refer to same type as parent");
```
klokane: this should be probable catched one level higher and rethrow it with sourcemap info

klokane:
Patch it self looks OK. We need resolve error info propagation to upper level of application, because in generators (JSON/JSONSchema) we have no required info. Probably we need to create new task for it, because there is same problem in JSON

src/refract/JSONSchemaVisitor.cc
w-vi: That's the problem we have discussed on Friday isn't it? We should take some more general approach to this. I think even in this state it should be merged and the throw should be fixed later in separate PR where we fix other related parts too.

w-vi:
@klokane I agree we should fix the exceptions with sourcemaps in separate PR fixing it in other places too. If it is otherwise good, them merge it please. We should release drafter soon, as some of the changes/fixes are high priority.

klokane:
Ok going to merge.

pksunkara:
:+1:

honzajavorek:
This PR doesn't fully solve the issue. Please see https://github.com/apiaryio/drafter/issues/248.

PR #228, opened 2016-02-09
Add a mechanism to re-generate incorrect test fixtures

kylef:
By setting the environment variable `GENERATE` to `1`, `make test` will now re-generate incorrect fixtures. This is handy when making large changes which result in many fixture changes. You can run `make test`, verify all of the changes and then run `make test` with `GENERATE` set to `1` as follows to re-generate the fixtures:

    $ env GENERATE=1 make test

This follows the same behaviour as [fury-adapter-swagger](https://github.com/apiaryio/fury-adapter-swagger/blob/d88abbae682caa9d3ab27c9deea855da8c39091e/test/adapter.js#L32-L36).

### DIFF
--- a/test/draftertest.h
+++ b/test/draftertest.h
@@ -66,11 +66,11 @@ namespace draftertest {
     public:
         ITFixtureFiles(const std::string& base) : base_(base) {}
 
-        typedef std::auto_ptr<std::istream> stream_type;
+        typedef std::auto_ptr<std::istream> input_stream_type;
 
         const std::string fetchContent(const std::string& filename) const {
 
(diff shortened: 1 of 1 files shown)


test/draftertest.h
```diff
@@ -66,11 +66,11 @@ namespace draftertest {
     public:
         ITFixtureFiles(const std::string& base) : base_(base) {}
 
-        typedef std::auto_ptr<std::istream> stream_type;
+        typedef std::auto_ptr<std::istream> input_stream_type;
```
kylef: **Note** *changed this name because now that `IIFixtureFiles` can write, it's no longer clear that `stream_type` should be an input stream.*

PR #230, opened 2016-02-11
Added request parameters to refract

pksunkara:
Depends on apiaryio/snowcrash#366 and apiaryio/snowcrash#367.

Need to update snowcrash submodule in this PR once the above PRs are merged.

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -255,6 +255,10 @@ namespace drafter {
             return element;
         }
 
+        if (!payload.node->parameters.empty()) {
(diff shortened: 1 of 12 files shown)


pksunkara:
This is ready to be reviewed and merged. /cc @w-vi @klokane

kylef:
:+1: Looks good, but the tests are failing due to other changes in Snowcrash.

pksunkara:
@kylef Updated.

w-vi:
Looking at the circular reference test, the json and schema are not correctly rendered. I think its good to go but we should fix these in next sprint.

pksunkara:
@w-vi Can you create an issue on drafter if you have an example at hand? We should track all this stuff.

w-vi:
Sure,  just doing it.

PR #234, opened 2016-02-15
Handle JSON Schema rendering for messageBody

pksunkara:
This PR handles the case where a user might give the content type of a messageBody to be `application/schema+json`.

### DIFF
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -8,6 +8,8 @@
 
 #include "Render.h"
 #include "RefractDataStructure.h"
+
+#include "SourceAnnotation.h"
 #include "BlueprintUtility.h"
 #include "RegexMatch.h"
 
@@ -100,14 +102,10 @@ namespace drafter {
 
                 return std::make_pair(result, NodeInfo<Asset>::NullSourceMap());
             }
-
-            default:
-                break;
         }
 
-        // Just to be cautious
-        delete expanded;
-        return body;
+        // Throw exception
+        throw snowcrash::Error("unknown content type for messageBody to be rendered", snowcrash::ApplicationError);
     }
 
     NodeInfoByValue<Asset> renderPayloadSchema(const NodeInfo<snowcrash::Payload>& payload,
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -16,11 +16,15 @@ using namespace snowcrash;
 namespace drafter {
 
     // JSON content-type regex
-    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json";
+    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json[[:blank:]]*(;.*|$)";
+    const char* const JSONSchemaRegex = "^[[:blank:]]*application/schema\\+json[[:blank:]]*(;.*|$)";
 
     RenderFormat findRenderFormat(const std::string& contentType) {
 
-        if (RegexMatch(contentType, JSONRegex)) {
+        if (RegexMatch(contentType, JSONSchemaRegex)) {
+            return JSONSchemaRenderFormat;
+        }
+        else if (RegexMatch(contentType, JSONRegex)) {
             return JSONRenderFormat;
         }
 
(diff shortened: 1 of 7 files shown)


src/Render.cc
```diff
@@ -16,11 +16,15 @@ using namespace snowcrash;
 namespace drafter {
 
     // JSON content-type regex
-    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json";
+    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json[[:blank:]]*(;.*|$)";
+    const char* const JSONSchemaRegex = "^[[:blank:]]*application/schema\\+json[[:blank:]]*(;.*|$)";
```
w-vi: would'nt just a substring matching be enough? as we are looking fo exact string with possibly something more, which we don't care about.

src/Render.cc
```diff
@@ -57,38 +61,52 @@ namespace drafter {
...
 
-                return std::make_pair(renderer.getString(), NodeInfo<Asset>::NullSourceMap());
+                return std::make_pair(result, NodeInfo<Asset>::NullSourceMap());
             }
 
             default:
                 break;
         }
 
+        // Just to be cautious
```
w-vi: We should never get here so I would say abort here, the same for the default switch case, it should never happen so either abort or throw some runtime exception, we should know if this ever happens and not silently let it go through

w-vi:
two points to discuss otherwise legit.

src/Render.cc
pksunkara: Not really. You can check the test case for this.

src/Render.cc
pksunkara: So, remove the `default:` and replace the last two lines with exception?

w-vi: yes, to be sure we catch these cases somehow if it ever happens.

src/Render.cc
w-vi: OK, from the test case I don't see it, as there is exact string `application/schema+json` and then something more which is just copied exactly to the output. But that's OK, I just wanted to point it out to give it a second thought if doing regex is necessary.If you say so I believe you.

pksunkara: Basically, we need to recognise that this is JSON Schema even if the content-type is not exactly equal to `application/schema+json`.

pksunkara:
@w-vi Updated. I left the `default` on because it's good practice to have `default` in `switch`. The code will still throw error when it goes to `default` though. So, it should not be an issue.

kylef:
> @w-vi Updated. I left the `default` on because it's good practice to have `default` in `switch`. The code will still throw error when it goes to `default` though. So, it should not be an issue.

I strongly disagree that having a `default` is good practise. I would say the complete opposite and it's a bad practise. When you add a new unhandled enum case, the compiler won't inform you about this and you will have to check all existing code for enum statements that need updating. Without a `default` clause, the compiler will start giving you warnings that the new case is unhandled (not hidden by a `default`).

Clang may actually give you warnings when you have a `default` clause while all other cases are explicitly handled (if warning was enabled or `-Weverything`). See ["Why does Clang/LLVM warn me about using default in a switch statement where all enumerated cases are covered?"](http://programmers.stackexchange.com/questions/179269/why-does-clang-llvm-warn-me-about-using-default-in-a-switch-statement-where-all).

Given the following code example:

```c
enum Foo {
  Bar
};

int main(int argc, const char *argv[]) {
  Foo foo = Bar;

  switch (foo) {
    case Bar:
      return 1;
  }
}
```

Once we add a new case such as using the following diff:

```diff
enum Foo {
-   Bar
+   Bar,
+   Bar2
}
```

Which will now explicitly tell you that the case is unhandled when you compile:

```
foo2.cpp:9:11: warning: enumeration value 'Bar2' not handled in switch [-Wswitch]
  switch (foo) {
          ^
```

If there is nothing to do, it's far better to do the following:

```c
case Bar2:
  break
```

Than add a `default` clause so you have to explicitly think about future additions to the enum.

src/Render.cc
w-vi: Please correct me if I am reading the regex incorrectly. Beginning of the line, possibly white space; exact string `application/shcema+json`;possibly white space and then either semicolon and some string and or end of line. So you are looking for exact substring. True the regex would fail on string like `asdf application/schema+json` but the substring will still match but is it likely to happen? Anyway this is just a discussion to improve my regex skills.

w-vi:
I agree with kyle here, either have the default but then do something useful in default case or omit it and let the compiler do some job for you.

pksunkara:
Removed `default`.

src/Render.cc
pksunkara: `asdf application/schema+json` would fail and substring wouldn't match. Because there is an `^` at the beginning of the regex which means content type should start with `application/schema+json` (blank space excluded) and either end there or possibly have any parameters after a semicolon.

w-vi: That's exactly what I said.

pksunkara: I guess I got confused by your sentence.

> but the substring will still match but is it likely to happen

src/Render.cc
```diff
@@ -16,11 +18,15 @@ using namespace snowcrash;
 namespace drafter {
 
     // JSON content-type regex
-    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json";
+    const char* const JSONRegex = "^[[:blank:]]*application/(.*\\+)?json[[:blank:]]*(;.*|$)";
```
kylef: Why was this changed? Can we get some test that would fail without this changed?

src/Render.cc
w-vi: when you would to just substring matching, not regex just `substr` it would match.

pksunkara: Ah. Got it now. Yeah, `substr` would work. But I thought `regex` would work better as a standard in this part of code.

src/Render.cc
pksunkara: There's a test `application/schema+json ; hello=world`. Not exactly for `application/json` but it does test the added regex part.

src/Render.cc
w-vi: Regex comes with a price, you have to set up the state machine and go through it, `substr` is faster, thats why I asked

PR #236, opened 2016-02-16
Only generate JSON Schema for JSON responses

kylef:
We should not be rendering JSON Schema's when the content-type is not JSON.

Here's examples:

```apib
+ Response 200
    + Attributes
        + name: kyle

+ Response 200 (application/xml)
    + Attributes
        + name: kyle
```


kylef:
Hmm, this is more tricky than I expected.

PR #237, opened 2016-02-19
Support for MSVS2015

zdne:
## Support for MSVS2015

This PR updates Snow Crash to the latest master in order to bring updated GYP version to support Microsoft Visual Studio 2015 (Microsoft Visual C++ 2015).

In addition it updates the Windows batch build file to support the `MSVC2015` option. The `.gitattributes` file is introduced to preserve LF line endings on Windows systems. Finally, the AppVeyor configuration has been changed to use the Visual Studio 2015 CTP image instead of previous Windows Server 2012.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,11 @@
 version: 1.0.{build}
 
+# Operating system (build VM template)
+os: Visual Studio 2015 CTP
+
 environment:
   matrix:
-#   - VS_VERSION: MSVC2012
-    - VS_VERSION: MSVC2013
+    - VS_VERSION: MSVC2015
 
 build_script:
   - git submodule update --init --recursive
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Always use LF for .apib files
+*.apib text eol=lf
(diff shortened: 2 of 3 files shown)


appveyor.yml
```diff
@@ -1,9 +1,11 @@
 version: 1.0.{build}
 
+# Operating system (build VM template)
+os: Visual Studio 2015 CTP
+
 environment:
   matrix:
-#   - VS_VERSION: MSVC2012
-    - VS_VERSION: MSVC2013
```
pksunkara: Why not enable 2013 for matrix builds?

zdne: This is not how it is done. You would have to crete a whole environments (see the `os` entry above)

zdne: @pksunkara Matrix, as far as understand it, is for environment variables (eg. debug mode or release). You would need different OS images for having actual different versions of Visual Studios (eg Windows Server has MSVS2012 "OS" and and MSV2015 "OS" image has MSVS15... 

I cannot afford exploring various images setups in AppVeyor at the moment. Maybe something we can put into sprint. 

Keep in mind that `VS_VERSION: MSVC2013` is just a env variable that is used only by our GYP tool (= it does not control the AppVeyor environment in any way)

PR #238, opened 2016-02-19
Render JSON Schema only if body content type is JSON

pksunkara:
This PR also changes the macros for the tests so that we can do the selective testing more easier.

### DIFF
--- a/src/Render.cc
+++ b/src/Render.cc
@@ -116,6 +116,7 @@ namespace drafter {
                                                const refract::Registry& registry) {
 
         NodeInfoByValue<Asset> schema = std::make_pair(payload.node->schema, &payload.sourceMap->schema);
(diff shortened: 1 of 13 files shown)


src/Render.cc
```diff
@@ -116,6 +116,7 @@ namespace drafter {
                                                const refract::Registry& registry) {
 
         NodeInfoByValue<Asset> schema = std::make_pair(payload.node->schema, &payload.sourceMap->schema);
+        RenderFormat renderFormat = findRenderFormat(getContentTypeFromHeaders(payload.node->headers));
```
w-vi: Can this be moved to where it is needed, doing regex match every time even if we know that we don't needed is useless.

pksunkara: @w-vi https://github.com/apiaryio/drafter/pull/239

Issue #245, opened 2016-02-25
Inconsistent types detected 


amuino:
We are getting the following error running Drafter:
```
$ drafter --validate --use-line-num sample.apib.md

OK.
warning: (4)  ExtendElement must be composed from Elements of same type
```

This is the `sample.apib.md`
```
FORMAT: 1A

# My Public API
## Request [GET /some/endpoint]
+ Response 200 (application/json; charset=utf-8)
  + Attributes(Booking State Response)

# Data Structures
## Point (array)
## Booking State Params
## Booking State Response
- loc (Point)
## O1
## O2
## O3
## O4
## O5
## O6
## O7
## O8
## O9
## O10
## O11
## O12
## O13
## O14
```

After adding some debug output and rebuilding, the issue arises because when reaching [`ExtendElement::push_back`](https://github.com/apiaryio/drafter/blob/master/src/refract/Element.cc#L167-L188) we have `baseType` identified as an `ArrayElement` but `type` is an `ObjectElement`.

This seems to be caused by some previous error that I could not trace back (I'm not proficient at C++).

Just by removing `O14` from the blueprint, the `type` is correctly detected as `ArrayElement` and no errors are detected. They are also solved by some renaming of the `Booking State Response` structure, for instance shortening it to `Response`.

This is related to https://github.com/apiaryio/dredd/issues/381 and https://github.com/danielgtaylor/aglio/issues/237

zdne:
Duplicate see #217

kylef:
@amuino This was fixed in [Drafter 2.3.0-pre.0](https://github.com/apiaryio/drafter/releases/tag/v2.3.0-pre.0), would you be able to give that a go and confirm you're no longer seeing this.

amuino:
@kylef thanks for the heads up. How can I test it? I can't see the version [on npm](https://www.npmjs.com/package/drafter)

amuino:
@kylef verified that it is solved in 2.3.0-pre.0. Thanks for the heads up!

PR #247, opened 2016-03-07
Fix extend element

### LINKED ISSUE #217: JSON Schema generator fails on ExtendElement in enum
```
curl -s https://raw.githubusercontent.com/hassansin/foxycart-api-
blueprint/master/apiary.apib | drafter -l


error: (4)  ExtendElement must be composed from Elements of same type
```

I can't reproduce using latest drafter but getting different error

```
curl -s https://raw.githubusercontent.com/hassansin/foxycart-api-
blueprint/master/apiary.apib | drafter -l


error: (4)  ExtendElement serialization Not Implemented
```

At least part of the problem is that JSON Schema generator is not expanding elements if in `enum`, that's where the second error is coming from.

klokane:
This patch fix recent issues around "ExtendElement"
Fix #217 #245 

* Registration of NamedTypes now happens in "two steps"
- create and registrate empty refract element with correct type info
- create element with complex info (meta/attrs/content)

This solution will allow knowledge correct type info of referenced
element, so later while element is expanded we will receive correct
base element type for extend element merging

* Correctly handle deeper dependencies in registry for nested types definition
* Add ability to correctly handle complex types for nested type definition like `- type (array[Comlex, Types])`


src/NamedTypesRegistry.cc
```diff
@@ -220,15 +343,61 @@ namespace drafter {
...
+
+        DependencyTypeInfo typeInfo(found);
+
+        std::sort(found.begin(), found.end(), InheritanceComparator(typeInfo));
+
+#ifdef DEBUG_DEPENDENCIES
+        std::cout << "==BASE TYPE ORDER==" << std::endl;
+#endif /* DEBUG_DEPENDENCIES */
+
+#if 1
```
klokane: remove this `#ifdef`

test/fixtures/extend/circular.apib
```diff
@@ -0,0 +1,20 @@
...
+- (array[B])
+
+## B (enum)
+- (A)
+- (array[A])
+
+
+
+
+
```
pksunkara: Why all these extra lines?

test/fixtures/extend/circular.json
```diff
@@ -0,0 +1,182 @@
...
+                              "element": "asset",
+                              "meta": {
+                                "classes": [
+                                  "messageBody"
+                                ]
+                              },
+                              "attributes": {
+                                "contentType": "application/json"
+                              },
+                              "content": "[\n  \"\",\n  \"\"\n]"
```
pksunkara: I wonder if the JSON generated here is the correct one. /cc @w-vi What do you think?

Baggz:
<3

test/fixtures/extend/circular.apib
klokane: Probably cut/past & edit cycle. Will cut

test/fixtures/extend/circular.json
klokane: Hard to say, but it is real extreme. In real life it has no solution because you are not able to determine real value. So there is fallback to empty string in enum.

w-vi: The whole thing does not make any sense so having such JSON is OK, what else could be there?

test/fixtures/extend/many.json
```diff
@@ -0,0 +1,415 @@
...
+                              "element": "asset",
+                              "meta": {
+                                "classes": [
+                                  "messageBodySchema"
+                                ]
+                              },
+                              "attributes": {
+                                "contentType": "application/schema+json"
+                              },
+                              "content": "{\n  \"$schema\": \"http://json-schema.org/draft-04/schema#\",\n  \"type\": \"object\",\n  \"properties\": {\n    \"loc\": {}\n  }\n}"
```
w-vi: Schema is incorrect it should be an array, it probably is a bug and should be a separate fix

w-vi:
@klokane @pksunkara Apart from the test comments I think it's good to go

klokane:
Ok, so I will cut `extend/circular.apib` and we can go

klokane:
Updated

w-vi:
@pksunkara If you have nothing more to add please merge as I'm cool with the current state

test/fixtures/extend/many.json
pksunkara: @w-vi Can you please create an issue so we don't forget this? Thanks.

test/test-ElementFactoryTest.cc
```diff
@@ -0,0 +1,73 @@
...
+#include "refract/Visitors.h"
+#include "refract/Element.h"
+
+#include "RefractElementFactory.h"
+
+
+using namespace refract;
+using namespace drafter;
+
+TEST_CASE("Create empty primitive element","[ElementFactory]") {
```
pksunkara: Shouldn't the `{` be in new lines for the tests?

klokane: There are in `test-ApplyVisitorTest.cc` in same position.
But we should finally create clang-format config and add into repo

src/NamedTypesRegistry.cc
```diff
@@ -220,15 +343,59 @@ namespace drafter {
...
         for (DataStructures::const_iterator i = found.begin(); i != found.end(); ++i) {
 
             if (!i->node->name.symbol.literal.empty()) {
+
+                const std::string& name = i->node->name.symbol.literal;
                 refract::IElement* element = MSONToRefract(*i);
+#ifdef DEBUG_DEPENDENCIES
+                refract::TypeQueryVisitor v;
+                v.visit(*element);
+                std::cout
```
pksunkara: Can you please move all of this `cout` to a single line?

src/NamedTypesRegistry.cc
```diff
@@ -220,15 +343,59 @@ namespace drafter {
...
+            GetNamedTypesRegistry().add(element);
+        }
 
         for (DataStructures::const_iterator i = found.begin(); i != found.end(); ++i) {
 
             if (!i->node->name.symbol.literal.empty()) {
+
+                const std::string& name = i->node->name.symbol.literal;
                 refract::IElement* element = MSONToRefract(*i);
+#ifdef DEBUG_DEPENDENCIES
```
pksunkara: Please make sure there is an empty line before and after these kind of blocks for easier reading comprehension.

src/NamedTypesRegistry.cc
```diff
@@ -137,25 +160,83 @@ namespace drafter {
...
 
-                // map inheritance
+                // map inheritance && element
                 for (DataStructures::const_iterator i = elements.begin() ; i != elements.end() ; ++i) {
+
+                    nameToElement[name(i->node)] = &(*i->node);
+
                     if (hasParent(i->node)) {
                         childToParent[name(i->node)] = parent(i->node);
+#ifdef DEBUG_DEPENDENCIES
```
pksunkara: As said in other comment, please go through all occurrences of these blocks and make sure they have an empty line before and after them.

pksunkara:
Reviewed.

klokane:
@pksunkara can you please do ALL STYLE RELATED comments in one "turn"
It is really tedious make again and again rebase and push force just with style changes

pksunkara:
@klokane Sorry, But I do the style comments when I go through the code. :(

w-vi:
Again style discussion, please @pksunkara have a look at https://github.com/apiaryio/snowcrash/issues/360 and give it a try so we can start using it

PR #250, opened 2016-03-21
Add tool for json and JSONschema pretty printing

w-vi:


tools/refract-filter.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python
+
+# This is just a
```
pksunkara: Is this take from somewhere? Can we have the full file (with comments) and URL, so that we can update it later easily if it's updated upstream?

w-vi: What do you mean by taken from somewhere? I have written it to aid my variable property name efforts so this is the upstream.

w-vi: Oh I see, the comment, I moved the stuff to help instead of a comment but left few words there.

tools/refract-filter.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python
+
+from __future__ import print_function
+import argparse
+import sys
+import yaml
```
kylef: This is a non-standard library import so it should be at the bottom of the imports as per PEP8.

http://pep8.org/#imports

kylef: Since this might not be installed, can we make this an optional dependency and feature since the user might only be parsing JSON?

w-vi: Oh yeah, true. I need to setup proper python environment in my emacs to have all these conventions automated

w-vi: Good point, you don't necessarily need yaml and therefore happily live with the standard lib.

tools/refract-filter.py
```diff
@@ -0,0 +1,84 @@
...
+        return object_pairs_hook(loader.construct_pairs(node))
+
+    OrderedLoader.add_constructor(
+        yaml.resolver.BaseResolver.DEFAULT_MAPPING_TAG,
+        construct_mapping)
+    return yaml.load(stream, OrderedLoader)
+
+def walk(node, cb):
+
+    if type(node) is OrderedDict:
```
kylef: What about doing:

```python
if isinstance(node, dict):
```

So these statements are less coupled to the implementation details of what type of dictionary these are (ordered, non-ordered).

w-vi:
@kylef addressed

w-vi:
@kylef ping

PR #256, opened 2016-03-22
Add variable properties to JSON Schema generator

w-vi:
First attempt to get the JSON Schema for variable property names ... please review, I might go thorught the code one more time to make it nice and refactor but more eyes will see more. @klokane

### DIFF
--- a/src/refract/Element.h
+++ b/src/refract/Element.h
@@ -478,7 +478,7 @@ namespace refract
     struct MemberElement : Element<MemberElement, MemberElementTrait>
     {
 
-        MemberElement() : Type() 
+        MemberElement() : Type()
         {
         }
 
@@ -532,7 +532,7 @@ namespace refract
 
     struct ObjectElementTrait : public ElementCollectionTrait
     {
-        // Use inherited ValueType definition instead of specialized std::vector<MemberElement*> 
+        // Use inherited ValueType definition instead of specialized std::vector<MemberElement*>
         // because ObjectElement can contain:
         // - (array[Member Element])
         // - (object)
@@ -557,6 +557,12 @@ namespace refract
         {
             set(value);
             renderType(render);
+        }
+
+        ObjectElement(IElement* e, IElement::renderFlags render = IElement::rDefault) : Type()
+        {
+            renderType(render);
+            value.push_back(e);
         }
 
         void push_back(IElement* e)
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -72,10 +72,17 @@ namespace refract
         }
     }
 
-    JSONSchemaVisitor::JSONSchemaVisitor(bool fixit /*= false*/) : fixed(fixit)
+    JSONSchemaVisitor::JSONSchemaVisitor(ObjectElement *pDefinitions /*= NULL*/,
+                                         bool fixit /*= false*/)
+        : pDefs(pDefinitions), fixed(fixit)
     {
         pObj = new ObjectElement;
         pObj->renderType(IElement::rCompact);
+
+        if (!pDefs) {
+            pDefs = new ObjectElement;
+            pDefs->renderType(IElement::rCompact);
+        }
     }
 
     JSONSchemaVisitor::~JSONSchemaVisitor()
@@ -84,6 +91,7 @@ namespace refract
             delete pObj;
         }
     }
+
 
     void JSONSchemaVisitor::setFixed(bool fixit)
     {
@@ -110,11 +118,10 @@ namespace refract
 
     void JSONSchemaVisitor::setSchemaType(const std::string& type)
     {
-        StringElement *s = new StringElement;
-
-        s->renderType(IElement::rCompact);
-        s->set(type);
-        addMember("type", s);
+        addMember("type",
+                  new StringElement(
+                      type,
+                      IElement::rCompact));
     }
 
     void JSONSchemaVisitor::addSchemaType(const std::string& type)
@@ -137,11 +144,10 @@ namespace refract
 
     void JSONSchemaVisitor::addMember(const std::string& key, IElement *val)
     {
-        MemberElement *m = new MemberElement;
-
-        m->renderType(IElement::rCompact);
-        m->set(key, val);
-        pObj->push_back(m);
+        pObj->push_back(new MemberElement(
+                            key,
+                            val,
+                            IElement::rCompact));
     }
 
     bool JSONSchemaVisitor::allItemsEmpty(const ArrayElement::ValueType* val)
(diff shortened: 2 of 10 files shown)


src/refract/Element.h
```diff
@@ -566,6 +566,8 @@ namespace refract
             hasContent = true;
             value.push_back(e);
         }
+
+        void addMember(const std::string& key, IElement* val, IElement::renderFlags render = IElement::rDefault);
```
klokane: I suggest to avoid add this member function, instead you can do:
```
object->push_back(new MemberElement(key, val, renderFlag));
```
Meaning of this is obvious and there is no so much additional writing.

Next argument is, we are not compatible among other kind of

src/refract/JSONSchemaVisitor.cc
```diff
@@ -144,6 +151,15 @@ namespace refract
...
     }
 
+    void JSONSchemaVisitor::addDefinition(const std::string& key, IElement *val)
+    {
+        MemberElement *m = new MemberElement;
+
+        m->renderType(IElement::rCompact);
+        m->set(key, val);
+        pDefs->push_back(m);
+    }
```
klokane: Next argument to do not define `ObjectElement::addMember()`
Nobody will remember special function to add member :)

```
pDefs->push_back(new MemberElement(key, val, IElement::rCompact));
```

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +231,67 @@ namespace refract
...
+        for(std::vector<MemberElement *>::const_iterator i = props.begin();
+            i != props.end();
+            ++i) {
+            ObjectElement *ob = new ObjectElement;
+            ob->renderType(IElement::rCompact);
+            StringElement *str = TypeQueryVisitor::as<StringElement>((*i)->value.first);
+            JSONSchemaVisitor renderer(pDefs, fixed);
+            renderer.visit(*(*i)->value.second);
+            addDefinition(str->value, renderer.getOwnership());
+            StringElement *s = new StringElement;
```
klokane: Probably can use "setting" constructor
```
StringElement *s = new StringElement("#/definitions/" + str->value, IElement::rCompact);
```

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +231,67 @@ namespace refract
...
+    ArrayElement* JSONSchemaVisitor::arrayFromProps(std::vector<MemberElement*>& props)
+    {
+        ArrayElement *a = new ArrayElement;
+        a->renderType(IElement::rCompact);
+        for(std::vector<MemberElement *>::const_iterator i = props.begin();
+            i != props.end();
+            ++i) {
+            ObjectElement *ob = new ObjectElement;
+            ob->renderType(IElement::rCompact);
+            StringElement *str = TypeQueryVisitor::as<StringElement>((*i)->value.first);
```
klokane: check against NULL

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +231,67 @@ namespace refract
...
+            JSONSchemaVisitor renderer(pDefs, fixed);
+            renderer.visit(*(*i)->value.second);
+            addDefinition(str->value, renderer.getOwnership());
+            StringElement *s = new StringElement;
+            s->renderType(IElement::rCompact);
+            s->set("#/definitions/" + str->value);
+            ob->addMember("$ref", s, IElement::rCompact);
+            a->push_back(ob);
+        }
```
klokane: Take look to `refract::Build`

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +231,67 @@ namespace refract
...
+
+        return a;
+    }
+
+    void JSONSchemaVisitor::addVariableProps(std::vector<MemberElement*>& props,
+                                             ObjectElement *o)
+    {
+        if (o->empty()) {
+            if (props.size() == 1) {
+                StringElement *str = TypeQueryVisitor::as<StringElement>(props.front()->value.first);
```
klokane: check against NULL

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +231,67 @@ namespace refract
...
+                JSONSchemaVisitor renderer(pDefs, fixed);
+                renderer.visit(*props.front()->value.second);
+                addDefinition(str->value, renderer.getOwnership());
+                StringElement *s = new StringElement;
```
klokane: Use setting constructor.
Maybe extract into function like `createReferenceToDefinition()` and use it in `arrayFromProps()` too

src/refract/JSONSchemaVisitor.cc
```diff
@@ -232,30 +300,42 @@ namespace refract
...
-                    o->push_back(m1);
+                if (IsVariableProperty(*mr->value.first)) {
+                    varProps.push_back(mr);
+                }
+                else {
+                    JSONSchemaVisitor renderer(pDefs, fixed);
+                    renderer.visit(*(*it));
+                    ObjectElement *o1 = TypeQueryVisitor::as<ObjectElement>(renderer.get());
+                    if (!o1->value.empty()) {
+                        MemberElement *m1 = TypeQueryVisitor::as<MemberElement>(o1->value[0]->clone());
```
klokane: sometime you use `value[0]` sometime `front()`, please choose one of them.

If `cloned` element not will be of type `MemeberElement` there will be leak memory.

klokane:
Concept of "definitions/reference" seems to be good.
warn: I did not check validity of generated Schema! Hope you did :)

Please refactor and i mean we can move forward

src/refract/JSONSchemaVisitor.cc
w-vi: Is it possible that ObjectElement members are not MemberElements?

src/refract/JSONSchemaVisitor.cc
w-vi: This didn't work for me so I just made it work, but will have a look into it, what was the problem.

src/refract/JSONSchemaVisitor.cc
klokane: Yes by refract spec. https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#object-element-element

w-vi:
I did validate the tests and the schemas generated, so this is correct at least for the tests :)

src/refract/JSONSchemaVisitor.cc
klokane: probably `const char* + std::string` you will need to do it in extra step, 
or alternative `std::string("#/definitions/") + str->value`

setting constructor probably expects `cons ElementString::ValueType&`

klokane:
:+1:

src/refract/JSONSchemaVisitor.cc
w-vi: So this bug is there since day 1 I think :)

klokane: goo to know :)

w-vi:
@klokane refactored to some degree, please review one more time

src/refract/JSONSchemaVisitor.cc
```diff
@@ -216,14 +222,72 @@ namespace refract
...
+        a->renderType(IElement::rCompact);
+        for(std::vector<MemberElement *>::const_iterator i = props.begin();
+            i != props.end();
+            ++i) {
+
+            StringElement *str = TypeQueryVisitor::as<StringElement>((*i)->value.first);
+            if (str) {
+                JSONSchemaVisitor renderer(pDefs, fixed);
+                renderer.visit(*(*i)->value.second);
+                pDefs->push_back(new MemberElement(str->value,
```
klokane: alignment there and block below

w-vi:
@klokane reformatted

klokane:
ok, going to merge

PR #240, opened 2016-02-22
SourceMaps always from SC

w-vi:
drafter now always wants sourcemaps from snowcrash and filters it from the output if not requested. This is not fixing the error reports from drafter yet but is a necessary step towards it.

*Note:*
The solution is not nice at all, and needs refactoring but  in order to not spend too much time on it leaving the refactoring for later so we can plan for it.

PR #260, opened 2016-03-23
Attach source maps to httpRequest and httpResponse

### LINKED ISSUE #259: Requests and Responses do not have source maps
Currently, HTTP Requests and HTTP Responses do not have source maps in the `refract` output. Their sub-elements and attributes have source maps, but there's no way to determine where the actual `+ Request` or `+ Response` happened to be in the original document.

In following situation I can distinguish the requests, because they happen to have headers. Headers have source maps, so I can say which request was on which position:

```apib
+ Request (application/json)
+ Response 200

+ Request (application/json)
+ Response 200
```

However, imagine following situation with "bare" requests:

```apib
+ Request
+ Response 200

+ Request
+ Response 200
```

In this case the requests do not have a single sub-element with a source map (except of the `method` attribute, but that doesn't actually say where the `+ Request` was written). I can say where responses were positioned, because I have source maps for the status codes. However, there is no way to say this for the requests. I can't distinguish which was which or whether they're the same and I can't re-construct number of the implicit HTTP transaction example (in this case, the first request is Example 1, the second is Example 2).

This fact blocks me to continue with https://github.com/apiaryio/dredd-transactions/pull/2. I can't flawlessly implement backward compatibility for https://github.com/apiaryio/dredd/issues/388 and I can't support [API Elements](http://api-elements.readthedocs.org/en/latest/overview/) in Dredd.

kylef:
Source map info was not attached to httpRequest and httpResponse, see #259 for more info.

#### Dependencies

- [x] https://github.com/apiaryio/snowcrash/pull/376
- [x] Account for changes in Snowcrash ([removal of source map options](https://github.com/apiaryio/snowcrash/pull/369)), I believe this was intended to be addressed in https://github.com/apiaryio/drafter/pull/240, if this isn't the case I'm happy to do it.

Closes #259

### DIFF
--- a/features/fixtures/refract.sourcemap.json
+++ b/features/fixtures/refract.sourcemap.json
@@ -611,6 +611,17 @@
                               },
                               "content": "<request name>"
                             },
+                            "sourceMap": [
+                              {
+                                "element": "sourceMap",
+                                "content": [
+                                  [
+                                    1039,
+                                    37
+                                  ]
+                                ]
+                              }
+                            ],
                             "headers": {
                               "element": "httpHeaders",
                               "content": [
@@ -807,6 +818,17 @@
                               },
                               "content": "200"
                             },
+                            "sourceMap": [
+                              {
+                                "element": "sourceMap",
+                                "content": [
+                                  [
+                                    1242,
+                                    27
+                                  ]
+                                ]
+                              }
+                            ],
                             "headers": {
                               "element": "httpHeaders",
                               "content": [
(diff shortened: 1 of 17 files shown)


w-vi:
Yes #240 is counting on the sourceMaps changes in snowcrash, @klokane knows the current situation and I guess he has changed snowcrash quite a bit, so I'd suggest wait for it, as per our discussion it should be tomorrow ready

PR #263, opened 2016-03-24
Add valgrind memcheck to CI test

w-vi:


klokane:
Build is failing :-)

circle.yml
```diff
@@ -28,9 +30,11 @@ test:
     - ./configure -i
   override:
     - env CXXFLAGS="-Werror" make test
+    - valgrind --leak-check=full --show-leak-kinds=definite,indirect,possible --error-exitcode=1 ./bin/test-libdrafter 2> memcheck.log
```
kylef: Do you think it would make sense to move this to the makefile?

```shell
$ make memory-test
```

I can see this useful for running locally without remembering all of these args.

w-vi: locally you can run just `valgrind drafter apiary.apib` and you see the result so I think it is not necessary.  The flags are there to fail the CI test and have enough info in the logfile.

PR #266, opened 2016-03-31
Smart Serialization for sourcemaps

klokane:
Automaticaly full rendering for elements with "SourceMaps"

### DIFF
--- a/src/refract/SerializeVisitor.cc
+++ b/src/refract/SerializeVisitor.cc
@@ -27,6 +27,7 @@ namespace refract
                         continue;
                     }
                 }
+
                 SerializeCompactVisitor s(exportSourceMap);
                 s.visit(*(*it));
                 result.set(s.key(), s.value());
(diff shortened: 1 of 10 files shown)


src/refract/SerializeVisitor.cc
```diff
@@ -54,6 +55,23 @@ namespace refract
             return array;
         }
 
+        IElement::renderFlags SelectSerializationType(const IElement& e, bool exportSourceMap) {
+
+            if (exportSourceMap) {
+                IElement::MemberElementCollection::const_iterator it = e.attributes.find("sourceMap");
+                // there is sourceMap in attributes
```
kylef: Btw, an "objective-c-ism" to turn this comment into a property:

```cpp
bool attributesIncludesSourceMap = it != e.attributes.end();

if (attributesIncludesSourceMap) {
    IElement::rFull
}
```

I think this is a great way to write more readable code without comments.

kylef:
This looks good to me, but Circle CI is failing because this introduces a build warning:

```
../src/refract/SerializeCompactVisitor.cc:19:84: error: type qualifiers ignored on function return type [-Werror=ignored-qualifiers]
         const bool operator()(const IElement* element, const bool exportSourceMap) const {
```

klokane:
ahh, will fix

klokane:
Fixed, push forced

PR #264, opened 2016-03-29
Resolve sourcemaps for throws in drafter

klokane:
Need to rebased on current master.

PR #268, opened 2016-04-01
Prevent recursion while converting MSON type members to refract

### LINKED ISSUE #267: Regression: Recursion when converting MSON Member to Refract with default types
With the following API Blueprint, we are looping over during converting the MSON Member to Refract.

```apib
## Data Structures
### Object
+ thing (optional, array[])
    + foo: bar
```

kylef:
While the type is the same as the default nested type, we would be causing recursion by calling ourself with the same arguments.

Fixes #267.

---

Perhaps there is a better solution, welcome to suggestions.

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -892,13 +892,15 @@ namespace drafter {
                 else if (ValueHasName(input.node) || ValueHasMembers(input.node)) {
                     return Trait::template Invoke<refract::ObjectElement>(input, defaultNestedType);
                 }
+                else if (nameType != defaultNestedType) {
+                    return MsonMemberToRefract<Trait>(input, defaultNestedType, defaultNestedType);
+                }
 
-                return MsonMemberToRefract<Trait>(input, defaultNestedType, defaultNestedType);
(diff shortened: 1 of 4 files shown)


src/RefractDataStructure.cc
```diff
@@ -900,13 +900,13 @@ namespace drafter {
...
+                else if (nameType != defaultNestedType) {
+                    return MsonMemberToRefract<Trait>(input, defaultNestedType, defaultNestedType);
+                }
             }
-
-            default:
-                throw snowcrash::Error("unknown type of mson member", snowcrash::MSONError);
         }
+
+        throw snowcrash::Error("unknown type of mson member", snowcrash::MSONError);
```
kylef: I didn't find any other tests for `unknown type of moon member`, unsure how we can go about testing this. Would love suggestions.

kylef:
Let's get #264 merged first and I will rebase this once that's done.

w-vi:
Testing exceptions and errors is not easy and we are not testing many of them yet. This needs to be addressed as well as other issues in our testing infrastructure.

PR #270, opened 2016-04-01
Treat snowcrash errors as errors and don't continue parsing AST

### LINKED ISSUE #269: Snow crash errors are not treated as "errors"
We shouldn't proceed parsing the AST when there are snowcrash errors.

```apib
# API
## Data Structures
### Child (Child)
```

kylef:
Before this pull request we would continue converting the AST to Refract (i.e, and rendering JSON / JSON Schema) which means we would continue even if there were errors such as circular references.

This pull request breaks early and returns Snowcrash errors instead of continuing. To be able to test that we've resolved #269, I have had to use parse result AST instead of plain API Blueprint AST in the tests so that the errors from the underlying wrap function are caught.

Closes #269


src/SerializeResult.cc
```diff
@@ -50,19 +50,21 @@ sos::Object WrapParseResultAST(snowcrash::ParseResult<snowcrash::Blueprint>& blu
     sos::Object object;
     snowcrash::Error error;
 
-    try {
-        object.set(SerializeKey::Version, sos::String(PARSE_RESULT_SERIALIZATION_VERSION));
-        object.set(SerializeKey::Ast, WrapBlueprint(blueprint, options.expandMSON));
-
-        if (options.exportSourceMap) {
-            object.set(SerializeKey::Sourcemap, WrapBlueprintSourcemap(blueprint.sourceMap));
+    if (blueprint.report.error.code == snowcrash::Error::OK) {
```
kylef: TODO: Can revert this change because now WrapBlueprint will throw  and this is caught below.

PR #279, opened 2016-04-05
Fix memory leaks in tests

w-vi:


circle.yml
```diff
@@ -28,9 +30,11 @@ test:
     - ./configure -i
   override:
     - env CXXFLAGS="-Werror" make test
+    - valgrind --leak-check=full --show-leak-kinds=definite,indirect,possible --error-exitcode=1 ./bin/test-libdrafter 2> memcheck.log
```
pksunkara: Does the memory leaks get printed when valgrind exits on error?

kylef: They go to the `memcheck.log` build artefact.

pksunkara: Can we echo that file to terminal if valgrind errors out? Will make our life a bit simpler.

w-vi: I think you are better off looking at the log file, the output is usually big and I didn't want to clutter the output in CI

pksunkara: circleCI has collapsible stuff for each command so that clutter should not be an issue.

pksunkara:
Valgrind failed in circleCI

w-vi:
I'll have a look why once the build is finsihed

circle.yml
kylef: I don't think this would be useful in CircleCI, with word-wrapping it's very difficult to read.

Example:

screen shot 2016-04-05 at 13 54 07

PR #265, opened 2016-03-30
Include missing element factory in the Xcode project

kylef:
Without these changes, the project is unbuildable inside Xcode because of missing `RefractElementFactory`.

PR #287, opened 2016-04-11
Xcode maintenance

pksunkara:
Also added `src/refract` as an include_dir to `test-libdrafter` target in gyp.

### DIFF
--- a/drafter.gyp
+++ b/drafter.gyp
@@ -111,6 +111,7 @@
       'type': 'executable',
       'include_dirs': [
         'src',
+        'src/refract',
         'test',
         'test/vendor/Catch/include',
         'test/vendor/dtl/dtl',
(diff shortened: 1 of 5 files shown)


drafter.gyp
```diff
@@ -111,6 +111,7 @@
       'type': 'executable',
       'include_dirs': [
         'src',
+        'src/refract',
```
klokane: Why this? I mean everywhere is used `#include "refract/something.h"

pksunkara: That way is not working in Xcode test target. So, I had to compromise

klokane: It meant, it doesn't work for a few months testing in Xcode?

pksunkara: Yes, It wasn't working for a long time and I neglected it since I was able to work around it.

https://github.com/apiaryio/drafter/pull/265

PR #289, opened 2016-04-11
Make sure drafter errors are serialized as errors

### LINKED ISSUE #255: Certain warnings are being treated as errors
I'm not sure what the correct behaviour is here. We are raising some "warnings" in parse results, however these are being treated as errors since the parse result doesn't contain other content.

You can find examples of this in both #252 and #254.

```yaml
element: "parseResult"
content:
  -
    element: "annotation"
    meta:
      classes:
        - "warning"
    attributes:
      code: 4
      sourceMap:
        -
          element: "sourceMap"
          content: []
    content: "no property name"
```

The solution is that these warnings should actually be errors, or other content (API name, etc) should be in the parse result.

pksunkara:
Fixes #255.

@kylef Remember the small bit of code we talked about being nonsense? That is the one causing this issue.

PS: The current test blueprint we have is producing an error in drafter due to bug. Once that bug is fixed, this test won't work anymore. I am unable to get a blueprint which produces error in drafter (without it not being due to bug).

Maybe @w-vi or @klokane has a few ideas for the test case.

### DIFF
--- a/src/SerializeResult.cc
+++ b/src/SerializeResult.cc
@@ -50,27 +50,25 @@ sos::Object WrapParseResultAST(snowcrash::ParseResult<snowcrash::Blueprint>& blu
     sos::Object object;
     snowcrash::Error error;
 
-    try {
-        object.set(SerializeKey::Version, sos::String(PARSE_RESULT_SERIALIZATION_VERSION));
-        object.set(SerializeKey::Ast, WrapBlueprint(blueprint, options.expandMSON));
+    object.set(SerializeKey::Version, sos::String(PARSE_RESULT_SERIALIZATION_VERSION));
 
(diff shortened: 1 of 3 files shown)


pksunkara:
Okay, whenever we are throwing in drafter, there's a memory leak. All the exceptions/errors thrown are only caught at the beginning of the `WrapParseResultRefract` function, any allocations which happen in `*ToRefract` functions get leaked.

I am removing the test case for now. The throws in drafter will happen only if we have a bug in the parser mostly. So, let's leave it for now and replace them with shared_ptr when we upgrade our compiler.

w-vi:
Why there is `ast` if there is no result? AFAIK this hasn't been the case.

pksunkara:
@w-vi Updated.

PR #291, opened 2016-04-11
Correcting mistake from #290

pksunkara:
Sorry, it was a bad edit which got into the commit.


w-vi:
BTW why this is still `ExportSourceMap` and not `Generate`?

pksunkara:
We will be changing that in the next major release.

PR #295, opened 2016-04-13
Introduce conversion context for thread safety

klokane:
This patch additionaly remove global access point `GetNamedTypeRegistry()` instead is `Registry` instance stored in `ConversionContext`

To be thread safe we need add `RefractElementFactory` into Context too.
It will be done in next pull request

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -18,12 +18,13 @@
 #include <set>
 
 #include "NamedTypesRegistry.h"
+#include "ConversionContext.h"
 
 namespace drafter {
 
     // Forward Declarations
     using refract::RefractElements;
-    refract::IElement* ElementToRefract(const NodeInfo<snowcrash::Element>& element);
+    refract::IElement* ElementToRefract(const NodeInfo<snowcrash::Element>& element, ConversionContext& context);
 
     namespace {
 
@@ -38,15 +39,31 @@ namespace drafter {
             elements.erase(std::remove_if(elements.begin(), elements.end(), IsNull<refract::IElement>), elements.end());
         }
 
+        template <typename T, typename Functor>
+        struct ContextBinder {
+
+            const Functor& functor;
+            ConversionContext& context;
+
+            ContextBinder(const Functor& functor, ConversionContext& context) : functor(functor), context(context) {}
+
+            refract::IElement* operator()(const NodeInfo<T>& nodeInfo) const {
+                return functor(nodeInfo, context);
+            }
+        };
+
         template <typename T, typename F>
-        void NodeInfoToElements(const NodeInfo<T>& nodeInfo, const F& transformFunctor, RefractElements& content)
+        void NodeInfoToElements(const NodeInfo<T>& nodeInfo, const F& transformFunctor, RefractElements& content, ConversionContext& context)
         {
             NodeInfoCollection<T> nodeInfoCollection(nodeInfo);
-            std::transform(nodeInfoCollection.begin(), nodeInfoCollection.end(), std::back_inserter(content), transformFunctor);
+
+            std::transform(nodeInfoCollection.begin(), nodeInfoCollection.end(),
+                           std::back_inserter(content),
+                           ContextBinder<typename T::value_type, F>(transformFunctor, context));
         }
 
         template<typename T, typename C, typename F>
-        T* CollectionToRefract(const NodeInfo<C>& collection, const F& transformFunctor, const std::string& key = std::string(), const refract::IElement::renderFlags renderType = refract::IElement::rCompact)
+        T* CollectionToRefract(const NodeInfo<C>& collection, ConversionContext& context, const F& transformFunctor, const std::string& key = std::string(), const refract::IElement::renderFlags renderType = refract::IElement::rCompact)
         {
             T* element = new T;
             RefractElements content;
@@ -55,7 +72,7 @@ namespace drafter {
                 element->element(key);
             }
 
-            NodeInfoToElements(collection, transformFunctor, content);
+            NodeInfoToElements(collection, transformFunctor, content, context);
 
             RemoveEmptyElements(content);
 
@@ -67,9 +84,9 @@ namespace drafter {
         }
     }
 
-    refract::IElement* DataStructureToRefract(const NodeInfo<snowcrash::DataStructure>& dataStructure, bool expand)
+    refract::IElement* DataStructureToRefract(const NodeInfo<snowcrash::DataStructure>& dataStructure, ConversionContext& context, bool expand)
     {
-        refract::IElement* msonElement = MSONToRefract(dataStructure);
+        refract::IElement* msonElement = MSONToRefract(dataStructure, context);
 
         if (expand) {
             refract::IElement* msonExpanded = ExpandRefract(msonElement, GetNamedTypesRegistry());
@@ -87,7 +104,7 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* MetadataToRefract(const NodeInfo<snowcrash::Metadata>& metadata)
+    refract::IElement* MetadataToRefract(const NodeInfo<snowcrash::Metadata>& metadata, ConversionContext& context)
     {
         refract::MemberElement* element = new refract::MemberElement;
 
@@ -103,7 +120,7 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* CopyToRefract(const NodeInfo<std::string>& copy)
+    refract::IElement* CopyToRefract(const NodeInfo<std::string>& copy, ConversionContext& context)
     {
         if (copy.node->empty()) {
             return NULL;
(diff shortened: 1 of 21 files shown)


src/RefractAPI.cc
```diff
@@ -38,15 +39,31 @@ namespace drafter {
             elements.erase(std::remove_if(elements.begin(), elements.end(), IsNull<refract::IElement>), elements.end());
         }
 
+        template <typename T, typename Functor>
```
pksunkara: Use `F` instead of `Functor`? We are doing it like that in the below functions.

src/RefractDataStructure.cc
```diff
@@ -62,7 +62,7 @@ namespace drafter {
 
     using refract::RefractElements;
 
-    static void SetElementType(refract::IElement* element, const mson::TypeDefinition& td)
+    static void SetElementType(refract::IElement* element, const mson::TypeDefinition& td, ConversionContext& context)
```
pksunkara: We don't need context here.

src/RefractDataStructure.cc
```diff
@@ -1050,7 +1060,7 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* ExpandRefract(refract::IElement* element, const refract::Registry& registry)
+    refract::IElement* ExpandRefract(refract::IElement* element, ConversionContext& context, const refract::Registry& registry)
```
pksunkara: We don't need the third argument here. We can get the registry from the context which is the second argument.

src/Render.h
```diff
@@ -21,11 +21,20 @@ namespace drafter {
...
+
     RenderFormat findRenderFormat(const std::string& contentType);
     std::string getContentTypeFromHeaders(const snowcrash::Headers& headers);
 
-    NodeInfoByValue<snowcrash::Asset> renderPayloadBody(const NodeInfo<snowcrash::Payload>& payload, const NodeInfo<snowcrash::Action>& action, const refract::Registry& registry);
-    NodeInfoByValue<snowcrash::Asset> renderPayloadSchema(const NodeInfo<snowcrash::Payload>& payload, const NodeInfo<snowcrash::Action>& action, const refract::Registry& registry);
+    NodeInfoByValue<snowcrash::Asset> renderPayloadBody(const NodeInfo<snowcrash::Payload>& payload,
+                                                        const NodeInfo<snowcrash::Action>& action,
+                                                        ConversionContext& context,
+                                                        const refract::Registry& registry);
```
pksunkara: We don't need the registry argument here anymore since we have context.

src/Render.h
```diff
@@ -21,11 +21,20 @@ namespace drafter {
...
+
+    NodeInfoByValue<snowcrash::Asset> renderPayloadSchema(const NodeInfo<snowcrash::Payload>& payload,
+                                                          const NodeInfo<snowcrash::Action>& action,
+                                                          ConversionContext& context,
+                                                          const refract::Registry& registry);
```
pksunkara: We don't need the registry argument here anymore since we have context.

src/RefractDataStructure.cc
klokane: You are right, I forgot to remove it

src/SerializeAST.cc
```diff
@@ -795,16 +803,16 @@ sos::Object WrapBlueprintAST(const Blueprint& blueprint)
...
     // Content
     blueprintObject.set(SerializeKey::Content,
-                        WrapCollection<Element>()(blueprint.content.elements(), WrapElement));
+                        WrapCollection<Element>()(blueprint.content.elements(), WrapElement, context));
 
     return blueprintObject;
 }
 
-sos::Object drafter::WrapBlueprint(const snowcrash::ParseResult<snowcrash::Blueprint>& blueprint, const bool expandMSON)
+sos::Object drafter::WrapBlueprint(const snowcrash::ParseResult<snowcrash::Blueprint>& blueprint, ConversionContext& context, const bool expandMSON)
```
pksunkara: Please move the `expandMSON` argument and `ExpandMSON` global variable into context.

klokane: will do in next PR

src/SerializeResult.cc
```diff
@@ -150,9 +153,10 @@ sos::Object WrapParseResultRefract(snowcrash::ParseResult<snowcrash::Blueprint>&
 }
 
 sos::Object drafter::WrapResult(snowcrash::ParseResult<snowcrash::Blueprint>& blueprint,
-                                const WrapperOptions& options)
+                                const WrapperOptions& options,
+                                ConversionContext& context)
```
pksunkara: We don't need a `context` argument in here. We should build the `context` inside this function.

klokane:
Context && Registry duplicity removed

src/RefractAPI.cc
```diff
@@ -103,7 +120,7 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* CopyToRefract(const NodeInfo<std::string>& copy)
+    refract::IElement* CopyToRefract(const NodeInfo<std::string>& copy, ConversionContext& context)
```
pksunkara: We don't need context here.

src/RefractDataStructure.cc
```diff
@@ -629,7 +634,7 @@ namespace drafter {
         };
 
         template<typename T>
-        void TransformElementData(T* element, ElementData<T>& data) {
+        void TransformElementData(T* element, ConversionContext& context, ElementData<T>& data) {
```
pksunkara: We don't need context here.

src/RefractDataStructure.cc
```diff
@@ -647,7 +652,7 @@ namespace drafter {
     }
 
     template<typename T>
-    refract::IElement* DescriptionToRefract(const ElementData<T>& data)
+    refract::IElement* DescriptionToRefract(const ElementData<T>& data, ConversionContext& context)
```
pksunkara: We don't need context here.

src/RefractDataStructure.cc
```diff
@@ -929,7 +937,7 @@ namespace drafter {
         return select;
     }
 
-    static refract::IElement* MsonMixinToRefract(const NodeInfo<mson::Mixin>& mixin)
+    static refract::IElement* MsonMixinToRefract(const NodeInfo<mson::Mixin>& mixin, ConversionContext& context)
```
pksunkara: We don't need context here.

pksunkara:
Just need to fix a few places where we don't need context. Also, we should build context inside `WrapResult` instead of outside it. Otherwise, looks good.

pksunkara:
Can you also squash this commit into a previous commit? https://github.com/apiaryio/drafter/pull/295/commits/d83d56b46deb6e84181a8e3b5cccb22eca819203

src/RefractAPI.cc
```diff
@@ -194,12 +215,12 @@ namespace drafter {
...
 
-    refract::IElement* ParametersToRefract(const NodeInfo<snowcrash::Parameters>& parameters)
+    refract::IElement* ParametersToRefract(const NodeInfo<snowcrash::Parameters>& parameters, ConversionContext& context)
     {
-        return CollectionToRefract<refract::ObjectElement>(parameters, ParameterToRefract, SerializeKey::HrefVariables, refract::IElement::rFull);
+        return CollectionToRefract<refract::ObjectElement>(parameters, context, ParameterToRefract, SerializeKey::HrefVariables, refract::IElement::rFull);
     }
 
-    refract::IElement* HeaderToRefract(const NodeInfo<snowcrash::Header>& header)
+    refract::IElement* HeaderToRefract(const NodeInfo<snowcrash::Header>& header, ConversionContext& context)
```
w-vi: We don't need it here either.

src/RefractAPI.cc
```diff
@@ -87,7 +104,7 @@ namespace drafter {
         return element;
     }
 
-    refract::IElement* MetadataToRefract(const NodeInfo<snowcrash::Metadata>& metadata)
+    refract::IElement* MetadataToRefract(const NodeInfo<snowcrash::Metadata>& metadata, ConversionContext& context)
```
w-vi: no need for context here

pksunkara: We need it here since we are using it at https://github.com/apiaryio/drafter/pull/295/files#diff-2ab1b5da6ee8cb5b29287e4ebfd30e94R538

src/RefractSourceMap.h
```diff
@@ -37,7 +37,7 @@ namespace drafter {
     }
 
     template<typename T>
-    refract::IElement* LiteralToRefract(const NodeInfo<std::string>& literal)
+    refract::IElement* LiteralToRefract(const NodeInfo<std::string>& literal, ConversionContext& context)
```
w-vi: Context not needed.

src/RefractSourceMap.h
```diff
@@ -10,10 +10,10 @@
 #define DRAFTER_REFRACTSOURCEMAP_H
 
 #include "Serialize.h"
+#include "ConversionContext.h"
```
w-vi: This is not needed too.

src/RefractSourceMap.h
pksunkara: We are using it here. https://github.com/apiaryio/drafter/pull/295/files/ece2ef58b8aea35970247e09490a91aa7c3e7b84#diff-2ab1b5da6ee8cb5b29287e4ebfd30e94R138

w-vi: I don't see it. What I see is that it is called with this parameter not that the parameter is used inside the function, but maybe I am missing something.

pksunkara: It is called with the parameter because of the [`ContextBinder`](https://github.com/apiaryio/drafter/pull/295/files/ece2ef58b8aea35970247e09490a91aa7c3e7b84#diff-2ab1b5da6ee8cb5b29287e4ebfd30e94R43) being used to optimise the code in `NodeInfoToElements`.

Maybe we can overload `NodeInfoToElements` with another function definition without the `context` and thus without the `ContextBinder`

w-vi: I see, bloody templates.

src/RefractDataStructure.cc
pksunkara: Looks like the body has an error, so will keep this argument.

src/RefractAPI.cc
pksunkara: We are using this due to the templates.

klokane:
ConversionContext params removed where are not necessary

pksunkara:
This is the only comment remaining. https://github.com/apiaryio/drafter/pull/295#discussion_r59540243

src/SerializeResult.cc
w-vi: I agree as we will probably expose this function somehow in drafter API.

klokane: In curent time it seems, but we will need merge, but we will need to merge Errors/Warnings at least there https://github.com/apiaryio/drafter/blob/master/src/main.cc#L82

pksunkara: Yes, but I think that should be a part of "cleaning public api" which
should take care of it without doing context building in so many places as
we have in this PR.

klokane: Ok, I will remove it.
Do not care, let it solve who will implementing this :)

pksunkara:
Merged. Please make the PR about expandMSON.

PR #296, opened 2016-04-14
Use conversion context instead of options for interior functions

pksunkara:


src/ConversionContext.h
```diff
@@ -16,7 +16,16 @@ namespace drafter {
         refract::Registry registry;
 
     public:
+        bool generateSourceMap;
```
klokane: why not directly `const WrapperOptions&`?
Will we add duplicit member fer every change in options?

pksunkara: I don't think `WrapperOptions` will survive our public API cleaning. So, we can change this when that happens.

klokane: uh, when I argue by "future" it is wrong and have to remove code :)

but there is another argue - pollution of class interface with every one new member.
It should be as clear as possible.

Related things should grouped.

w-vi:
I think the options and context are pretty close to each other and the public API change will resolve them, that's why I merged it without comment

src/ConversionContext.h
pksunkara: I understand and I thought of actually doing the same thing, but I never liked the `WrapperOptions` and I am pretty sure we will remove it soon.

pksunkara:
Basically, we will merging `WrapperOptions` and `snowcrash::BlueprintOptions` into a single thing and we should use that directly in `Context` when we clean the API

klokane:
At least this members should be `const`...

pksunkara:
I think I will make the PR. That way, we will remember to change this later.

w-vi:
`const`ness yes I overlooked that. good point @klokane

pksunkara:
https://github.com/apiaryio/drafter/pull/297

PR #297, opened 2016-04-14
Move WrapperOptions inside context

pksunkara:


src/ConversionContext.h
```diff
@@ -16,16 +16,12 @@ namespace drafter {
         refract::Registry registry;
 
     public:
-        bool generateSourceMap;
-        bool expandMSON;
+        const WrapperOptions options;
```
klokane: hold as reference

pksunkara:
Updated.

klokane:
:+1:

PR #301, opened 2016-04-15
Added context to SerializeRefract and updated tests

pksunkara:
Just some regular test maintenance. Combined `result` and `parse-result` folder.


klokane:
Just for be sure, all changes in test, are just renamed files, It is correct?
Or output is generated as output after changes?

pksunkara:
@klokane Just renamed some files. And added a few AST, AST Sourcemap and Refract Sourcemap tests for all test cases in `parse-result` fixture folder.

PR #304, opened 2016-04-20
Some errors should not stop parsing.

w-vi:
This probably is not a complete list but there is no way how to trigger
those I leave it as is with the infrastructure in place so we can add
them anytime later.

Couldn't come with a test cases as all the errors are impossible to
trigger if there is not a bug in snowcrash.

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -323,13 +323,19 @@ namespace drafter {
         // ideal solution is add warning into collection
         // but there is no way how to do it
         // in current time we solve it by rethrow
-
         catch (snowcrash::Error& e) {
-            e.location = payload.sourceMap->sourceMap;
-            throw e;
+            context.warnings.push_back(
+                snowcrash::Warning("unable to render JSON/JSONSchema. " + e.message,
+                                   snowcrash::ApplicationError,
+                                   payload.sourceMap->sourceMap));
         }
         catch (refract::LogicError& e) {
-            throw snowcrash::Error(e.what(), snowcrash::ApplicationError, payload.sourceMap->sourceMap);
+            context.warnings.push_back(
+                snowcrash::Warning(std::string(
+                                       "unable to render JSON/JSONSchema. ").append(
+                                           e.what()),
+                                   snowcrash::ApplicationError,
+                                   payload.sourceMap->sourceMap));
         }
 
         RemoveEmptyElements(content);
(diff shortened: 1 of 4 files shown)


src/RefractAPI.cc
```diff
@@ -323,13 +323,17 @@ namespace drafter {
         // ideal solution is add warning into collection
         // but there is no way how to do it
         // in current time we solve it by rethrow
-
         catch (snowcrash::Error& e) {
-            e.location = payload.sourceMap->sourceMap;
-            throw e;
+            context.warnings.push_back(
```
pksunkara: I feel like these warnings should have message. `'unable to render .... content.' + e.message`

src/RefractAPI.cc
```diff
@@ -323,13 +323,17 @@ namespace drafter {
...
+                snowcrash::Warning(e.message,
+                                   snowcrash::ApplicationError,
+                                   payload.sourceMap->sourceMap));
         }
         catch (refract::LogicError& e) {
-            throw snowcrash::Error(e.what(), snowcrash::ApplicationError, payload.sourceMap->sourceMap);
+            context.warnings.push_back(
+                snowcrash::Warning(e.what(),
```
pksunkara: See above comment.

src/SerializeResult.cc
```diff
@@ -131,6 +131,10 @@ sos::Object WrapParseResultRefract(snowcrash::ParseResult<snowcrash::Blueprint>&
     }
 
     snowcrash::Warnings& warnings = blueprint.report.warnings;
+    if (!context.warnings.empty()) {
+        warnings.insert(warnings.end(), context.warnings.begin(), context.warnings.end());
+    }
+
```
pksunkara: Why the extra line?

pksunkara:
Reviewed.

w-vi:
Adressed.

src/RefractAPI.cc
```diff
@@ -323,13 +323,18 @@ namespace drafter {
         // ideal solution is add warning into collection
         // but there is no way how to do it
         // in current time we solve it by rethrow
-
         catch (snowcrash::Error& e) {
-            e.location = payload.sourceMap->sourceMap;
-            throw e;
+            context.warnings.push_back(
+                snowcrash::Warning("Unable to render JSON/JSONSchema: " + e.message,
```
pksunkara: Please use standard format for warnings and errors. Everything in small letters. And use `.` instead of `:`.

Yes, this is something that we are told to follow long ago.

w-vi: But was never written anywhere so any newcomer will not follow it.

klokane:
:+1:

Issue #254, opened 2016-03-22
Using mixins when the referenced object inherits from another object fails


kylef:
When using a mixin via "Include" where the mixin has a parent type (other than directly object) the `Mixin must refer to same type as parent` warning is thrown.

#### Example blueprint to reproduce

```apib
# API
## Resource [/]
### Action [GET]

+ Response 200 (application/json)
    + Attributes
        + Include Example

## Data Structures
### Example (Other)
### Other
```

Note, changing the following would prevent this warning.

```diff
- ### Example (Other)
+ ### Example
```

Drafter `v2.3.0-pre.0` output:

```yaml
element: "parseResult"
content:
  -
    element: "annotation"
    meta:
      classes:
        - "warning"
    attributes:
      code: 4
      sourceMap:
        -
          element: "sourceMap"
          content: []
    content: "Mixin must refer to same type as parent"
```

zdne:
@kylef by

> Note, changing the following ....

You mean removing the `Other` altogether or just to not inherit from `Other` in the definition of `Example`?

kylef:
@zdne to not inherit from another type, other than `object`. So, not inheriting form `Other`. I've updated description to be more clear.

```diff
- ### Example (Other)
+ ### Example
```

pksunkara:
This might be because of another bug. https://github.com/apiaryio/snowcrash/issues/362.

Also related to #255 #243

pksunkara:
Looks like this is an issue with inheritance in drafter

PR #306, opened 2016-04-21
Allow mixin referenced named type to inherit another named type

pksunkara:


src/refract/VisitorUtils.h
```diff
@@ -119,6 +119,41 @@ namespace refract
         }
     };
 
+    template <typename T>
+    void CheckMixinParent(refract::IElement* element)
+    {
+        const T* resolved = TypeQueryVisitor::as<T>(element);
+
+        if (!resolved) {
+            throw snowcrash::Error("mixin base type should be the same as parent base type. objects should contain object mixins. arrays should contain array mixins", snowcrash::MSONError);
```
w-vi: Why are you throwing `snowcrash::Error` and not some `refract` error? We are in `refract` namespace

pksunkara:
@w-vi @klokane This is now ready to be reviewed and merged. Fixes #254

src/refract/VisitorUtils.h
pksunkara: Because this error should be a MSONError.

PR #315, opened 2016-05-05
Serialization by apply visitor

klokane:
Everywhere is removed dependenci on IVisitor only `ApplyVisitor` is used everywhere


w-vi:
I vote to rename the ApplyVisitor to Visitor it will make things easier to read

klokane:
Agree, with rename. I start work on it.
I also remove dependencies on dynamic_cast≤≥ but still no ready for
commit.  Will commit on tomorrow morning.

w-vi:
Adding do not merge label so you can do the renaming

klokane:
Great, thanks

src/refract/TypeQueryVisitor.h
```diff
@@ -90,11 +90,11 @@ namespace refract
             }
 
             TypeQueryVisitor tq;
-            tq.visit(*e);
+            Visit(tq, *e);
 
             E type;
             TypeQueryVisitor eq;
-            type.content(eq);
+            Visitby(eq, type);
```
pksunkara: Shouldn't this be `Visitby(type, eq)`?

klokane: We discuss it with Vilda. Finel solutio is this.
Reason:
`Visit(visitor, element)` - shortcut for `visitor.visit(element)` in meaning "visitor visiting element"
`VisitBy(element, visitor)` - shortcut `element.content(visitor)` in meaning "Element visited by visitor"

pksunkara: Yes, that is why, this should be `VisityBy(type, eq)`. `type` is element and `eq` is visitor here.

pksunkara: Also, it has a type where `b` is a small letter.

src/refract/PrintVisitor.cc
```diff
@@ -10,14 +10,18 @@
...
+#define VISIT_IMPL( ELEMENT ) void PrintVisitor::operator()(const ELEMENT ## Element& e) \
     {                                                                    \
         typedef ELEMENT ## Element::ValueType::const_iterator iterator;  \
         PrintVisitor ps(indent + 1, os);                                 \
+        Visitor visitor(ps);                                             \
                                                                          \
         os << "" #ELEMENT "Element  {\n";                                \
         for (iterator it = e.value.begin(); it != e.value.end(); ++it) { \
-            ps.visit(*(*it));                                            \
+            visitor.visit(*(*it));                                       \
```
pksunkara: Why not use `Visit` or `VisitBy` here.

src/refract/JSONSchemaVisitor.cc
```diff
@@ -160,20 +164,21 @@ namespace refract
...
     {
         JSONSchemaVisitor renderer(pDefs);
+        Visitor visitor(renderer);
 
         if (e.value.second) {
             if (IsTypeAttribute(e, "fixed") || fixed) {
                 renderer.setFixed(true);
             }
-           renderer.visit(*e.value.second);
+            visitor.visit(*e.value.second);
```
pksunkara: Why not use `Visit` or `VisitBy` here?

src/refract/TypeQueryVisitor.h
klokane: There is something wrong it is `Visitby` literaly with small **b**. Probably not used template because there is nowhere `const IElement` used.
Will fix

src/refract/PrintVisitor.cc
klokane: Visitor is defined one layer upper. We need no repatly create Visitor for every one iteration in this case

pksunkara:
Everything looks good now. Will merge once tests pass.

src/refract/JSONSchemaVisitor.cc
klokane: missed, will fix

w-vi:
:+1:  I really like that it has more deleted code than added

Issue #244, opened 2016-02-25
Building Catch fails on modern versions of GCC (5.x)


x2002uwh:
When I run **make test,** I got following errors:

```
In file included from ../test/vendor/Catch/include/internal/catch_expression_lhs.hpp:12:0,
                 from ../test/vendor/Catch/include/internal/catch_result_builder.h:91,
                 from ../test/vendor/Catch/include/internal/catch_capture.hpp:11,
                 from ../test/vendor/Catch/include/catch.hpp:28,
                 from ../test/draftertest.h:4,
                 from ../test/test-drafter.cc:10:
../test/vendor/Catch/include/internal/catch_evaluate.hpp:43:5: warning: identifier ‘nullptr’ is a keyword in C++11 [-Wc++0x-compat]
     inline std::nullptr_t opCast(std::nullptr_t) { return nullptr; }
     ^
In file included from ../test/vendor/Catch/include/internal/catch_expression_lhs.hpp:12:0,
                 from ../test/vendor/Catch/include/internal/catch_result_builder.h:91,
                 from ../test/vendor/Catch/include/internal/catch_capture.hpp:11,
                 from ../test/vendor/Catch/include/catch.hpp:28,
                 from ../test/draftertest.h:4,
                 from ../test/test-drafter.cc:10:
../test/vendor/Catch/include/internal/catch_evaluate.hpp:43:17: error: ‘nullptr_t’ in namespace ‘std’ does not name a type
     inline std::nullptr_t opCast(std::nullptr_t) { return nullptr; }
                 ^
../test/vendor/Catch/include/internal/catch_evaluate.hpp:165:58: error: ‘template<Catch::Internal::Operator Op, class T> bool Catch::Internal::compare’ conflicts with a previous declaration
     template<Operator Op, typename T> bool compare( std::nullptr_t, T* rhs ) {
                                                          ^
../test/vendor/Catch/include/internal/catch_evaluate.hpp:159:44: note: previous declaration ‘namespace Catch::Internal { }::compare’
     template<Operator Op, typename T> bool compare( T* lhs, int rhs ) {
                                            ^
../test/vendor/Catch/include/internal/catch_evaluate.hpp:165:53: error: ‘nullptr_t’ is not a member of ‘std’
     template<Operator Op, typename T> bool compare( std::nullptr_t, T* rhs ) {
                                                     ^
../test/vendor/Catch/include/internal/catch_evaluate.hpp:165:70: error: expected primary-expression before ‘*’ token
[...]

PR #317, opened 2016-05-11
Add -std=c++11 option to cflags

KarelJakubec:
This should be fixing #244 on gcc 5.x version. However I did not test it in other environments, it would probably be worth testing if it does not break builds with older gcc.

Build and tests pass and but it produces bambilion of warnings cause auto_ptr  seems to be deprecated in c++11. Drafter itself works like a charm.

### DIFF
--- a/configure
+++ b/configure
@@ -165,7 +165,7 @@ output = {
   'include_dirs': [],
   'libraries': [],
   'defines': [],
-  'cflags': [],
+  'cflags': [ '-std=c++11' ],
 }
 
 configure_drafter(output)


pksunkara:
We need to test C++11 protagonist on windows with several nodejs versions before deciding to move to C++11. We thank you for the PR and would ask you for a bit of patience while we work through this.

KarelJakubec:
Yeah, no problem, using this I have a working instance of drafter on my system so I am good. :)

w-vi:
It is all on our radar and I personally am looking forward of getting rid of the flawed `auto_ptr`. Soon to come in your computers :)

w-vi:
Closing in favour of merged #331

PR #322, opened 2016-05-19
Fix #246

### LINKED ISSUE #246: Incorrectly parsed object sample
Following blueprint (MSON):

```apib
# Data Structures

## A
+ myArray (array)
    - A
    - B
    - myObject 
          - key1: value1
          - key2: value2 - Nested Object in Array description
```

Gets parsed without warning and `myObject` is considered as an object sample value (??!?)

```yaml
-
  element: "object"
  attributes:
    samples:
      -
        -
          element: "string"
          content: "myObject"
```

Instead the parser MUST warn about unexpected / unrecognized literals `myObject` and ignore them

klokane:
Fix problem with "named object as array member"
e.g:
obj (object)

- Warning from snowcrash
- silently ignore value in conversion

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -474,6 +474,7 @@ namespace drafter {
 
             template <typename S>
             void operator()(S& storage, const NodeInfo<mson::ValueMember>& valueMember) {
+
                 const mson::BaseTypeName type = SelectNestedTypeSpecification(valueMember.node->valueDefinition.typeDefinition.typeSpecification.nestedTypes);
 
                 const RefractElementFactory& elementFactory = FactoryFromType(type);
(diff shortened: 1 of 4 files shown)


src/RefractDataStructure.cc
```diff
@@ -703,13 +738,7 @@ namespace drafter {
 
         if (!value.node->valueDefinition.values.empty() && (valuesCount != data.values.size())) {
```
pksunkara: Was this `if` loop needed? Won't it be covered by https://github.com/apiaryio/drafter/pull/322/files#diff-337a9124ddc8edd098cabd329b901d4bR703?

klokane: Yes it is. They are checking different conditions.
part `valuesCount != data.values.size()` saying, do this only in case  there are some values coming from TypeSections -> move first value into examples

It is posible to remove first part of condition

pksunkara: Ah sorry I misunderstood the second condition. But let's improve the first one.

PR #323, opened 2016-05-19
Fix #318

### LINKED ISSUE #318: Non valid behavior for attribute `- key: value (object)`
For following blueprint
```
# GET /
- Response 200 (application/json)
    - Attribute (A)

# Data Structures

## A
- myObject: abc (object)
```

I will receive message
```
warning: (1)  unable to render JSON/JSONSchema. Invalid member type of object in MSON definition :11:32
```

Expected behavior is report invalid syntax for definition `- myObject: abc (object)`

klokane:
warn on property of type `object` with string value definition e.g:
- key: value (object)

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -736,7 +736,7 @@ namespace drafter {
 
         std::for_each(typeSections.begin(), typeSections.end(), ExtractTypeSection<T>(data, context, value));
 
-        if (!value.node->valueDefinition.values.empty() && (valuesCount != data.values.size())) {
+        if (valuesCount != data.values.size()) {
             // there are some values coming from TypeSections -> move first value into examples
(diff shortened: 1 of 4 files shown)


pksunkara:
Can you please add a unit test to https://github.com/apiaryio/snowcrash/pull/387 before we merge this? Thanks.

klokane:
Another duplicit test?

Additional, with standard `MSONPropertyMemberParser` it will not work because it need context.
And in this case it is not unit test, it will not test just unit.
It will be next integration test...

pksunkara:
Can you update the drafter submodule? And also, you can remove the test from here if you want.

klokane:
Updated,
Test should not be removed from there, because
- it testing more difrerent cases
- it testing if warnings are in result
- changes are related to specialization of `MoveFirstValueToSample` (changed in previous PR)

PR #325, opened 2016-05-23
Fix #312

### LINKED ISSUE #312: Incorrect JSON rendering of named primitive types
Named primitive type, when used as a nested or member type is rendered as its type name instead of its value.

For example, the following blueprint:

```apib
# Rendering Test
## A [/a]
### Retrieve [GET]
- response 200 (application/json)
    - attributes
        - a (array[A Type])

### Data Structures

### A Type (string)
#### Sample
Hello
```

renders the response JSON as:

```json
{
  "a": [
    "A Type"
  ]
}
```

the correct behavior should be:

```json
{
  "a": [
    "Hello"
  ]
}
```

klokane:
Incorrect JSON rendering of named primitive types

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -387,13 +387,15 @@ namespace drafter {
                 RefractElements types;
                 for (mson::TypeNames::const_iterator it = typeNames.begin(); it != typeNames.end(); ++it) {
                     mson::BaseTypeName typeName = it->base;
+                    FactoryCreateMethod method = eValue;
 
                     if (typeName == mson::UndefinedTypeName && !it->symbol.literal.empty()) {
                         typeName = GetMsonTypeFromName(it->symbol.literal, context);
-                    }
+                        method = it->symbol.variable ? eSample : eElement;
+                    } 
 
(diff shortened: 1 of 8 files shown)


src/RefractDataStructure.cc
```diff
@@ -387,13 +387,18 @@ namespace drafter {
...
-                    }
+                        method = it->symbol.variable ? eSample : eElement;
+                    } 
 
                     const RefractElementFactory& f = FactoryFromType(typeName);
-                    types.push_back(f.Create(it->symbol.literal, it->symbol.variable));
+                    types.push_back(f.Create(it->symbol.literal, method));
+
+                    if (!it->symbol.literal.empty()) {
+                    }
```
pksunkara: Empty if condition?

pksunkara:
@klokane Just one comment. Otherwise, looks good.

src/RefractDataStructure.cc
klokane: sorry, it is rest form testing implementation, will remove

pksunkara:
Will merge once tests pass.

PR #329, opened 2016-05-25
Fix #328

### LINKED ISSUE #328: Infinite loop
Following malformed API Blueprint causes infinite loop in drafter.

```
## C [/c]

### Add C([POST])

+ Attributes (C)
```

`DependencyTypeInfo::ResolveType(snowcrash::DataStructure const*) const (NamedTypesRegistry.cc:291)` is the last non-STL code.

klokane:


pksunkara:
The fixtures are confusing. Let's discuss them offline.

test/fixtures/render/issue-328-1.apib
```diff
@@ -0,0 +1,6 @@
+## C [/c]
+
```
pksunkara: Please change this fixture to the following:

```
## C [/c]
## header description
+ Attributes (C)
```

test/test-RenderTest.cc
```diff
@@ -38,3 +38,6 @@ TEST_AST("render", "object-array-string");
 TEST_AST("render", "issue-246");
 TEST_AST("render", "issue-318");
 TEST_AST("render", "issue-312");
+
+TEST_AST("render", "issue-328-1");
+TEST_AST("render", "issue-328-2");
```
pksunkara: Why are these being tested in AST? We should be testing them in refract.

test/fixtures/render/issue-328-2.apib
```diff
@@ -0,0 +1,11 @@
...
+
+## C [/c]
+
+### Add C(`[POST
++ Attributes (C)
+
+## Data Structures
+
+## C (object)
+
```
pksunkara: Please change the fixture to the following

```apib
## C [/c]
## header description
+ Attributes (C)

## Data Structures
### C (object)
```

pksunkara:
My issue with this fix is that before this fix,

```apib
## C [/c]
## header description
+ Attributes (C)
```
 used to give infinite loops and the following was parsing okay

```apib
## C [/c]
non-header description
+ Attributes (C)
```

I don't understand why so. Maybe there is something we are missing.

PR #331, opened 2016-05-30
replace auto_ptr

w-vi:
Updates all dependecies


pksunkara:
I have a few comments on snowcrash PR which you might want to update. Other than waiting for that, this looks good.

pksunkara:
I merged this since we don't need any C++11 compiler changes other than the ruby stuff.

PR #284, opened 2016-04-07
Remove MSON AST related code

pksunkara:
Long time coming.

PR #335, opened 2016-06-02
Release 2.4.0

kylef:


pksunkara:
Are we releasing? Should I delete the branch?

kylef:
After making this, I though we might as well chain these changes into 3.0 with the public Drafter API, since it has breaking change (minimum compiler version). Public Drafter API is being worked on in the current sprint.

pksunkara:
Well if we are not making any release currently, we can merge https://github.com/apiaryio/drafter/pull/284 then.

PR #341, opened 2016-06-16
Klokane/new api

klokane:
Still do not merge
- [x] C-API test
- [x] update doc (updated by Kyle, merge required)
- [x] Fix appveyor build (dllexport/dllimport definitions)
- [x] add valgind check for in CI for C-API


src/API.cc
```diff
@@ -0,0 +1,146 @@
...
+    return ret;
+}
+
+namespace sc = snowcrash;
+
+/* Parse API Bleuprint and return result, which is a opaque handle for
+ * later use*/
+DRAFTER_API int drafter_parse_blueprint(const char* source, drafter_result** out) {
+
+    if (!source) {
```
w-vi: I would just `if (!source || !out) return -1.`

src/API.cc
```diff
@@ -0,0 +1,146 @@
...
+        *stream << "\n";
+        *stream << std::flush;
+    }
+
+}
+
+/* Serialize result to given format*/
+DRAFTER_API const char* drafter_serialize(drafter_result *res, drafter_options options) {
+
+    if (!res) {
```
w-vi: `return NULL`

src/API.cc
klokane: Yes, I remember, we spoke about it, but I did not touch code since this time because of support

w-vi: I know, I just wanted to have it somewhere so we don't forget :)

src/API.h
```diff
@@ -0,0 +1,101 @@
...
+
+typedef struct {
+    bool sourcemap;
+    drafter_format format;
+} drafter_options;
+
+/* Parse API Bleuprint and serialize it to given format.*/
+DRAFTER_API int drafter_parse_blueprint_to(const char* source,
+                               const char ** out,
+                               drafter_options options);
```
pksunkara: Bad indenting.

src/API.h
```diff
@@ -0,0 +1,101 @@
...
+DRAFTER_API const char* drafter_serialize(drafter_result *res, drafter_options options);
+
+/* Free memory alocated for result handler */
+DRAFTER_API void free_drafter_result(drafter_result* result);
+
+/* Parse API Blueprint and return only annotations, if NULL than
+ * document is error and warning free.*/
+DRAFTER_API drafter_result* drafter_check_blueprint(const char* source);
+
+DRAFTER_API unsigned int drafter_c_version(void);
```
pksunkara: Why the `_c_` here and the below function? Can't they be `drafter_version` and `drafter_version_string`?

src/SerializeResult.h
```diff
@@ -16,10 +16,19 @@ namespace snowcrash {
...
+
 namespace drafter {
 
+    struct ConversionContext;
+
     sos::Object WrapAnnotation(const snowcrash::SourceAnnotation& annotation);
     sos::Object WrapResult(snowcrash::ParseResult<snowcrash::Blueprint>& blueprint, const WrapperOptions& options);
+
+    refract::IElement* WrapParseResultRefract(snowcrash::ParseResult<snowcrash::Blueprint>& blueprint,
+                                   ConversionContext& context);
```
pksunkara: Bad indentation.

src/refract/Query.h
```diff
@@ -18,20 +18,27 @@ namespace refract
...
+
+        public:
+
+            ByName(const std::string& name) : name(name) {}
 
-            operator bool() const {
-                return is;
+            bool operator()(const IElement& e) {
+                return false;
+                //return e.element() == name;
```
pksunkara: What's with this line?

klokane:
OMG there is written in PR comment, it is intended as reference for documentation.
Do not care about indenting, return values and anything other at this time...

pksunkara:
Ah, sorry, didn't realise that.

src/API.h
```diff
@@ -0,0 +1,101 @@
+//
+//  API.h
```
kylef: I think we should call this header `drafter.h` so people will include `drafter.h` and not `API.h` when using this library. This will be clearer and consistent with other C/C++ libraries.

Examples:

- Curl (`#include <curl/curl.h>`)
- gnutls (`#include <gnutls/gnutls.h>`)

klokane: Agree it is in PR comment

kylef: Sorry, glanced over that 👍

src/API.h
```diff
@@ -0,0 +1,101 @@
...
+} drafter_format;
+
+typedef struct {
+    bool sourcemap;
+    drafter_format format;
+} drafter_options;
+
+/* Parse API Bleuprint and serialize it to given format.*/
+DRAFTER_API int drafter_parse_blueprint_to(const char* source,
+                               const char ** out,
```
kylef: Shouldn't this just be `char **out`? The caller won't be able to free a const char?

src/drafter.h
```diff
@@ -1,37 +1,101 @@
...
+
+/* Parse API Bleuprint and return result, which is a opaque handle for
+ * later use*/
+DRAFTER_API int drafter_parse_blueprint(const char* source, drafter_result** out);
+
+/* Serialize result to given format*/
+DRAFTER_API const char* drafter_serialize(drafter_result *res, drafter_options options);
+
+/* Free memory alocated for result handler */
+DRAFTER_API void free_drafter_result(drafter_result* result);
```
kylef: I feel like this function should be called `drafter_free_result` so it has the same prefix as the others. Drafter keyword is always at the start. This also allows users to "tab"-complete drafter functions consistently.

Any thoughts?

w-vi: Agreed, all public functions must be in `drafter_*` form

klokane: agree, will fix.

src/drafter.h
```diff
@@ -1,37 +1,102 @@
...
+    DRAFTER_SERIALIZE_YAML = 0,
+    DRAFTER_SERIALIZE_JSON
+} drafter_format;
+
+typedef struct {
+    bool sourcemap;
+    drafter_format format;
+} drafter_options;
+
+/* Parse API Bleuprint and serialize it to given format.*/
```
kylef:
```diff
- Bleuprint
+ Blueprint
```

src/drafter.h
```diff
@@ -1,37 +1,102 @@
...
+DRAFTER_API int drafter_parse_blueprint_to(const char* source,
+                               char ** out,
+                               const drafter_options options);
+
+/* Parse API Bleuprint and return result, which is a opaque handle for
```
kylef:
```diff
- Bleuprint
+ Blueprint
```

src/drafter.cc
```diff
@@ -1,22 +1,192 @@
...
+namespace sc = snowcrash;
+
+/* Parse API Bleuprint and return result, which is a opaque handle for
+ * later use*/
+DRAFTER_API int drafter_parse_blueprint(const char* source, drafter_result** out) {
+
+    assert(source);
+    assert(out);
+
+    if (!source) {
```
kylef: Shouldn't this check out is not NULL too?

w-vi: There is assert so I think even this check is not needed. Depends on approach.

kylef: The assert is compiled out during production. I think this function shouldn't crash during invalid inputs in production, only when compiled for debug. At least, that's what happens for `source` and parameters in the other public API functions.

w-vi: This a matter of taste, I think sending NULL pointer is a programmer's error and assert is for that.

kylef: I'm happy either way, as long as it's consistent.

The other functions use both assert and if to return `-1` for bad input when assert is compiled with `NDEBUG`.

```c
DRAFTER_API int drafter_parse_blueprint_to(const char* source, char ** out, const drafter_options options) {
    assert(source);
     assert(out);

     if (!source || !out) {
        return -1;
    }
```

w-vi: I agree, @klokane choose your style and keep it everywhere, using either assert or check for nullptr.

w-vi:
@klokane  If using the checks like `if (!out || !source)` please remove the asserts, as they make the behaviour different in debug and release.

src/drafter.cc
```diff
@@ -2,21 +2,196 @@
...
+
+#include <string.h>
+
+#include <assert.h>
+
+DRAFTER_API int drafter_parse_blueprint_to(const char* source,
+                                           char ** out,
+                                           const drafter_options options) {
+
+    assert(source);
```
w-vi: I am talking about these

klokane:
@w-vi `if()` check there should stay there because NULL can lead to crash.
`assert()` expanded to nothing if NDEBUG is not define.

w-vi:
So it behaves differently if NDEBUG is defined or not, that's my point. If id debug I get abort() if in release  I get `return -1` I should get `-1` in bot cases or `abort`

klokane:
@w-vi I mean `assert()` is way how to say to programmer, you are doing something wrong way. Just check required preconditions. 
While check against NULL is protect library against crash

w-vi:
I have big problem with the inconsistency. I am unable to test for NULL input because I will get `abort` via `assert` in case of Debug and in Release I will not get `abort` but will see `-1`.

w-vi:
:+1:

w-vi:
I would merge it but the circle CI is stuck somehow ...

src/drafter.cc
```diff
@@ -2,21 +2,185 @@
...
+    sc::ParseResult<sc::Blueprint> blueprint;
+    sc::parse(source, snowcrash::ExportSourcemapOption, blueprint);
+
+    drafter::WrapperOptions options(drafter::RefractASTType);
+    drafter::ConversionContext context(options);
+    refract::IElement* result = WrapParseResultRefract(blueprint, context);
+
+    *out = result;
+
+    *out = result;
```
pksunkara: What? 2 times the same thing?

klokane: mistake, will make patch

PR #342, opened 2016-06-21
Use fixed from top-level mson element for JSON Schema

pksunkara:


pksunkara:
Fixes https://github.com/apiaryio/mson/issues/60

test/test-RenderTest.cc
```diff
@@ -41,3 +41,6 @@ TEST_AST("render", "issue-312");
 
 TEST_AST("render", "issue-328-1");
 TEST_AST("render", "issue-328-2");
+
+TEST_AST("render", "fixed-attributes-section");
+TEST_AST("render", "fixed-named-type")
```
klokane: no newline

klokane:
Just formal problem, other things looks good

pksunkara:
@klokane Updated.

w-vi:
Shouldn't the tests be in schema section rather then render? Render is for JSON and schema for JSON Schema

pksunkara:
Ah, I didn't realise we had a separate folder for schema which is weird because they can still be considered render.

w-vi:
Well it is but it has different code tested so that's why it is separated, even though I think we should at some point improve the testing so that we need less fixtures covering more. Focused on the language constructs and not the results as we test many times the same blueprints in different places for different bits of the output.

PR #344, opened 2016-06-23
Update Xcode for Drafter

kylef:
This updates the Xcode project for your changes @klokane

### DIFF
--- a/src/refract/Query.h
+++ b/src/refract/Query.h
@@ -11,6 +11,7 @@
 #include "ElementFwd.h"
 
 #include <functional>
+#include <string>
 
 namespace refract
 {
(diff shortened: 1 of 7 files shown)


src/refract/Query.h
```diff
@@ -11,6 +11,7 @@
 #include "ElementFwd.h"
 
 #include <functional>
+#include <string>
```
kylef: For some reason, without this I get a build error since std::string (used below) is not defined.

```
/Users/kyle/Projects/apiaryio/drafter/src/refract/Query.h:24:31: Implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
```

Not sure how this works for you? Unless functional in some compilers include string too.

klokane: Have no idea why. It is first time when I meet similar request.

klokane:
Thanks, will merge on top, after final review

PR #345, opened 2016-06-23
Fix memory leak while failing to open file

kylef:
I know we're exiting right away so it's not really a problem, but static analysers and such tools flag this as a leak.


w-vi:
@kylef can you make it one commit, please?

kylef:
@w-vi Can you merge & squash in GitHub UI? Would be easier

w-vi:
I have rebased it onto master but now getting the error as on the conversion context PR, @klokane  can you have a look on it?

klokane:
@w-vi @kylef windows problem build fixed (Kyle, you probably need rebase all latest commits on top of this, once merged)

klokane:
@w-vi pls merge

PR #346, opened 2016-06-24
Document new API in README

kylef:
I've re-ordered the usage section so that the command line tool is demonstrated first. I'd expect most readers are coming to Drafter to use it as a command line tool.

@klokane

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ Additionally Drafter provide set of Wrappers for serialization, of parsing resul
 
 Drafter also provides the user ability to select the type of the output. There are two possible values:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -106,6 +67,85 @@ content:
...
+  "## GET /message\n"
+  "+ Response 200 (text/plain)\n"
+  "\n"
+  "      Hello World!\n";
+
+drafter_options options;
+options.format = DRAFTER_SERIALIZE_JSON;
+options.sourcemap = true;
+
+char* result;
```
klokane: Good practice is automatic initiate defined pointer to NULL

PR #351, opened 2016-06-28
ConversionContext is a class not struct

kylef:
```
In file included from ../src/main.cc:22:
../src/ConversionContext.h:18:5: warning: 'ConversionContext' defined as a class here but previously declared as a struct
      [-Wmismatched-tags]
    class ConversionContext {
    ^
../src/SerializeResult.h:25:5: note: did you mean class here?
    struct ConversionContext;
    ^~~~~~
    class
```

### DIFF
--- a/src/SerializeResult.h
+++ b/src/SerializeResult.h
@@ -22,7 +22,7 @@ namespace refract {
 
 namespace drafter {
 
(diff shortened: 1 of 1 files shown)


klokane:
Rebased on fixed win build, will merge after CI

PR #338, opened 2016-06-09
Remove drafter.js from repository

kylef:
We're moving the emscripten build of Drafter from this repository into it's own repository. Please see https://github.com/apiaryio/drafter.js/pull/41.

PR #354, opened 2016-06-30
Fix JSON typo in drafter.js example.

nfisher:


kylef:
Thanks for PR, but actually these instructions shouldn't even be here. We've split this into a separate drafter.js repo as per #338. Looks like I forgot to remove the instructions from there in #338.

I'll remove them from here in a PR and update it in drafter.js (https://github.com/apiaryio/drafter.js#protagonist) shortly.

Thanks for spotting this @nfisher.

PR #362, opened 2016-07-14
Use API intenal

klokane:


src/reporting.h
```diff
@@ -22,5 +23,17 @@ void PrintReport(const snowcrash::Report& report,
                  const std::string& source,
                  const bool isUseLineNumbers);
 
+/**
+ *  \brief Print parser report to stderr.
+ *
+ *  \param report A parser report to print
+ *  \param source Source data
+ *  \param isUseLineNumbers True if the annotations needs to be printed by line and column number
```
kylef: I think the `is` here feels misplaced. `useLineNumbers` would be clear on what the name is and I think I'd prefer the name `useLineNumbers`.

klokane: will remove it is copy'n'paste definition from legacy code

src/main.cc
```diff
@@ -78,10 +74,63 @@ int main(int argc, const char *argv[])
...
+                 << "\n"
+                 << std::flush;
+
+            free(output);
+        }
+    }
+
+    PrintReport(result, inputStream.str(), config.lineNumbers, ret);
+
+    delete result;
```
kylef: Shouldn't we use our public API here? `drafter_free_result`?

src/reporting.cc
```diff
@@ -159,3 +158,106 @@ void PrintReport(const snowcrash::Report& report,
...
+
+    const std::string operator()(const refract::IElement* annotation) {
+        std::stringstream output;
+
+        if (annotation->element() != "annotation") {
+            return output.str();
+        }
+
+        if (const refract::ArrayElement* classes = GetMemberValue<refract::ArrayElement>(annotation->meta, "classes")) {
+            if (classes->value.size() == 1) {
```
kylef: I guess there should never be a case where there are multiple classes since we currently use warning and error. But perhaps it would be a good idea to print to stdout if this assumption isn't true, or perhaps print all the classes comma separated? Just incase we ever add something here, it will be silently unhandled.

src/reporting.cc
```diff
@@ -159,3 +158,106 @@ void PrintReport(const snowcrash::Report& report,
         PrintAnnotation("warning:", *it, source, isUseLineNumbers);
     }
 }
+
+template <typename T>
```
w-vi: I think this could go to VisitorUtils as it is a generic function working with refract.

w-vi:
I really like how the code is now clearer.

src/main.cc
klokane: `drafter_free_result` is intended for C because you can not invoke `delete` from C directly.
But on other side it is true if we will use `drafter_free_result` it will look more consistent.
I've no strong opinion about it. Let vote :)

src/reporting.cc
klokane: Agree, but it is question what to do. 
There is more similar places in code while conversion annotation to string

src/reporting.cc
klokane: agree, will move

src/main.cc
w-vi: If one day `drafter_free_result` is more complicated then simple `delete` we would need to fix all of those. Another reason is that this code is an example of usage of the API and it should be the best practice so using the API properly will help others.

pksunkara:
👍

pksunkara:
We should have removed APIB AST from CLI when we moved to 3.0. 😞

kylef:
> We should have removed APIB AST from CLI when we moved to 3.0. 😞

I think we should hold off for a little while, since we are using this format at Apiary it's useful to be able to get hold of it easily on the command line.

klokane:
@pksunkara it is not good idea, until APIB AST is obsoleted and removed from all Apiary service

PR #368, opened 2016-07-29
Use default/sample enum values over members when rendering JSON

kylef:
The following precedence of enum examples with rendering will now be taken:

- Sample Values
- Default Values
- Values
- Nullable

### DIFF
--- a/src/refract/VisitorUtils.h
+++ b/src/refract/VisitorUtils.h
@@ -125,6 +125,32 @@ namespace refract
         }
     };
 
+    struct GetEnumValue {
+        const EnumElement& element;
+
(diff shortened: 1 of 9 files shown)


src/refract/VisitorUtils.h
```diff
@@ -125,6 +125,32 @@ namespace refract
         }
     };
 
+    struct GetEnumValue {
```
kylef: Is there anyway to overload `GetValue` for enum element? I tried and failed, but maybe a trick I'm not aware of.

pksunkara: I think you can. But I will let @klokane answer this. Otherwise the PR looks good.

PR #373, opened 2016-08-11
Add warnings from context to blueprint.report for ast

pksunkara:


src/SerializeResult.cc
```diff
@@ -73,6 +73,10 @@ sos::Object WrapParseResultAST(snowcrash::ParseResult<snowcrash::Blueprint>& blu
         }
     }
 
+    for (auto& warning : context.warnings) {
```
w-vi: Is auto generally supported in the compilers we require?

src/main.cc
```diff
@@ -51,7 +51,8 @@ void Serialization(std::ostream *stream,
     *stream << std::flush;
 }
 
-int ProcessAST(const Config& config, std::unique_ptr<std::istream>& in, std::unique_ptr<std::ostream>& out) {
```
w-vi: I thought that we use `func(prams...) {` so the opening bracket is on the same line

pksunkara: This file had them on new line except for these 2 functions, so I moved them for consistency, but it will be a moot point soon.

src/SerializeResult.cc
pksunkara: I am not sure. Tests passed. How do I check this?

src/main.cc
kylef: Yeah, are we happy with https://github.com/apiaryio/snowcrash/pull/380 ? If there isn't any problems I'd like to move forward with this format and I will recreate my PR from master.

pksunkara: Can we do that **just** before the release? There are several things we are working on which might be messed up.

kylef: 👍 Yes, let's settle on format and I'll push the PR through at the end of a sprint when we've got no open work.

pksunkara:
Just checked. Auto is supported everywhere.

src/SerializeResult.cc
klokane: what about:
```
std::copy(context.warnings.begin(), context.warnings.end(), std::back_inserter(blueprint.report.warnings));
```

It is supported long time in C++ and we can avoid auto until we check support in all compilers

klokane:
Just hint about `auto`, but it is not blocking

klokane:
All other seem to be OK

pksunkara:
@klokane As per my previous comment, I just checked all our compiler versions. It's supported everywhere.

PR #375, opened 2016-08-17
Added check for contradictory type attributes

pksunkara:


src/RefractDataStructure.cc
```diff
@@ -917,11 +917,49 @@ namespace drafter {
...
     };
 
+    static void CheckTypeAttributesClash(const mson::TypeAttributes& attributes,
+                                         const snowcrash::SourceMap<mson::ValueDefinition>& sourceMap,
+                                         ConversionContext& context) {
+
+        if ((attributes & mson::FixedTypeAttribute) != 0 &&
+            (attributes & mson::OptionalTypeAttribute) != 0) {
+
+            context.warn(snowcrash::Warning("cannot have 'fixed' and 'optional' together",
```
klokane: propose: s/have/use

klokane: cannot -> can not

klokane: But i'm not so strong in English :)

pksunkara: `cannot` is the better way to use it.

pksunkara:
Updated.

src/RefractDataStructure.cc
```diff
@@ -917,11 +917,49 @@ namespace drafter {
...
+        }
+    }
+
     template <typename Trait>
     static refract::IElement* MsonMemberToRefract(const typename Trait::InputType& input,
                                                   ConversionContext& context,
                                                   const mson::BaseTypeName nameType,
-                                                  const mson::BaseTypeName defaultNestedType) {
+                                                  const mson::BaseTypeName defaultNestedType,
+                                                  bool checkTypeAttributes = true) {
```
klokane: I mean it is not good idea add param for "checking". It will expand again and again with another one check.

pksunkara: Hmm.. no actually. If you look below, you can see why sometimes this check happens more than once for the same MSON member. This param makes sure it doesn't.

w-vi: Is it a big issue to do the check more than once? Probably not and in the future we are likely  do it differently.

pksunkara: If the check happens more than once, we will get duplicate warnings. I guess, with the introduction of `context.warn` we can remove this part but logically this is still correct. What do you guys think?

w-vi: We won't get duplicate warnings because of the `context.warn`, having these flags passed around is correct but harder to read, not having it makes it simpler. But leaving to @klokane to make the last call.

klokane: Too many params typical shows there is probably something wrong. But ok let it pass.

src/Serialize.h
```diff
@@ -239,7 +239,7 @@ namespace drafter {
         }
 
         template<typename Collection, typename Functor, typename Arg1, typename Arg2>
-        R operator()(const Collection& collection, Functor &wrapper, Arg1 arg1, Arg2 arg2) const {
+        R operator()(const Collection& collection, Functor &wrapper, Arg1 arg1, Arg2& arg2) const {
```
klokane: Why arg1 is send by value and arg2 by reference? Make it same way please.

src/ConversionContext.h
```diff
@@ -26,6 +26,33 @@ namespace drafter {
 
         ConversionContext(const WrapperOptions& options)
         : options(options) {}
+
+        void warn(const snowcrash::Warning& warning)
```
klokane: Please introduce `cc` file and move function body into.

src/Serialize.h
pksunkara: Won't work. We need the first arg1 to be by value since it was passed pointers for the payload sections. And yes, its necessary and no, we can't change that.

klokane: And reason why must be arg2 by reference?
There is option use `std::ref()` if reference is required,

pksunkara: Because this arg is used for `context` and if we don't pass it by reference then the warnings we add will just disappear.

klokane: you can send context via `std::ref()` it should work. It is not good idea have template which expect some special cases without specialization.

w-vi: Oh yeah, we are C++11 so `std:ref` is there and it is supported in the compiler versions, cool.

pksunkara:
@klokane Updated

PR #376, opened 2016-08-18
Warning on invalid value for Boolean/Number 

klokane:
Not yet finished - do not merge.

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -1,3 +1,4 @@
+//
 //  RefractDataStructure.cc
 //  drafter
 //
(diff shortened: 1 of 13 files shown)


src/RefractDataStructure.cc
```diff
@@ -129,7 +129,7 @@ namespace drafter {
         void operator()(const ValueInfo& value, ConversionContext& context) {
 
             if (!std::get<2>(value)) {
-                context.warn(snowcrash::Warning("invalid value format for 'number' type. Valid format is as specifified in 'C' function 'strtod()' ", snowcrash::MSONError, std::get<1>(value).sourceMap));
+                context.warn(snowcrash::Warning("invalid value format for 'number' type. Valid format is as specifified in 'C language' function 'strtod()' ", snowcrash::MSONError, std::get<1>(value).sourceMap));
```
kylef: Perhaps it would be best to say the format is found within the MSON Specification, and we add a section on what is a valid number. Not everyone that uses API Blueprint is a programmer.

klokane: Ok will change

klokane: what about just part "invalid value format for 'number' type"?

kylef: That sounds fine.

src/RefractDataStructure.cc
```diff
@@ -620,54 +628,60 @@ namespace drafter {
...
-                // if more default values
-                // use last one, all other we will drop
-                element->attributes[SerializeKey::Default] = e;
+            template <typename U>
+            void operator()(const U& defaults, refract::IElement* element) {
+                if (defaults.empty()) {
+                    return;
+                }
 
-                std::for_each(defaults.begin(), defaults.end(), Deleter<refract::IElement>);
```
pksunkara: @klokane I don't see the logic related to this `Deleter` in the new code.

klokane: what are you asking for? 
`Deleter` can be removed from code? - yes, it can

Why it is not required since now?
It is because `ElementData` now used vector of "ValueType" to hold default and samples instead of creating refract elements directly "Extract Functors". Elements (for default samples) are now created later while `TransformElementData` is invoked.

pksunkara: Okay.

src/RefractSourceMap.h
```diff
@@ -40,7 +40,7 @@ namespace drafter {
     template<typename T>
     refract::IElement* LiteralToRefract(const NodeInfo<std::string>& literal, ConversionContext& context)
     {
-        std::pair<bool, const T&> parsed = LiteralTo<T>(*literal.node);
+        std::pair<bool, T> parsed = LiteralTo<T>(*literal.node);
```
pksunkara: @klokane This is the one causing tests to fail on windows.

klokane: :+1:

PR #393, opened 2016-09-16
Add support for using MSON with URL Encoded Forms

kylef:
This pull request adds support for using `application/x-www-form-urlencoded` with MSON Data structures in requests and responses.

Example:

``` apib
# GET /

+ Response 200 (application/x-www-form-urlencoded)
    + Attributes
        + name: Kyle
```

Which gets rendered as:

```
name=Kyle
```


src/refract/RenderFormURLEncodedVisitor.cc
```diff
@@ -0,0 +1,162 @@
...
+        if (value && !value->empty()) {
+            operator()(*value->front());
+        } else {
+            result << "null";
+        }
+
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const ArrayElement& e) {
+        // Rendering an array is not supported
```
kylef: I was looking at how to add a warning for this case but I'm not sure what the best way to be. We wouldn't have source maps here?

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Master
 
+### Enhancements
+
+* `application/x-www-form-urlencoded` is not supported as an MSON serialisation
```
kylef: now*

src/refract/RenderFormURLEncodedVisitor.cc
```diff
@@ -0,0 +1,201 @@
...
+                Visit(visitor, *itElement);
+                result << visitor.getString();
+            }
+        }
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const NullElement& element) {
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const StringElement& element) {
```
w-vi: Can you simplify this using template, something along the lines like in JSONSchema https://github.com/apiaryio/drafter/blob/master/src/refract/JSONSchemaVisitor.cc#L157

kylef: Addressed.

src/refract/RenderFormURLEncodedVisitor.cc
```diff
@@ -0,0 +1,198 @@
...
+    void RenderFormURLEncodedVisitor::operator()(const IElement& element) {
+        VisitBy(element, *this);
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const MemberElement& element) {
+        if (!element.value.first) {
+            return;
+        }
+
+        StringElement *keyElement = static_cast<StringElement*>(element.value.first);
```
kylef: This needs to be updated to handle the case where there isn't a `StringElement` but instead a `SelectElement`.

```apib
# GET /

+ Response 200 (application/x-www-form-urlencoded)
    + Attributes
        + One Of
            + foo
            + bar
```

src/refract/RenderFormURLEncodedVisitor.cc
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    RenderFormURLEncodedVisitor::~RenderFormURLEncodedVisitor() {
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const IElement& element) {
+        VisitBy(element, *this);
+    }
+
+    void RenderFormURLEncodedVisitor::operator()(const MemberElement& element) {
```
pksunkara: As discussed on hangout, I think we should skip serialising objects for now.

src/refract/RenderFormURLEncodedVisitor.cc
```diff
@@ -0,0 +1,198 @@
...
+                 ++it) {
+                if (!*it) {
+                    continue;
+                }
+
+                if (!result.str().empty()) {
+                    result << "&";
+                }
+
+                RenderFormURLEncodedVisitor visitor(baseKey + "[]");
```
pksunkara: According to RFC 6570, the serialisation should be `list=green&list=red&list=blue`. No `[]` anywhere. If we have doubts regarding this, we can skip arrays for now. But, I would just remove the brackets.

PR #396, opened 2016-09-19
bugfix: warn on action with non-absolut URI

klokane:
Fix#350

### DIFF
--- a/test/fixtures/syntax/issue-350.apib
+++ b/test/fixtures/syntax/issue-350.apib
@@ -1,0 +1,8 @@
+## Users [/users]
+### View a User [GET {username}]
+
++ Parameters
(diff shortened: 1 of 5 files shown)


test/fixtures/syntax/issue-350.apib
```diff
@@ -0,0 +1,8 @@
+## Users [/users]
+### View a User [GET /{username}]
```
kylef: Shouldn't this fixture not have an absolute path here?

klokane: sorry you are right, I made some "tests" and i forgot to change it back

klokane: Updated

PR #397, opened 2016-09-21
no sourcemap for references Fix #386

### LINKED ISSUE #386: no sourcemap for references
Given that `xxx` is the reference to `xxx` dataStructure object

#### Current
The dataStructure where `xxx` is only object has no sourcemaps

```
{
   "element":"dataStructure",
   "meta":{

   },
   "attributes":{

   },
   "content":[
      {
         "element":"xxx",
         "meta":{

         },
         "attributes":{

         }
      }
   ]
}
```

#### Expected
The dataStructure where `xxx` is only object has  sourcemaps

```
{
   "element":"dataStructure",
   "meta":{

   },
   "attributes":{

   },
   "content":[
      {
         "element":"xxx",
         "meta":{

         },
         "attributes":{
            "sourceMap":[
               {
                  "element":"sourceMap",
                  "meta":{

                  },
                  "attributes":{

                  },
                  "content":[
                     [
                        791,
                        40
                     ]
                  ]
               }
            ]
         }
      }
   ]
}
```

klokane:


src/RefractAPI.cc
```diff
@@ -300,6 +300,7 @@ namespace drafter {
 
         content.push_back(CopyToRefract(MAKE_NODE_INFO(payload, description)));
         content.push_back(DataStructureToRefract(MAKE_NODE_INFO(payload, attributes), context));
+        //content.push_back(DataStructureToRefract(MakeNodeInfo(payload.node->attributes, payload.sourceMap->attributes), context));
```
klokane: forgoted will remove

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Master
 
+* Add missing sourcemaps for references
```
pksunkara: `for named type references`

klokane:
Damn forgot update integration test

CHANGELOG.md
klokane: fixed

pksunkara:
Please don't merge this without my approval. I think there's a small issue here and I am still looking into it.

klokane:
@pksunkara status?

features/fixtures/refract.sourcemap.json
```diff
@@ -334,6 +334,19 @@
...
+                            "element": "sourceMap",
+                            "content": [
+                              [
+                                173,
+                                20
+                              ]
+                            ]
+                          }
+                        ]
+                      },
```
pksunkara: This is a source map for the `object` keyword in `+ Attributes (object)`.  I don't think we need a source map here.

test/fixtures/api/action-attributes.sourcemap.json
```diff
@@ -119,6 +119,19 @@
...
+                                "element": "sourceMap",
+                                "content": [
+                                  [
+                                    73,
+                                    11
+                                  ]
+                                ]
+                              }
+                            ]
+                          },
```
pksunkara: This is a source map for the `object` keyword in `+ Attributes` even though it is not defined.  I don't think we need a source map here.

pksunkara: This is occurring at several place below too.

klokane:
@tu1ly can you please look at? 

Source:
https://github.com/apiaryio/drafter/blob/295f957b3bb6da3c1e067b97a7af69c216dfff0c/test/fixtures/api/issue-386.apib

Result:
https://github.com/apiaryio/drafter/blob/295f957b3bb6da3c1e067b97a7af69c216dfff0c/test/fixtures/api/issue-386.sourcemap.json

klokane:
@pksunkara reason?

pksunkara:
@klokane Forget about my previous comment about `+ Attributes (object)` or base types. But let's not add source maps to `object` element when the user give `+ Attributes` with implied base type.

pksunkara:
Reason for that is that since it's implied and generated by us and not the user, we don't want to give source maps there.

klokane:
But untyped `+ Attributes` is implicit object by Blueprint/MSON definition

pksunkara:
Yes, but since the user didn't write the word `object` there, we shouldn't add source map for it.

klokane:
I disagree, I do not see difference while `object` is not defined `explicitly` by type or implicitly just by `Attributes` keyword. In both cases it is still object and is defined by user.

pksunkara:
Look, we are not adding source maps a lot of places where user implicitly defined something. That is our policy from the beginning. We can't just change it in the middle just for 1 feature. If we want to add source maps for the implicitly defined things, then we need add another them completely everywhere which is out of scope for now.

klokane:
Ok, so where is source map for keyword `Attributes` strictly defined by user?

pksunkara:
It isn't. Just like we don't have source maps for `Headers` keyword, or `Group` keyword. They are API Blueprint stuff which are not transformed into anything in Refract. Which is why they don't have source map.

As I said, let's follow this for now, and if we want to chang this, let's make it separate.

klokane:
Sure, but  `Headers` or `Groups` are not converted to MSON object opposed to `Attributes`. I do not see reason for higher complexity of code to show difference between implicit and explicit object definition.

pksunkara:
`Attributes` is not converted into MSON. The `(object)` is quietly added which is converted into MSON.

klokane:
`Attributes` itself is object defined by implicit, (if user do not specify otherwise).

pksunkara:
No. Attributes is an APIB keyword. Not an MSON keyword.

klokane:
Agree, Attributes is part od Blueprint spec (not MSON, that is correct). But by definition `Attributes` can not be anything other than MSON. 

> section describes a data structure using the Markdown Syntax for Object Notation (MSON). 

In other words: It can not be nothing other than MSON, and if you do not specify type explicitly, it is always MSON `object`.

klokane:
In any time when user specify `Attributes` in blueprint, he automaticly specify MSON object

pksunkara:
But he didn't actually write that word. So,  no source map

klokane:
But he wrote `Attribute` it is `object`.

klokane:
By your argue, we should remove ALL properties/members where user do not specify type and it is chosen implicitly (by definition `object` or `string`)

pksunkara:
No, that would be data structure element

klokane:
Similar for DataStructure section. ANY named type without specified type, there use do not write `object`

pksunkara:
No, I am taking about source maps only. Not the actual data

klokane:
Yes me too, everywhere where is not specified explicitly type we should remove source map...

pksunkara:
Yes and we do not provide source map there and similarly we should not here

klokane:
So, your position is "remove sourcemap evereywhere user do not specify MSON type"?

pksunkara:
Let me clarify more:

```
+ Attributes (object)
```

is converted into `dataStructure` element which contains `object` element. You have added source map for `object` which is okay. The `dataStructure` element doesn't have a source map which is a bug.

If the user writes:

```
+ Attributes
```

It is converted into `dataStructure` element which contains `object` element. We shouldn't add source map for `object`. The `dataStructure` element should contain the source map of the `Attributes` keyword. **NOT THE OBJECT ELEMENT**

pksunkara:
> So, your position is "remove sourcemap evereywhere user do not specify MSON type"?

Yes, and we don't need to remove it currently anywhere because we have been following this since the beginning of MSON AST 2 yrs ago.

klokane:
so you mean, this definition should contain sourcemaps:
```
+ a (object)
    + m1 (string)
```
While this one not?
```
+ a
    + m1
```

pksunkara:
Only the element types. The actual member types would still have source
maps

klokane:
Ok so this cases:
```
# Data Structures

# #A (object)

## B
``` 

`A` should, while `B` not (user did not specify it is object, it is by implicit definition)?

pksunkara:
Yes. But understand that meta.id would still have source map here

klokane:
Nope. That is weird, why should tool author check two places for sourcemap of MSON object, just because one object is explicitly defined while second one is implicitly defined?

klokane:
And for both cases `A` and `B` too, we actually generate  sourcemaps.

w-vi:
I have written `## B` so I expect a sourcemap for this.

kylef:
I agree with @w-vi and @klokane here. If I as a user write:

```apib
## Data Structures
### A (object)
### B
``` 

Then both should have a source map. B is an object by default. It's documented in the API Blueprint / MSON specification that `### B` defines a type in this circumstance which inherits from `object`. So the syntax that a user writing `### B` inherits from an object should have a source map.

---

I think we should move ahead with this pull request. Is there anything else blocking it?

pksunkara:
Yes, you will get a source map for `## B` in `dataStructure` element and `meta.id`. But not in the `object` element.

That said, I still need to work through a few questions from @klokane

pksunkara:
From @klokane's earlier comment:

> so you mean, this definition should contain sourcemaps:
```
+ a (object)
    + m1 (string)
```
While this one not?
```
+ a
    + m1
```

Both of them give the same parse result because we don't give out source maps for `object` and `string` here.

```yaml
element: "member"
content:
  key:
    element: "string"
    attributes:
      sourceMap:
        -
          element: "sourceMap"
          content:
            -
              - 24
              - 11
    content: "a"
  value:
    element: "object"
    content:
      -
        element: "member"
        content:
          key:
            element: "string"
            attributes:
              sourceMap:
                -
                  element: "sourceMap"
                  content:
                    -
                      - 41
                      - 12
            content: "m1"
          value:
            element: "string"
```

is the result. But this result is wrong for the first case where `object` is mentioned because the `object` element should have source map.

So, yes, I still stand for my original point.

klokane:
Ok, let's continue, What about this case?

```
## Data Structures
### A (object)
### B
### C (A)
### D (B)
```

pksunkara:
Same as I am saying above. First `object` element should contain source map while the second one shouldn't. `A` and `B` references should contain source maps.

pksunkara:
TODO: We need to add the inferred link to the object element.

Issue #100, opened 2015-09-09
MSON - serialize "samples" and "default" attributes in refract form instead simple value


klokane:
```mson
- a: 3,4 (array, sample)
```

is now serialized with "samples" in non-refracted form
```json
"samples" : [["3","4"]]
```

It should be serialized in same form as content:
```json
"samples": [
  [{
    "element": "string",
     "content": "3"
  },
  {
    "element": "string",
     "content": "4"
  }],
]
```

pksunkara:
Yes. This should be fixed before release.

pksunkara:
Fixed by #104

PR #49, opened 2015-06-23
refract update samples

klokane:
Update samples/default attributes to follow changes in refract spec

PR #398, opened 2016-09-23
fix(API Elements): Don't wrap default values in an array

kylef:
As per the API Elements specification (http://api-elements.readthedocs.io/en/latest/element-definitions/#data-structure-base-api-element). The content of a the default value should not be an array.

> The type of of default attribute MUST match the type of element's content.

This is not a breaking change to our own uses:

- Attributes Kit supports the correct behaviour: https://github.com/apiaryio/attributes-kit/blob/87f5e500b9435d5660a6bd4254f747ffbb83eff3/src/Modules/SamplesPreprocessor/SamplesPreprocessor.js#L31-L43
- Metamorphoses **only** supports the correct behaviour. I am fixing this in Drafter because I realised it was broken when integrating with Metamorphoses: https://github.com/apiaryio/metamorphoses/blob/22b13a6a97aa5d29b711c6821715a88fe2714bda/src/adapters/refract/getUriParameters.coffee#L24

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -153,7 +153,7 @@ namespace drafter {
 
         // Add default value
         if (!parameter.node->defaultValue.empty()) {
-            element->attributes[SerializeKey::Default] = CreateArrayElement(LiteralToRefract<T>(MAKE_NODE_INFO(parameter, defaultValue), context), true);
+            element->attributes[SerializeKey::Default] = LiteralToRefract<T>(MAKE_NODE_INFO(parameter, defaultValue), context);
         }
 
         return element;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 * Correctly set the content type for custom JSON Schema.
   [#392](https://github.com/apiaryio/drafter/issues/392)
 
+* Default values were incorrectly wrapped in an array inside dataStructures for
+  API Elements output in Drafter. As per the API Elements specification this
+  isn't the correct behaviour and as such this no longer is the case.
+
 
 ## 3.1.1
 
--- a/features/fixtures/refract.json
+++ b/features/fixtures/refract.json
@@ -85,12 +85,7 @@
                                 }
                               ]
                             ],
-                            "default": [
-                              {
-                                "element": "string",
-                                "content": "<default value>"
-                              }
-                            ]
+                            "default": "<default value>"
                           },
                           "content": [
                             {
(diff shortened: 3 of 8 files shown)


pksunkara:
Wait, It is the correct behaviour. Don't you recall we had a long discussion on this during the development? @klokane will definitely know more.

The type of `default` should be the type of `content`.
The type of `samples` should be an array of type of `content`.

klokane:
This change will not affect other test fixtures? Just "integration" tests? It is strange

kylef:
> This change will not affect other test fixtures? Just "integration" tests? It is strange

Nope, it also effects the test for resource parameters: `test/fixtures/api/resource-parameters.apib`.

klokane:
As far as I remember from discussion, value of default/samples should be able to replace "content" directly without changes.

e.g. https://github.com/apiaryio/drafter/blob/master/test/fixtures/mson/primitives.apib#L13
https://github.com/apiaryio/drafter/blob/master/test/fixtures/mson/primitives.ast.json#L99

pksunkara:
Default should be able to replace. Samples should be an array of it.

klokane:
@pksunkara sure

klokane:
From point of view "replaceability" seem original behavior correct

klokane:
I mean, it should be possible to find original discussion related to this (at least part of it)

klokane:
Found PR, but there is no related discussion https://github.com/apiaryio/drafter/pull/49

pksunkara:
Commit message in https://github.com/apiaryio/drafter/pull/49/commits/43f554440e6acf4f35e86d7420a3e1e2972b0252

klokane:
And something related is there  https://github.com/apiaryio/drafter/issues/100

src/RefractAPI.cc
```diff
@@ -153,7 +153,7 @@ namespace drafter {
 
         // Add default value
         if (!parameter.node->defaultValue.empty()) {
-            element->attributes[SerializeKey::Default] = CreateArrayElement(LiteralToRefract<T>(MAKE_NODE_INFO(parameter, defaultValue), context), true);
+            element->attributes[SerializeKey::Default] = LiteralToRefract<T>(MAKE_NODE_INFO(parameter, defaultValue), context);
```
pksunkara: Since this is run only for `enum`, default should always be an array. Thus, this change is wrong

kylef: Okay I understand.

IMO it would have been clearer if API Elements didn't say that enum's default should be an array. There can only ever be one default and it doesn't make sense that it should be wrapped in an array.

w-vi: Can we change the spec to make it more sensible?

PR #399, opened 2016-09-23
fix(API Elements): Don't double wrap samples in array

kylef:
I don't believe that these sample values should be double wrapped in arrays and I think this is a mistake.

API Elements (http://api-elements.readthedocs.io/en/latest/element-definitions/#properties_13) specifies that:

> The type of items in samples array attribute MUST match the type of element's content.

### DIFF
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -147,7 +147,7 @@ namespace drafter {
         // Add sample value
         if (!parameter.node->exampleValue.empty()) {
             refract::ArrayElement* samples = new refract::ArrayElement;
-            samples->push_back(CreateArrayElement(LiteralToRefract<T>(MAKE_NODE_INFO(parameter, exampleValue), context), true));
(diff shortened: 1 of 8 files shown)


pksunkara:
From my comment on other PR:

> Wait, It is the correct behaviour. Don't you recall we had a long discussion on this during the development? @klokane will definitely know more.

> The type of `default` should be the type of `content`.
> The type of `samples` should be an array of type of `content`.

src/RefractAPI.cc
```diff
@@ -147,7 +147,7 @@ namespace drafter {
         // Add sample value
         if (!parameter.node->exampleValue.empty()) {
             refract::ArrayElement* samples = new refract::ArrayElement;
-            samples->push_back(CreateArrayElement(LiteralToRefract<T>(MAKE_NODE_INFO(parameter, exampleValue), context), true));
+            samples->push_back(LiteralToRefract<T>(MAKE_NODE_INFO(parameter, exampleValue), context));
```
pksunkara: Since this code is run in `enum`, the samples should always be an array of `content`. Thus, this change is wrong.

Issue #401, opened 2016-09-28
Error when `member` as a named type


pksunkara:
That said, the following API Blueprint will raise an unintuitive error message, I would treat this as a bug in [Drafter](https://github.com/apiaryio/drafter) the underlying API Blueprint parser. We should improve the error message to make it clear to users that they are using reserved keywords incorrectly in cases we can detect.

```apib
## Data Structures
### member (object)
```

```
error: (4)  You can not register a basic element
```

However, if I use `member` as titlecase I don't get an error at all:

```apib
## Data Structures
### Member (object)
```

w-vi:
Can we consider this as fixed by #405 ? The error message has been improved.

pksunkara:
Yup.

PR #405, opened 2016-10-04
fix: Improve error message when using keywords

w-vi:
Partially solves #401. Allowing reserved keywords altogether has some consequences which need to be solved separately.

### DIFF
--- a/src/NamedTypesRegistry.cc
+++ b/src/NamedTypesRegistry.cc
@@ -360,7 +360,13 @@ namespace drafter {
             refract::IElement* element = factory.Create(std::string(), eValue);
             element->meta["id"] = refract::IElement::Create(name);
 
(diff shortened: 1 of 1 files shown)


src/NamedTypesRegistry.cc
```diff
@@ -381,7 +387,13 @@ namespace drafter {
                 context.GetNamedTypesRegistry().remove(name);
                 delete pre;
 
-                context.GetNamedTypesRegistry().add(element);
+                try {
+                    context.GetNamedTypesRegistry().add(element);
+                } catch (refract::LogicError& e) {
+                std::ostringstream out;
```
pksunkara: Bad indentations.

pksunkara:
What kind of consequences?

PR #412, opened 2016-10-11
JSON Schema generation for fixed-type variable property with multiple items

kylef:
Extends from https://github.com/apiaryio/drafter/pull/411 to handle the case where there is an fixed-type array of multiple items.

### DIFF
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -232,7 +232,8 @@ namespace refract
 
             StringElement *str = TypeQueryVisitor::as<StringElement>((*i)->value.first);
             if (str) {
-                JSONSchemaVisitor renderer(pDefs, fixed);
+                bool fixedType = IsTypeAttribute(*(*i), "fixedType");
+                JSONSchemaVisitor renderer(pDefs, fixed, fixedType);
                 Visit(renderer, *(*i)->value.second);
                 pDefs->push_back(new MemberElement(
                                      str->value,
--- a/test/fixtures/schema/variable-property-fixed-type2.apib
+++ b/test/fixtures/schema/variable-property-fixed-type2.apib
@@ -1,0 +1,8 @@
+# GET /
+
++ Response 200 (application/json)
+    + Attributes
+        + *y* (array, fixed-type)
+            + (object)
+                + name (required)
+        + test
--- a/test/fixtures/schema/variable-property-fixed-type2.json
+++ b/test/fixtures/schema/variable-property-fixed-type2.json
@@ -1,0 +1,174 @@
+{
+  "element": "parseResult",
+  "content": [
+    {
+      "element": "category",
+      "meta": {
+        "classes": [
+          "api"
+        ],
+        "title": ""
+      },
+      "content": [
+        {
+          "element": "category",
+          "meta": {
+            "classes": [
+              "resourceGroup"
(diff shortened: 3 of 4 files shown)


test/fixtures/schema/variable-property-fixed-type.json
```diff
@@ -144,7 +210,7 @@
...
+                              "content": "{\n  \"$schema\": \"http://json-schema.org/draft-04/schema#\",\n  \"type\": \"object\",\n  \"allOf\": [\n    {\n      \"$ref\": \"#/definitions/y\"\n    },\n    {\n      \"$ref\": \"#/definitions/z\"\n    }\n  ],\n  \"definitions\": {\n    \"y\": {\n      \"type\": \"array\",\n      \"items\": {\n        \"type\": \"object\",\n        \"properties\": {\n          \"name\": {\n            \"type\": \"string\"\n          }\n        },\n        \"required\": [\n          \"name\"\n        ]\n      }\n    },\n    \"z\": {\n      \"type\": \"array\",\n      \"items\": [\n        {\n          \"type\": \"object\",\n          \"properties\": {\n            \"first\": {\n              \"type\": \"string\"\n            }\n          },\n          \"required\": [\n            \"first\"\n          ]\n        },\n        {\n          \"type\": \"object\",\n          \"properties\": {\n            \"last\": {\n              \"type\": \"string\"\n            }\n          },\n          \"required\": [\n            \"last\"\n          ]\n        }\n      ]\n    }\n  }\n}"
```
kylef: For PR readability. Here is the prettified JSON Schema:

```json
{
  "type": "object",
  "allOf": [
    {
      "$ref": "#/definitions/y"
    },
    {
      "$ref": "#/definitions/z"
    }
  ],
  "definitions": {
    "y": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          }
        },
        "required": [
          "name"
        ]
      }
    },
    "z": {
      "type": "array",
      "items": [
        {
          "type": "object",
          "properties": {
            "first": {
              "type": "string"
            }
          },
          "required": [
            "first"
          ]
        },
        {
          "type": "object",
          "properties": {
            "last": {
              "type": "string"
            }
          },
          "required": [
            "last"
          ]
        }
      ]
    }
  }
}
```

pksunkara: Wait, why is `items` in `z` is an array?

pksunkara: No, the test is having both `y` and `z`.

kylef: Ah good point. In that case it should use a separate test file otherwise it will drop test coverage for https://github.com/apiaryio/drafter/pull/411.

I've updated PR to take this into account.

w-vi: I am bit confused, where is the `z` coming from?

kylef: @w-vi `z` doesn't exist in the current test fixture. This thread is from previous commits.

pksunkara: Stupid github.

w-vi: OK, sorry ... this new reviewing system sucks in this regard

Issue #413, opened 2016-10-11
JSON Schema in `variable-property-name` test case is wrong


pksunkara:
## Blueprint

``` apib
# GET /
+ Response 200 (application/json)
    + Attributes (A)

# Data Structures

## StrSample (string)
### Sample
str sample

## StrDefault (string)
### Default
str default

## StrEmpty (string)

## StrObj (object)


# A
- s: start

- *direct*: v1
- *str (string)*: v2

- *(StrSample)*: s1
- *ss (StrSample)*: s2

- *(StrDefault)*: d1
- *sd (StrDefault)*: d2

- *(StrEmpty)*: e1
- *se (StrEmpty)*: e2

- e: end
```
## Current Schema

``` json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "allOf": [
    {
      "$ref": "#/definitions/direct"
    },
    {
      "$ref": "#/definitions/str"
    },
    {
      "properties": {
        "s": {
          "type": "string"
        },
        "ss": {
          "type": "string"
        },
        "sd": {
          "type": "string"
        },
        "": {
          "type": "string"
        },
        "se": {
          "type": "string"
        },
        "e": {
          "type": "string"
        }
      }
    }
  ],
  "definitions": {
    "direct": {
      "type": "object",
      "patternProperties": {
        "": {
          "type": "string"
        }
      }
    },
    "str": {
      "type": "object",
      "patternProperties": {
        "": {
          "type": "string"
        }
      }
    }
  }
}
```
## Expected

The `ss`, `sd` and `se` variable properties must also be definitions like `direct` and `str`. The `""` should not exist at all.

kylef:
What is wrong about it? can you please elaborate.

pksunkara:
Yeah, was updating the description.

tjanc:
@pksunkara master currently produces the bellow schema. Does this fix the issue?

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "s": {
      "type": "string"
    },
    "e": {
      "type": "string"
    }
  },
  "patternProperties": {
    "(?:)": {
      "type": "string"
    }
  }
}
```

PR #414, opened 2016-10-11
fix: JSON Schema for variable properties

### LINKED ISSUE #367: Variable Property Name doesn't seem to be correctly implemented
If I try to parse

```apib
FORMAT: 1A

# Beehive API

## Honey [/honey]

### Retrieve [GET]

    + Attributes

        + todosById
            + *8*
                + text: Prepare Dinner (string, required)
```

with `drafter@3.0.0`, I get

#### JSON

```json
{
  "todosById": {
    "8": {
      "text": "Prepare Dinner"
    }
  }
}
```

#### JSON Schema

```json
{
  "type": "object",
  "properties": {
    "todosById": {
      "$ref": "#/definitions/8"
    }
  },
  "definitions": {
    "8": {
      "type": "object",
      "properties": {
        "text": {
          "type": "string"
        }
      },
      "required": [
        "text"
      ]
    }
  }
}
```

I think the schema is incorrect, because the property has [Variable Property Name](https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#322-variable-property-name), but the schema looks like it forces you to have a property called `8` exactly. In my opinion this is a bug.

--------------------

This issue was filed as an outcome of investigation of https://github.com/apiaryio/dredd/issues/534

pksunkara:
Fixes #367. But I discovered a new bug #413. No, I don't think it should be fixed in this PR because it is completely different.


pksunkara:
I have rebased this on master now.

src/refract/JSONSchemaVisitor.cc
```diff
@@ -222,23 +222,39 @@ namespace refract
         }
     }
 
+    ObjectElement* JSONSchemaVisitor::definitionFromVariableProperty(JSONSchemaVisitor& renderer)
+    {
+        ObjectElement* definition = new ObjectElement();
+
+        definition->push_back(new MemberElement("type", new StringElement("object", IElement::rCompact), IElement::rCompact));
```
w-vi: Can these super long lines be somehow indented?

pksunkara: I don't think it's that long. It's only 126,

w-vi: And I do ... so this brings again the bloody style issue, what is our column limit? Is it 80, 120, 280? According to the clang format it is 120, https://github.com/apiaryio/snowcrash/pull/380/files#diff-69a68f7ce1ea3687a38d8e4705e6fa48R25

pksunkara: I can fix this later once you finish the review.

w-vi:
Fixture https://github.com/apiaryio/drafter/blob/master/test/fixtures/schema/variable-property2.apib

produces following schema 
```
{
  "type": "object",
  "properties": {
    "item": {
      "allOf": [
        {
          "$ref": "#/definitions/link"
        },
        {
          "properties": {
            "id": {
              "type": "object",
              "properties": {
                "name": {
                  "type": "string"
                },
                "email": {
                  "type": "string"
                }
              }
            }
          }
        }
      ]
    }
  },
  "definitions": {
    "link": {
      "type": "object",
      "patternProperties": {
        "": {
          "type": "object",
          "properties": {
            "href": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```

but I am not sure if this is correct as following JSON seems to validate against it event though iot seems invalid to me. Maybe we have some fundamental flaw in our design.

```
{
 "l": {
      "h": "a URI"
    },
    "i": {
      "n": "",
      "e": ""
    }
}
```

pksunkara:
Maybe the validator is wrong? 😄. I will look into it.

pksunkara:
Actually, we do not do any required properties here and we are not restricting random properties. I guess @honzajavorek can test this for us in dredd since this issue came from him originally.

w-vi:
You are right, otherwise it LGTM.

pksunkara:
@w-vi Will you be merging this?

w-vi:
I have another example which I am not sure what the result should be.

Take the https://github.com/apiaryio/drafter/blob/master/test/fixtures/schema/variable-property.apib and make the `*self*` required.  The schema is definitely not correct. 

```

{
  "type": "object",
  "properties": {
    "_links": {
      "$ref": "#/definitions/self",
      "required": [
        "self"
      ]
    }
  },
  "definitions": {
    "self": {
      "type": "object",
      "patternProperties": {
        "": {
          "type": "object",
          "properties": {
            "href": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```

pksunkara:
Well, the schema is correct. MSON means that a variable property should be present. And JSON Schema uses `self` as a sample. So, It's kinda correct.

The issue is conceptually, it doesn't mean much in MSON. Since both required and variable property are targetted at difference scenarios.

Maybe, later we can add a warning saying that variable property can't be required or optional.

w-vi:
@pksunkara please rebase so it can be merged.

I think we are still not handlingg  the schema correctly, but at least now we let slip something rather then restricting it.

w-vi:
Oh and please indent the long lines.

pksunkara:
Done and done.

PR #416, opened 2016-10-13
Update snowcrash

kylef:
See the CHANGELOG for the changes in the diff.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@
   Schemas.
   [#390](https://github.com/apiaryio/drafter/issues/390)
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
   Schemas.
   [#390](https://github.com/apiaryio/drafter/issues/390)
 
+* Drafter will now check the validity of header names and provide a warning
```
w-vi: It will also skip those headers with invalid chars. Don't know if that is needed to mention or not but previously it issued warning too but left the header intact.

PR #418, opened 2016-10-17
chore: Maintenance of C API

pksunkara:
This is to allow the protagonist to wrap the sos object in v8.
Otherwise, it will have to JSON parse the returned string by
drafter which will be a performance issue.

### DIFF
--- a/test/test-CAPI.c
+++ b/test/test-CAPI.c
@@ -26,7 +26,7 @@ int test_parse_and_serialize() {
     char* out = drafter_serialize(result, options);
     assert(out);
 
-    int len = strlen(expected);
+    size_t len = strlen(expected);
 
     assert(strncmp(out, expected, len) == 0);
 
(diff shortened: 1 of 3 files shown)


pksunkara:
@w-vi This is not ready to be reviewed. /cc @kylef

pksunkara:
Sorry, **NOW**, not **not**.

test/test-CAPI.c
```diff
@@ -26,7 +26,7 @@ int test_parse_and_serialize() {
     char* out = drafter_serialize(result, options);
     assert(out);
 
-    int len = strlen(expected);
+    int len = (int) strlen(expected);
```
klokane: some reason to do not use `size_t`?

w-vi: beacue the `len` is `int` it shoud actually be `size_t len = strlen(expected)`

w-vi: https://linux.die.net/man/3/strlen

PR #421, opened 2016-10-20
feat(api): Added some temporary private C api functions

pksunkara:


src/drafter_private.h
```diff
@@ -0,0 +1,34 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct {
+    bool requireBlueprintName;
+} drafter_parse_options;
+
+DRAFTER_API int drafter_parse_blueprint_with_options(const char* source,
+                                                     drafter_result** out,
```
kylef: I think we should put the options before the output that seems to make more logical sense to me. What do you think?

w-vi:
CI failing, otherwise LGTM.
I suppose that once we do major release we will join the options under the  `drafter_options` right?

pksunkara:
> I suppose that once we do major release we will join the options under the drafter_options right?

Though we can try to see if `drafter_parse_options` and `drafter_serialize_options` should be separate.

pksunkara:
Passing now.

src/drafter.cc
```diff
@@ -125,38 +108,10 @@ DRAFTER_API char* drafter_serialize(drafter_result *res, const drafter_options o
...
-        std::transform(filter.elements().begin(), filter.elements().end(),
-                       std::back_inserter(elements),
-                       std::bind(&refract::IElement::clone, std::placeholders::_1, refract::IElement::cAll));
-
-        out = new refract::ArrayElement(elements);
-        out->element(drafter::SerializeKey::ParseResult);
-    }
-
-    drafter_free_result(result);
+    drafter_parse_options options;
```
w-vi: `drafter_parse_options options = {false};`

pksunkara:
Regarding the options,

`drafter_parse` function would need only parse options
`drafter_serialize` function would need only serialise options
`drafter_parse_to` function would need both.

So, we have to find a way to do this smartly.

PR #425, opened 2016-11-18
fix(schema): Add property defaults to schema

### LINKED ISSUE #424: Output "default" MSON as "default" JSON Schema
Hi there,

Currently MSON defaults are only emitted as `application/json` content and not as `application/schema+json` which reminds me of #392.

Related to our use apiaryio/mson#70 use case I would like to be able to export MSON Defaults as described [here](https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#45-default), for example:

```
 - deleted: false (boolean, required, default)
```

as JSON Schema defaults that are part of the [JSON Schema Validation spec](http://json-schema.org/latest/json-schema-validation.html#anchor101) in this way:

```
    "deleted": {
      "type": "boolean",
      "default": false
    }
```

Which will enable us to deal with the [*Boolean Gotcha* for the form generator](https://github.com/mozilla-services/react-jsonschema-form/issues/48#issuecomment-199671252) we use and avoid empty required boolean checkboxes to be seen as `undefined` and be seen as `false` instead.

image

kylef:
Fixes #424

### DIFF
--- a/src/refract/JSONSchemaVisitor.cc
+++ b/src/refract/JSONSchemaVisitor.cc
@@ -194,6 +194,7 @@ namespace refract
         if (ext) {
             IElement* merged = ext->merge();
             str = TypeQueryVisitor::as<StringElement>(merged);
+
             if (!str) {
                 delete merged;
             }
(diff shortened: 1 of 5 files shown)


src/refract/JSONSchemaVisitor.cc
```diff
@@ -211,6 +211,15 @@ namespace refract
             if (IsTypeAttribute(e, "nullable")) {
                 renderer.addSchemaType("null");
             }
+
+            if (e.value.second) {
+                IElement::MemberElementCollection::const_iterator defaultIt = e.value.second->attributes.find("default");
+
+                if (defaultIt != e.value.second->attributes.end()) {
+                    renderer.addMember("default", (*defaultIt)->clone());
```
kylef: This works for primitive types, however for other types such as enum, array and object this falls down (see test examples).

The default value for primitive types are StringElement, BooleanElement, NumberElement which renders correctly. But in types where the default is not, it will be an ArrayElement etc. Where it will render as an object element.

Is there any easy way to solve this problem without creating some other visitor that will handle rendering the default element?

w-vi:
Just a quick note I haven't gone through your changes we have `GetDefault` template function for retrieving defaults but I don't see it used, any reason?
https://github.com/apiaryio/drafter/blob/57c4067e85cb51f41f8d330daa36709ec1768740/src/refract/VisitorUtils.h#L63

kylef:
> Just a quick note I haven't gone through your changes we have GetDefault template function for retrieving defaults but I don't see it used, any reason?

I couldn't make use of the `GetDefault` templated function, since it is templated and I would first need to know the source and destination element types. In this case it can be variable.

pksunkara:
This PR messes up the schema for the following test case:

```
# GET /
    + Attributes
        + colors (array)
            + Default: red
```

It somehow serialises the refract element directly into JSON Schema

w-vi:
Isn't it because it is not a primitve type

pksunkara:
Yeah, I just wanted to document the main issue here.

PR #430, opened 2016-11-24
Fix url protocol of submodule /test/vendor/catch

dewe:
git clone --recursive did not work, caused by a submodule reference using git:// instead of https://.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/apiaryio/snowcrash.git
 [submodule "test/vendor/Catch"]
 	path = test/vendor/Catch
-	url = git://github.com/philsquared/Catch.git
+	url = https://github.com/philsquared/Catch.git
 [submodule "ext/sos"]
 	path = ext/sos
 	url = https://github.com/apiaryio/sos.git


dewe:
Forgot to mention that this happened due to company network restrictions.

dewe:
Fixed, and merged from upstream master

pksunkara:
Can you please remove the merge commit and leave the only first commit? It's not a good git practice.

kylef:
We will also need to update the dependencies of snowcrash (https://github.com/apiaryio/snowcrash/blob/master/.gitmodules#L1) otherwise a recursive clone would still fail in snowcrash (dependency of drafter).

Probably a good idea to do that before merging this PR so this PR can update the reference to Snowcrash.

dewe:
Ok, fixed.

dewe:
@kylef Oh, I thought the PR was trivial. If that's not the case, perhaps we just should ditch it and focus on better things. I can work around the initial networking issues on my side.

kylef:
@dewe I think it's still trivial, just that snowcrash needs to be updated too.

pksunkara:
Please update snowcrash submodule to apiaryio/snowcrash#417 and we can merge this.

PR #432, opened 2016-11-28
chore: Release 3.2.0

pksunkara:


w-vi:
why is it a minor and not just a patch?

pksunkara:
Because the source maps rework changes the output a bit for code blocks.

PR #433, opened 2016-11-29
chore: update snowcrash version (fix crash - undislosed item list)

klokane:


test/fixtures/syntax/undisclosed-listitem.json
```diff
@@ -0,0 +1,59 @@
...
+                  "element": "dataStructure",
+                  "content": [
+                    {
+                      "element": "object",
+                      "content": [
+                        {
+                          "element": "member",
+                          "content": {
+                            "key": {
+                              "element": "string"
```
kylef: A key without content. Should we perhaps issue a warning to the user that they didn't supply a key for an MSON property?

klokane: it is original behavior (before sundown changes), but I can introduce fix for it if you like

kylef: No need, just a though.

pksunkara: I am not sure, but IIRC we had a discussion on that and decided to leave it for now.

PR #444, opened 2017-01-09
fix (C api): cleanup and improve the v4 C API

w-vi:


src/drafter.cc
```diff
@@ -27,27 +27,36 @@
...
-        return -1;
+    if (!source) {
+        return DRAFTER_EINVALID_INPUT;
     }
 
+    if (!out) {
+        return DRAFTER_EINVALID_OUTPUT;
+    }
+
+
```
pksunkara: Unnecessary extra line

src/drafter.cc
```diff
@@ -110,18 +123,36 @@ namespace { // FIXME: cut'n'paste from main.cc - duplicity
...
+    }
+
+    drafter::WrapperOptions wrapperOptions(serialize_opts.sourcemap);
     drafter::ConversionContext context(wrapperOptions);
 
     sos::Object result = drafter::SerializeRefract(res, context);
 
-    std::unique_ptr<sos::Serialize> serializer(CreateSerializer(options.format == DRAFTER_SERIALIZE_JSON ? drafter::JSONFormat : drafter::YAMLFormat));
+
+
```
pksunkara: Unnecessary extra lines

src/drafter.cc
```diff
@@ -110,18 +123,36 @@ namespace { // FIXME: cut'n'paste from main.cc - duplicity
...
+        case DRAFTER_SERIALIZE_YAML:
+            format = drafter::YAMLFormat;
+            break;
+
+        default:
+            return nullptr;
+    }
+
+    drafter::WrapperOptions wrapperOptions(serialize_opts.sourcemap);
```
pksunkara: I thought we were removing the wrapper options in this new version.

src/drafter.cc
```diff
@@ -27,27 +27,36 @@
 
 #include <string.h>
 
+
+#define DRAFTER_EUNKNOWN -1
+#define DRAFTER_EINVALID_INPUT -2
+#define DRAFTER_EINVALID_OUTPUT -3
```
klokane: What about header file? it probably should be public info

src/drafter.cc
w-vi: I don't understand.

pksunkara: This wrapper options is temporary hack we introduced. We decided to remove it along with blueprint AST. Didn't we?

w-vi: This PR is not about removing AST stuff but about API cleanup. I'd rather do the removal in separate PR.

pksunkara: I am okay with that. I am just trying to think if removing wrapperOptions later will mess things up or not. Can you take a look over it and make sure we are okay with not removing it now?

w-vi: It is not publicly visible, so we can change it any time as long as we keep the parameters and function declarations.

src/drafter.h
```diff
@@ -62,43 +63,73 @@ typedef refract::IElement drafter_result;
...
     bool sourcemap;
     drafter_format format;
-} drafter_options;
+} drafter_serialize_options;
+
 
-/* Parse API Blueprint and serialize it to given format.*/
-// TODO: Add parse options to this
+/* Error codes */
+#define DRAFTER_EUNKNOWN -1
```
kylef: What about putting these in an enumeration?

```c
enum drafter_result_t: int {
    DRAFTER_OK -1
    DRAFTER_EUNKNOWN = -1
    DRAFTER_EINVALID_INPUT = -2
    DRAFTER_EINVALID_OUTPUT = -3
}
```

That away they are grouped together and allow proper type safety in the result.

w-vi: Is this a valid ANSI C enum declaration? I don't remember.

kylef: You're right, but does it need to be declared as an int? I think there are more benefits in the type-safety you'd gain so it would be more worthwhile than using int.

enum example in correct syntax for C:

```c
enum drafter_result_t {
    DRAFTER_OK = -0,
    DRAFTER_EUNKNOWN = -1,
    DRAFTER_EINVALID_INPUT = -2,
    DRAFTER_EINVALID_OUTPUT = -3,
};
```

w-vi: The down side is that when using FFI you need to recreate it but with int you just check if it is 0 or not. I am ok with both styles I just picked the int approach as the easier one to work with.

w-vi: Actually we are returning positive int codes from snowcrash so the `enum` is not complete ... that's why I picked just plain `int`.

klokane: +1 for `#define`
it is common practice for C APIs

w-vi: Merging it as is ... possible changes before release are allowed aren't they.

kylef: Yes, if not enum I'd suggest they should be actual constants with symbols so they will be available at run-time for FFI.

I'll make a follow up PR shortly to open that discussion.

PR #446, opened 2017-01-15
Update snowcrash

kylef:
This PR is against the `stable/3.x` branch, once we've made a release of it we should merge the branch into master so the changes and changelog are merged. See changelog for contents of change.

PR #447, opened 2017-01-16
Serialise parameters in API Elements as string like API Blueprint AST

kylef:
These changes will change how parameter values and types are expose in API Elements to align them with API Blueprint AST.

Previously we would treat parameters values as Refract elements directly which would cause the following problems:

- The underlying type can be silently lossed (i.e, `datetime` type will be shown as string in other tools).
- The value would be interpreted as boolean or number and subsequently serialised in a different way. (Boolean being true or false instead of the value a user specified). These changes keep the original string in tact because there are no specified rules on how types should be serialised in a URI and not doing so would be a breaking change.

This change would make all parameter content to be "string" and we would add additional attributed `typeLabel` (other name suggestions welcome). These will not be breaking changes because any existing tool will already understand string type and do the right thing. Use of the label may need to be added, but this is backwards-compatible.

I think we should go ahead with these changes at the moment to keep the current behaviour working, especially since the API Blueprint Specification mentions no rules on limitations of how formats will be serialised. **In a future where we would introduce MSON Parameters we should consider how types will be serialised and make a decision on how to store them in API Elements**. We shouldn't however bundle these breaking changes with switching from API Blueprint AST to API Elements.

I will rebase these changes once https://github.com/apiaryio/drafter/pull/446 is merged and add a CHANGELOG entry.


src/RefractAPI.cc
```diff
@@ -187,26 +187,18 @@ namespace drafter {
...
-            value = ExtractParameter<bool>(parameter, context);
-        }
-        else if (parameter.node->type == "number") {
-            value = ExtractParameter<double>(parameter, context);
-        }
-        else {
-            value = ExtractParameter<std::string>(parameter, context);
-        }
-
+        refract::IElement *value = ExtractParameter<std::string>(parameter, context);
```
klokane: It is no more required to be `template`.
Please remove template behavior `ExtractParameter()` - and use strictly `std::string` inside of function

kylef: I've addressed this in https://github.com/apiaryio/drafter/pull/447/commits/c65755cd0a3bfa846cf47fff0b8413c949b4e353.

klokane:
approved

w-vi:
:+1:

pksunkara:
Please don't merge this until I am updated about any offline discussions you guys had about this.

kylef:
I've updated this PR to use `type` instead of `typeLabel` and filed API Elements PR at https://github.com/apiaryio/api-elements/pull/20

pksunkara:
We still need to discuss about the second part of the PR. I have been working on some alternatives for now.

kylef:
@pksunkara what is the second part of the PR you would like to discuss?

pksunkara:
`Serialize parameter values as string in API Elements`. We decided to implement it another way, but you did it like this. Do you remember?

kylef:
> We decided to implement it another way, but you did it like this. Do you remember?

How did we "decide" to implement it? I think string is the only way that makes sense. I talked through this in standup with @w-vi and I've jotted down everything from the discussion at https://github.com/apiaryio/api-elements/pull/20 please read and respond with any reasons against it there please.

pksunkara:
The issue of "The value would be interpreted as boolean or number and subsequently serialised in a different way." can be separated from the "type" issue and improved in a different way which we discussed before I left.

I am okay with the fix for "type" issue but not the boolean value thing.

kylef:
@pksunkara Could you please offer a counter proposal of how you are suggested to solve this problem.

pksunkara:
Let's say a user has a boolean parameter for which he's using `0` and `1`. The solution we came up with was to allow drafter to read it as boolean values, but serialise them into `true` or `false`.

kylef:
> Let's say a user has a boolean parameter for which he's using `0` and `1`. The solution we came up with was to allow drafter to read it as boolean values, but serialise them into `true` or `false`.

In that case, how would subsequent tooling understand that the boolean values true/false should be serialised as `0` and `1`?

pksunkara:
> should be serialised as 0 and 1?

I am tempted to say that's not our concern and we need to stay true to the concept of API Elements. But I guess, we can add something like the `type` here.

Issue #469, opened 2017-05-11
No schema for elements of fixed-type array


goganchic:
I have such apib:

```apib
FORMAT: 1A

# Project

# Data Structures

## VehicleReportRow
+ vehicle: `123` (string, required)

# Group G

## Reports [/reports]

### Get Report [GET]

+ Response 200 (application/json)
    + Attributes
        + result (array[VehicleReportRow], required, fixed-type)
```

Drafter produces such schema

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "result": {
      "type": "array"
    }
  }
}
```

But if I change name of VehicleReportRow to something else (VehicleReportRow1, VehiclesReportRow, Test, etc) - drafter produces correct schema:

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "result": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "vehicle": {
            "type": "string"
          }
        },
        "required": [
          "vehicle"
        ]
      }
    }
  },
  "required": [
    "result"
  ]
}
```

Why?

goganchic:
even in such case everything is fine: ` + result (array[VehicleReportRow] , required, fixed-type)` (I've added space before comma)

goganchic:
It was an error in attributes parsing. I've prepared a fix.

pksunkara:
Fixed by #465. Will be released soon.

goganchic:
@pksunkara can you please tell me when this fix will be in master branch?

pksunkara:
Master is for version 4 which will take a few weeks. We will be releasing a 3.x.x version with the stable branch soon.

goganchic:
ok, thanks!

PR #470, opened 2017-05-11
Fix attributes parsing

goganchic:
fix https://github.com/apiaryio/drafter/issues/469


pksunkara:
Please use the stable branch instead of master.

pksunkara:
And this has been fixed in stable already.

goganchic:
and what's about master?

PR #475, opened 2017-05-29
Refract v1: Ref Element and Metadata

pksunkara:


klokane:
I know, there is it from historical reason. But there is question if there is no good time to introduce RefElement

pksunkara:
> I know, there is it from historical reason. But there is question if there is no good time to introduce RefElement

I will be doing this in the next PR. Please merge it in the meanwhile.

PR #476, opened 2017-05-31
feat: Added Ref Element

pksunkara:


src/refract/RenderJSONVisitor.cc
```diff
@@ -61,8 +61,8 @@ namespace refract
                     continue;
                 }
 
-                if ((*it)->element() == "ref") {
-                    HandleRefWhenFetchingMembers<T>(*it, members, FetchMembers<T>);
+                if (RefElement* ref = TypeQueryVisitor::as<RefElement>(*it)) {
```
pksunkara: This line is going into for loop.

klokane: Probably because 

https://github.com/apiaryio/drafter/pull/476/files#diff-b5d89872d6f93c63ff1667145fb3d8a9L240

You can not remove Expanding Reference

pksunkara: I don't think that's related according to the debugging I did.

pksunkara:
This is now ready.

PR #478, opened 2017-06-06
fix: Data Structures should be directly serialized

pksunkara:


w-vi:
What's direct element? I don't remember seeing in the spec anywhere, can you please provide me a link.

pksunkara:
@w-vi Basically, a `refract::Element` can contain an `refract::Element` in content. Since we are not using it anywhere in drafter until now, we haven't implemented it.

klokane:
@w-vi we discuss it with @pksunkara. 

There were ability (by spec, not impl) to hold object in ObjectElement. But this ability was removed here:
https://github.com/refractproject/refract-spec/commit/de356616d583999a4f37b05de746f780f687cd96#diff-4eef42b68375cd377e0657ef7ea34f6bL242

But there is in refract spec ability to hold `Element` in `Element` 
https://github.com/refractproject/refract-spec/blame/master/refract-spec.md#L65

I also do not like idea if DirectElement for this purpose. We need some wide discussion about it.

kylef:
> There were ability (by spec, not impl) to hold object in ObjectElement. But this ability was removed here:
https://github.com/refractproject/refract-spec/commit/de356616d583999a4f37b05de746f780f687cd96#diff-4eef42b68375cd377e0657ef7ea34f6bL242

To be clear, the ability to directly place `object` (not `ObjectElement`) was removed from Refract specification to prevent ambiguities.

w-vi:
DirectElement is basicaly an Element correct? Feels strange.

PR #480, opened 2017-06-28
Bump minimum supported compiler versions

kylef:
Since we are near to releasing Drafter 4, this gives us another opportunity to make breaking changes in terms of supported compilers. I would propose we move to the following minimum versions:

| Compiler | Minimum Supported Version |
|----------|---------------------------|
| Clang    | 4.0 |
| GCC      | 5.3 |
| MSVC++   | 2015 |

This would allow us to make use of most [C++14 features and a few C++17 features](http://en.cppreference.com/w/cpp/compiler_support). I do not believe that supporting older compilers is as important as it once used to be, now that we have a pure JavaScript version of Drafter, the minimum compiler version doesn't add a barrier to using the API Blueprint parser and surrounding Node.JS tools such as Dredd as they can use the pure JavaScript version of Drafter.

The biggest requirement for the compilers would be supporting our development machines which are often the very latest compilers and also the platforms we use for parsing services. Here is the list of default compilers on various platforms:

| Platform | Compiler |
|--|--|
| Ubuntu 16.04 LTS | >= GCC 5.3.1 |
| CentOS 7 | GCC 4.8.5 |
| macOS (Xcode 8.3) | Clang 4 |
| Debian Stable (Stretch) | GCC 6 |
| ArchLinux | GCC 7.1.1 / Clang 4.0.1 |
| Gentoo | GCC 5.4 |
| Fedora 24-25 | GCC 6 |

I would like to have gone for GCC 6 to allow a lot of C++17, however I think Ubuntu 16.04 LTS support is important at the moment and therefore selected GCC 5.3 as the lowest supported version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Drafter Changelog
 
+## Master
+
+### Breaking
+
+* The following compiler versions are supported:
+
+  * Microsoft Visual C++ 2015 or higher
+  * GCC 5.3 or higher
+  * Clang 4.0 or higher
+
 ## 3.2.7
 
 ## Bug Fixes
--- a/README.md
+++ b/README.md
@@ -147,6 +147,15 @@ if (result) {
 ```
 
 ## Build
+
+### Compiler Support
+
+| Compiler | Minimum Supported Version |
+|----------|---------------------------|
+| Clang    | 4.0 |
+| GCC      | 5.3 |
+| MSVC++   | 2015 |
+
 1. Clone the repo + fetch the submodules:
 
     ```sh


kuba-kubula:
@kylef My five cents: Oracle Linux, RHEL 7 -- they're at gcc 4.8.5 (same as CentOS 7 is).
Do you really want to drop support for these server-like systems?

w-vi:
@kuba-kubula Not so many users are using these systems and it just means that it might not work out of box, extra package is needed. It does not mean that it would be impossible to install drafter or protagonist.

pksunkara:
@kylef Please rebase on master instead of merging master into PR branches. Makes commit graph better.

kylef:
@pksunkara Use "Rebase and Merge" and then the graph will be the same. Or better use "Squash and Merge" because this is single commit and not necessary to make it a merge commit.

pksunkara:
It is not enabled for our repos. I had to enable it, merge it and then disable it. Please use rebase in the future.

Issue #193, opened 2015-12-10
fixed attribute ignored when nested


w-vi:
AST is the same for 2 different Blueprints. the `fixed` attribute is somehow ignored along the way

*Blueprint1.apib*
```
# Resource [/example]

## Action [GET]

+ Response 200 (application/json)

    + Attributes
        - tags (array)
            - (string)
            - (number)
```

*Blueprint2.apib*
```
# Resource [/example]

## Action [GET]

+ Response 200 (application/json)

    + Attributes
        - tags (array)
            - (string, fixed)
            - (number, fixed)

```

for both the AST of the MSON looks like the same

```
  "content": [
    {
      "element": "dataStructure",
      "content": [
        {
          "element": "object",
          "content": [
            {
              "element": "member",
              "content": {
                "key": {
                  "element": "string",
                  "content": "tags"
                },
                "value": {
                  "element": "array",
                  "content": [
                    {
                      "element": "string"
                    },
                    {
                      "element": "number"
                    }
                  ]
                }
              }
            }
          ]
        }
      ]
    }
```

kylef:
Is this actually a bug? This looks as intended from what I understand. Since array itself doesn't have to contain the number or string fixed elements in MSON. The JSON Schema would be identical for these two data structures.

pksunkara:
JSON Schema is correct, but the actual refract is wrong. You don't see any `fixed` in `typeAttributes` in the refract tree.

PR #484, opened 2017-07-06
Enable array-restricted-to-type test fixture

kylef:
Since #193 was resolved, this fixture can be enabled (and updated to new serialisation).


kylef:
There was also a fixture (https://github.com/apiaryio/drafter/pull/484/commits/90ad4fe4e31fbf4b1e48fca3fcfe5949771d174d) completely missing from the tests so this was added.

PR #489, opened 2017-07-19
refactor: print visitor

tjanc:


src/refract/PrintVisitor.cc
```diff
@@ -6,180 +6,179 @@
...
-        }                                                                \
-        indentOS(indent);                                                \
-        os << "}\n";                                                     \
-    }                                                                    \
-
 namespace refract
 {
-    PrintVisitor::PrintVisitor()
-    : indent(0), os(std::cout) {}
+    PrintVisitor::PrintVisitor() : indent(0), os(std::cerr), ommitSrcMap(false)
```
pksunkara: Please rename `ommitSrcMap` to `ommitSourceMap`. We need to use the full name of the variable.

src/refract/PrintVisitor.cc
```diff
@@ -6,180 +6,179 @@
...
-        printAttr(e);
+        indented() << "+ " << e.element() << '\n';
+
+        PrintVisitor pv{indent + 1, os, ommitSrcMap};
+
+        refract::VisitBy(e, pv);
 
-        VisitBy(e, *this);
+        pv.printMeta(e);
+        pv.printAttr(e);
```
pksunkara: Please print the meta and attribute after element name and before the visiting of content.

src/refract/PrintVisitor.cc
```diff
@@ -6,180 +6,179 @@
...
+        out << "- Boolean ";
         if (v) {
-            os << *v;
-        } else {
-            os << "null";
+            out << *v;
         }
-        os << ")" << "\n";
+        out << "\n";
     }
```
pksunkara: Why not write both the above functions in similar style to the `StringElement` function?

src/refract/PrintVisitor.cc
```diff
@@ -6,180 +6,179 @@
...
 
     void PrintVisitor::operator()(const MemberElement& e)
     {
-        os << "MemberElement {\n";
-        indentOS(indent + 1);
+        StringElement* str = TypeQueryVisitor::as<StringElement>(e.value.first);
+        assert(str);
+
+        if (ommitSrcMap && (str->value.compare("sourceMap") == 0)) return;
```
pksunkara: I think this check should be in `printAttr` function.

src/refract/PrintVisitor.h
```diff
@@ -9,26 +9,37 @@
...
+
+       private:
+        std::ostream& indented();
 
-        void indentOS(int ind);
         void printMeta(const IElement& e);
         void printAttr(const IElement& e);
 
-    public:
+        template <typename ElementT>
```
pksunkara: `typename T` is good enough.

src/refract/PrintVisitor.cc
tjanc: Based on assumptions from the original implementation. I assumed both Boolean and Number differ from String in that nullptr actually is a valid value. In the original, String just expects the value to be dereferencable while for Boolean and Number do an explicit check.

tjanc: In that case I will just assert and print the value, as in my implementation of PrintVisitor::operator()(const StringElement&). Is that alright?

pksunkara:
Yes

src/refract/PrintVisitor.cc
tjanc: And, on a similar note, is it valid programmatically to have the first value of Member be of other type than String? If no, I'd suggest to gut the part where we're throwing an exception and make it an assert as well.

pksunkara: Member Element can contain any elements in key and value.

PR #490, opened 2017-07-20
Conform to our clang-format

kylef:
This PR conforms to the clang-format configuration that we previously agreed upon.

```
clang-format version 5.0.0 (tags/google/stable/2017-06-22)
```


kylef:
Does seem that our clang-format is breaking the syntax in some area (CI is failing).

w-vi:
@kylef I guess that it needs some little tweaking, have you tried it locally?

tjanc:
Copied the .clang-format from snowcrash/master and applied it after a rebase.

w-vi:
:clap:

.clang-format
```diff
@@ -0,0 +1,74 @@
+Language:        Cpp
```
kylef: We could probably DRY this up and make it a symbolic reference to the clang format in the sundown sub module.

tjanc: How will non-unix systems deal with that?

kylef: NTFS has symbolic links so I'd expect it to work on Windows too, but not 100% sure.

PR #492, opened 2017-08-07
Update build system to allow C++14

tjanc:
Based on the list of supported compilers, we decided to allow most C++14 features in drafter.
This PR merely modifies build system configuration to append the -std=c++14 flag and not to append -std=c++11. The latter was necessary because current SC build configuration appends -std=c++11, overriding previous declarations.

### DIFF
--- a/drafter.xcodeproj/project.pbxproj
+++ b/drafter.xcodeproj/project.pbxproj
@@ -705,7 +705,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				CLANG_CXX_LANGUAGE_STANDARD = "c++0x";
+				CLANG_CXX_LANGUAGE_STANDARD = "c++14";
 				CLANG_CXX_LIBRARY = "libc++";
 				CLANG_ENABLE_MODULES = YES;
 				CLANG_ENABLE_OBJC_ARC = YES;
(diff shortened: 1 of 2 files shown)


kylef:
I am not sure this is working as intended, there are conflicting configurations in snowcrash.

When I run `./configure` and `make` on macOS I see that `-std=c++0x -stdlib=libc+` is passed to the C++ compiler.

- c++11 flag is used https://github.com/apiaryio/snowcrash/blob/master/common.gypi#L106
- rtti is enabled: https://github.com/apiaryio/snowcrash/blob/master/common.gypi#L131
- c++0x standard is used for Mac: https://github.com/apiaryio/snowcrash/blob/master/common.gypi#L140

@tjanc Could you make a pull request to snowcrash first, and update this PR to reference the snowcrash commit so we can test this fully works on Windows/Linux CI and I can verify on macOS.

tjanc:
@kylef The intent was to allow C++14 in drafter, not in snowcrash. But I'm more than happy to move snowcrash forward as well.

tjanc:
[Update build system to allow C++14](https://trello.com/c/25ZQPzDD/781-update-build-system-to-allow-c14)

pksunkara:
Yeah, you should probably move them as well. Starting with markdown-parser, then snowcrash and then finally drafter.

kylef:
@tjanc Since the configuration of drafter is inherited from Snowcrash, when you build drafter it was being built as C++0X at least on macOS.

tjanc:
@kylef Agreed. I only supressed `-std=c++11`.

tjanc:
Updated snowcrash submodule and rebased.

PR #494, opened 2017-08-08
Refactoring allows reimplemt template specialization for different EnumElement processing

klokane:
Changes:
* ElementData now uses string for Primitives and RefractElements (std::vector<refract::IElement*>) for complex Element
* Reimplement specialisation for conversion processing
* Move checking of primitive types conversion
* Simplify stored info about ElementData
* Rename some function to better names

### DIFF
--- a/src/ConversionContext.h
+++ b/src/ConversionContext.h
@@ -23,6 +23,7 @@ namespace drafter {
         std::vector<snowcrash::Warning> warnings;
 
         inline refract::Registry& GetNamedTypesRegistry() { return registry; }
+        inline const refract::Registry& GetNamedTypesRegistry() const { return registry; }
 
         ConversionContext(const WrapperOptions& options)
         : options(options) {}
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -25,14 +25,26 @@ namespace drafter {
     struct ElementData {
         typedef T ElementType;
 
-        typedef typename T::ValueType ValueType;
+        typedef typename ElementType::ValueType ValueType;
         typedef snowcrash::SourceMap<ValueType> ValueSourceMapType;
 
-        typedef std::tuple<ValueType, ValueSourceMapType, bool> ValueInfo; // [value, sourceMap, validity(by LiteralTo<>)]
+        static const bool IsPrimitive = std::conditional<
+            std::is_same<ValueType, refract::RefractElements>::value, // check for primitive values
+            std::false_type, std::true_type>::type::value;
 
-        std::vector<ValueInfo> values;
-        std::vector<ValueInfo> defaults;
-        std::vector<ValueInfo> samples;
+        typedef typename std::conditional<
+            IsPrimitive, 
+            std::string,
+            refract::RefractElements
+            >::type StoredType;
+
+        typedef std::tuple<StoredType, ValueSourceMapType> ElementInfo; // [value, sourceMap]
+
+        typedef std::vector<ElementInfo> ElementInfoContainer;
+
+        ElementInfoContainer values;
+        ElementInfoContainer defaults;
+        ElementInfoContainer samples;
 
         std::vector<std::string> descriptions;
         std::vector<snowcrash::SourceMap<std::string> > descriptionsSourceMap;
@@ -60,7 +72,7 @@ namespace drafter {
         typedef T ElementType;
         typedef V ValueType;
         ElementType*& element;
-        typedef typename ElementData<T>::ValueInfo ValueInfo;
+        typedef typename ElementData<T>::ElementInfo ElementInfo;
 
         Append(ElementType*& e) : element(e)
         {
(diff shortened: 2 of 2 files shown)


src/RefractDataStructure.cc
```diff
@@ -25,14 +25,26 @@ namespace drafter {
     struct ElementData {
         typedef T ElementType;
 
-        typedef typename T::ValueType ValueType;
+        typedef typename ElementType::ValueType ValueType;
         typedef snowcrash::SourceMap<ValueType> ValueSourceMapType;
 
-        typedef std::tuple<ValueType, ValueSourceMapType, bool> ValueInfo; // [value, sourceMap, validity(by LiteralTo<>)]
+        static const bool IsPrimitive = std::conditional<
```
tjanc: Consider whether using a `static bool` here is what we want. The STL comes with the approach of using free type constructors like the following for such use cases:
```C++
template<typename T>
struct is_primitive {
  using type = std::integral_constant<bool, !std::is_same<T, refract::RefractElements>::value>
};
```
Keep in mind that std::true_type and std::false_type implicitly convert to a runtime bool if necessary, but by itself it is a compile-time concept that doesn't have a runtime imprint. The static bool above does.

src/ConversionContext.h
```diff
@@ -23,6 +23,7 @@ namespace drafter {
         std::vector<snowcrash::Warning> warnings;
 
         inline refract::Registry& GetNamedTypesRegistry() { return registry; }
+        inline const refract::Registry& GetNamedTypesRegistry() const { return registry; }
```
tjanc: Consider adding `noexcept`.

src/RefractDataStructure.cc
```diff
@@ -618,10 +646,10 @@ namespace drafter {
 
                 refract::ArrayElement* a = new refract::ArrayElement;
 
+                ElementInfoToElement<T> fetch;
+
                 for (auto sample : samples) {
```
tjanc: Please refactor this to use either `auto&` or `const auto&`. It isn't good practice to use mere `auto` in ranged for loops because of implicit copying.

src/RefractDataStructure.cc
```diff
@@ -415,22 +428,23 @@ namespace drafter {
...
 
-        template<bool dummy>
-        struct Fetch<RefractElements, dummy> {
-            ValueInfo operator()(const mson::TypeNames& typeNames, ConversionContext& context) {
+        template <typename E> 
+        struct Fetch<E, false> {
+            void operator()(const mson::TypeNames& typeNames, const ConversionContext& context, ElementInfoContainer& values) {
                 RefractElements types;
 
                 for (mson::TypeNames::const_iterator it = typeNames.begin(); it != typeNames.end(); ++it) {
```
tjanc: Consider refactoring this to use ranged for.

src/RefractDataStructure.cc
```diff
@@ -819,8 +854,7 @@ namespace drafter {
         }
 
         std::string description;
-        std::string& descriptionRef = description;
-        Join join(descriptionRef);
+        Join join(std::ref(description));
```
tjanc: Why is the reference_wrapper needed here?

src/RefractDataStructure.cc
```diff
@@ -607,6 +612,29 @@ namespace drafter {
...
+            E* operator()(const typename ElementData<E>::ElementInfo& value) {
+                return new E(std::get<0>(value));
+            }
+        };
+
+        //template <> 
+        //struct ElementInfoToElement<refract::EnumElement, ElementData<refract::EnumElement>::IsPrimitive> {
+        //    refract::EnumElement* operator()(const typename ElementData<refract::EnumElement>::ElementInfo& value) {
+        //    }
+        //};
```
pksunkara: Should we delete this part?

klokane: No, I will use it while enum impl :)

src/ConversionContext.h
klokane: ok

src/RefractDataStructure.cc
klokane: There were something in earlier version just place string into was not working.
We can actually remove

src/RefractDataStructure.cc
klokane: I did not change this part, but agree, it is good idea

src/RefractDataStructure.cc
klokane: There will be changes in Enum, let it pass for now. Will fix while enum impl

src/RefractDataStructure.cc
klokane: Ok, will change to type definition instead of value

src/RefractDataStructure.cc
```diff
@@ -25,14 +25,26 @@ namespace drafter {
     struct ElementData {
         typedef T ElementType;
 
-        typedef typename T::ValueType ValueType;
+        typedef typename ElementType::ValueType ValueType;
```
klokane: ok

src/RefractDataStructure.cc
tjanc: ok

pksunkara:
@tjanc @klokane Can we resolve this PR?

klokane:
Sry, I forgoit to close it. It is already merged via pull/498

PR #495, opened 2017-08-08
Hold std::vector as member instead of base class in MemberElementCollection

tjanc:
- std::vector is not meant to be derived from; it doesn't even have a virtual destructor. In the future, this could cause strange behaviour.
- Parts of the interface continues to be implemented sub-optimally or inconsistently; changing this was not in scope of the task, therefore FIXME comments were added where relevant

### DIFF
--- a/src/refract/Element.cc
+++ b/src/refract/Element.cc
@@ -47,34 +47,29 @@ namespace refract
     {
         ComparableVisitor v(name);
         Visitor visitor(v);
-        const_iterator it;
 
-        for (it = begin(); it != end(); ++it) {
-            (*it)->value.first->content(visitor);
-
-            if (v.get()) {
-                return it;
-            }
-        }
-
-        return it;
+        return std::find_if(elements.begin(), elements.end(),
+                            [&visitor, &v](const MemberElement* e) {
+                                e->value.first->content(visitor);
+                                return v.get();
+                            });
     }
 
     IElement::MemberElementCollection::iterator IElement::MemberElementCollection::find(const std::string& name)
     {
         ComparableVisitor v(name);
         Visitor visitor(v);
-        iterator it;
 
-        for (it = begin(); it != end(); ++it) {
-            (*it)->value.first->content(visitor);
+        return std::find_if(elements.begin(), elements.end(),
+                            [&visitor, &v](const MemberElement* e) {
+                                e->value.first->content(visitor);
+                                return v.get();
+                            });
+    }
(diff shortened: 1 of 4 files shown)


src/refract/Element.cc
```diff
@@ -43,38 +43,33 @@ namespace refract
         return reserved.find(element) != reserved.end();
     }
 
-    IElement::MemberElementCollection::const_iterator IElement::MemberElementCollection::find(const std::string& name) const
+    IElement::MemberElementCollection::Container::const_iterator IElement::MemberElementCollection::find(const std::string& name) const
```
klokane: What about alias `Container::const_iterator`  as `const_iterator` via `typedef`?

src/refract/Element.h
```diff
@@ -67,16 +67,52 @@ namespace refract
...
-            virtual ~MemberElementCollection();
+
+            /// clone elements from `other` to `this`
+            void clone(const MemberElementCollection& other);
+
+            // FIXME erase(const std::string) deletes pointer, whereas erase(iterator) does not
+            void erase(const std::string& key);
+            void erase(iterator it) { elements.erase(it); }
+
+            // FIXME pointers are not deleted
```
klokane: Where are actually pointers freed?

src/refract/Element.cc
tjanc: I added the alias afterwards, will introduce it.

klokane: :+1:

src/refract/Element.h
tjanc: That depends on the use case. MemberElementCollection::clear was called somewhere in the tests directly on MemberElementCollection while it still was deriving from std::vector. The implementation on line 110 mirrors this behaviour, i. e. pointers get dropped without deletion. The same applies for erase(iterator).
I already started fixing the spots in the code where both are used, but then stashed that away because the scope of this task is removing std::vector as a base in contrast to fixing memory management inconsistencies.

klokane: Do I understand correctly, it happen just only in tests?

tjanc: Clear is called only from tests AFAIR, but erase is actually used in production code. At that spot it is used to move an attribute from one subtree to the other.

klokane:
I'm ok with

klokane:
Just please add alias for iterators

pksunkara:
👍

PR #497, opened 2017-08-18
Use base GCC image in Dockerfile

kylef:
There are problems with the apiary base image as it is pretty large and Docker Hub is having trouble building it.


w-vi:
Is this going to be our official drafter image?

PR #498, opened 2017-08-21
Enum implementation

klokane:
It is build on top of my previous commit, most of Review requirement is implemented there

### Implementation notes

All Samples/Defaults for Enum Element are generated as is described there
https://gist.github.com/klokane/db31ca1383e33cc71a4e168e398c1271

Direct storing of values corrupt rule "store array or element of origin"

It is possible simply move it to enumeration as was discussed, but I want to have in first step mimmal amount of changes to be possible to check what happens on output.

JSONSchema do not add Samples/Default values into possible enums, for same reasun as above

### Missing things:

- RefractDataStructures.cc - Possible memory leak (valgrind says it is ok), 
it can happen while call `LastElementToAttribute()`, (saving default), if is picked just last element, no one of previous is not freed
Simplest solution: clone last element and free foll container

- ExpandVisitor.cc - commented out implementation for EnumElement (description in FIXME comment) it is not invoked. Possible reason is, we have no EnumElement expansion covered by tests

- PrintVisitor - no EnumElement implemented.

 - !!! Crash - there is crashing test on 
./test/test-RefractAPITest.cc:35
It seem, there is problem with double freeing, but can not find reason:
it is for fixture `test/fixtures/api/attributes-named-type-enum-reference.apib`

- Test output - changes for generated JSON in Enum for `circular` need to decide about correct form
    - test/fixtures/mson/enum-empty-default.json
    - test/fixtures/circular/array.json
    - test/fixtures/circular/enum.json
    - test/fixtures/circular/array-in-object.json
    - test/fixtures/extend/circular.json

In original file there were generated empty `array` for circular definition, in current it is `object` I'm not sure about both of them


src/ElementData.h
```diff
@@ -0,0 +1,57 @@
...
+        typedef snowcrash::SourceMap<ValueType> ValueSourceMapType;
+
+        typedef typename std::conditional<
+            std::is_same<ValueType, refract::RefractElements>::value || std::is_same<ValueType, refract::IElement*>::value, // check for primitive values
+            std::false_type, std::true_type>::type
+            IsPrimitive;
+
+        // This is required because snowcrash internal stuctures holds data 
+        // for primitive types as "string" for complex types as "element array"
+        // it will be converted into apropriated element typer once all required data are colected
```
kylef: I think `typer` is a typo.

```diff
- element typer
+ element type
```

src/RefractAPI.cc
```diff
@@ -90,6 +90,7 @@ namespace drafter {
         refract::IElement* msonElement = MSONToRefract(dataStructure, context);
 
         if (context.options.expandMSON) {
+        //if (true) {
```
kylef: NIT: Left over from debugging?

src/RefractDataStructure.cc
```diff
@@ -446,55 +405,97 @@ namespace drafter {
...
-        void operator()(const NodeInfo<mson::TypeDefinition>& typeDefinition) {
-            ValueInfo value = Fetch<typename T::ValueType>()(typeDefinition.node->typeSpecification.nestedTypes, context);
+        template <typename E, typename IsPrimitive = typename ElementData<E>::IsPrimitive, bool dummy = true> struct Store;
 
-            if (std::get<2>(value)) {
-                data.values.push_back(value);
+        template <typename E, bool dummy> 
+        struct Store<E, PrimitiveType, dummy> {
+            void operator()(ElementData<E>& data, const mson::TypeNames& typeNames, const ConversionContext& context){
+                // do nothing
```
kylef: This comment doesn't add any value. It is stating what this does not why. The what it does it clear from the code (it literally does nothing), the why is unclear so might be worth adding why (or removing the comment).

src/RefractDataStructure.cc
```diff
@@ -446,55 +405,97 @@ namespace drafter {
...
+                const mson::BaseTypeName type = SelectNestedTypeSpecification(valueMember.node->valueDefinition.typeDefinition.typeSpecification.nestedTypes);
+
+                const RefractElementFactory& f = FactoryFromType(type);
+                const mson::Values& values = valueMember.node->valueDefinition.values;
 
-                ValueInfo result = std::make_tuple(val.second, sourceMap, val.first);
+                RefractElements elements;
 
-                CheckValueValidity<T>()(result, context);
+                for (mson::Values::const_iterator it = values.begin(); it != values.end(); ++it) {
```
kylef: Can we simplify this with high order functions like map while reducing the amount of mutation (and possibly allowing the optimiser to do a better job)?

```cpp
const RefractElements elements = std::transform(values.begin(), values.end(), ^ (mson::Values::const_iterator iterator) {
    return f.Create(it->literal, it->variable ? eSample : eValue);
});
```

I apologise, this example code is Pseudo-C++ as I do not remember the C++ block syntax off the top of my head

drafter.xcodeproj/project.pbxproj
```diff
@@ -18,7 +18,6 @@
 		19A1298F1B70ABE100366AA7 /* SerializeResult.h in Headers */ = {isa = PBXBuildFile; fileRef = 19A129791B70ABE100366AA7 /* SerializeResult.h */; };
 		19A129921B70ABE100366AA7 /* stream.h in Headers */ = {isa = PBXBuildFile; fileRef = 19A1297C1B70ABE100366AA7 /* stream.h */; };
 		19A129931B70ABE100366AA7 /* Version.h in Headers */ = {isa = PBXBuildFile; fileRef = 19A1297D1B70ABE100366AA7 /* Version.h */; };
-		19A129AF1B70AC9A00366AA7 /* AppendDecorator.h in Headers */ = {isa = PBXBuildFile; fileRef = 19A129971B70AC9A00366AA7 /* AppendDecorator.h */; };
```
kylef: NOTE: file was removed, but new files was not added (test/test-ElementDataTest.cc)

src/refract/Element.h
```diff
@@ -242,6 +241,7 @@ namespace refract
             }
 
             if (flags & cValue) {
+                element->hasContent = self->hasContent;
```
kylef: good catch

src/refract/ExpandVisitor.cc
```diff
@@ -64,14 +64,23 @@ namespace refract
...
+        template <>
         struct ExpandValueImpl<RefractElements> {
 
             template <typename Functor>
             RefractElements operator()(const RefractElements& value, Functor& expand) {
                 RefractElements members;
 
-                for (RefractElements::const_iterator it = value.begin() ; it != value.end() ; ++it) {
-                    members.push_back(expand(*it));
+                for (auto const& item: value) {
```
kylef: Could be simplified using high-order transform?

```c++
const RefractElements members = std::transform(value.begin(), value.end(), expand);
```

src/refract/RenderJSONVisitor.cc
```diff
@@ -116,20 +86,25 @@ namespace refract
         RenderJSONVisitor renderer;
 
         if (e.value.second) {
-            if (IsTypeAttribute(e, "nullable") && e.value.second->empty()) {
+            if (EnumElement* enm = TypeQueryVisitor::as<EnumElement>(e.value.second)) {
+                // We nned nadle Enum individualy because of attr["enumerations"]
```
kylef: Some typos here, not sure of the intent trying to be conveyed from the comment.

> We nned nadle Enum

src/refract/RenderJSONVisitor.cc
```diff
@@ -143,24 +118,19 @@ namespace refract
...
-            }
-            else {
-                enumValue = new StringElement;
-            }
+        const IElement* val = GetValue<EnumElement>(e);
+        if (val && !val->empty()) {
+            val = val->clone();
+        }
+        else {
+            val = new StringElement;
```
kylef: Wondering why we default to a string element? Shouldn't there be no value if I have an empty enum (i.e):

```apib
+ myThing (enum)
```

src/refract/VisitorUtils.h
```diff
@@ -125,32 +125,68 @@ namespace refract
...
 
-        operator const EnumElement::ValueType*() {
+        operator const IElement*() {
+            return GetEnumValue(element);
+        }
+
+        IElement* GetEnumValue(const EnumElement& element) {
             if (const EnumElement* s = GetSample(element)) {
-                return &s->value;
+                return GetEnumValue(*s);
```
kylef: I don't understand why this would be changed to an enum specific extraction of the sample value. Now that enum default/sample/value behaviour is identical to other elements surely it could have the same behaviour, except for falling back to `enumerations` when no values/samples/defaults are found?

src/refract/VisitorUtils.h
```diff
@@ -125,32 +125,68 @@ namespace refract
...
-            if (element.empty() && IsTypeAttribute(element, "nullable")) {
-                return NULL;
+            if (const ArrayElement* e = GetEnumerations(element)) {
+                if (e && !e->empty()) {
+                    for (const auto& item : e->value) {
+                        if (!item) {
+                            continue;
+                        }
+
+                        // We need hadle Enum individualy because of attr["enumerations"]
```
kylef:
```diff
- hadle
+ to handle
```

src/refract/VisitorUtils.h
```diff
@@ -125,32 +125,68 @@ namespace refract
...
+            if (element.value) {
+                return element.value;
             }
 
-            if (element.empty() && IsTypeAttribute(element, "nullable")) {
-                return NULL;
+            if (const ArrayElement* e = GetEnumerations(element)) {
+                if (e && !e->empty()) {
+                    for (const auto& item : e->value) {
+                        if (!item) {
```
kylef: Maybe stupid question, is there cases where nil would be in the array and enumerating over it gives nil as a value?

drafter.xcodeproj/project.pbxproj
klokane: have no Xcode and do not know how to "add"

src/refract/ExpandVisitor.cc
klokane: Seems to be good idea

src/refract/VisitorUtils.h
klokane: Usually not, but when it happen, we will end with SIGFAULT crash.

src/refract/VisitorUtils.h
klokane: There is difference between Enum and Other Elements. For Enum you need a pick one value from possible values.

Another "complication" is we want to be backward compatible (at least first time of iteration) before redefine Enum behavior

src/refract/VisitorUtils.h
klokane: In general. It array of pointers. It is really good idea to check if pointer is not NULL every time

klokane:
@apiaryio/adt review call. 
Please do not merge once finished review.
I would like to have review over this branch before merging, to avoid changes in codebase related to master.
I will make rebase on top of master once review will be done

src/refract/VisitorUtils.h
kylef: > For Enum you need a pick one value from possible values.

That should not be true in the new form, `default` is an element and `samples` is an array of samples (array of elements) so this matches other element behaviour:

https://github.com/apiaryio/api-elements/pull/28

klokane: Then we will have "non compatible" building of elements default/samples for enum. All others contain full element in default/sample.

And it mean there will be again incompatibility for calling `GetSample()`, `GetDefault()` voth returns Same Type of Element

klokane: You typically call GetSample/GetValue and pick value from them.
```
ArrayElement* a = GetDefault<ArrayElement>(element);
if (a->empty()) {
  return a->value;
}
```

klokane:
@apiaryio/adt review call.

src/refract/Element.cc
```diff
@@ -217,56 +227,56 @@ namespace refract
...
                             }
-                        } else if (RefElement* ref = TypeQueryVisitor::as<RefElement>(*it)) {
+                        } else if (RefElement* ref = TypeQueryVisitor::as<RefElement>(it)) {
                             refBase[ref->value] = ref;
                         }
                     }
 
-                    for (RefractElements::const_iterator it = merge.value.begin(); it != merge.value.end(); ++it) {
-                        if (MemberElement* member = TypeQueryVisitor::as<MemberElement>(*it)) {
+                    for (auto const& it : merge.value) {
```
tjanc: Add a C assert on the pointer element. I imagine a `nullptr` here hard to debug as the below line would seem like it couldn't interpret something as a `MemberElement`.

src/refract/Element.cc
```diff
@@ -217,56 +227,56 @@ namespace refract
                     MapKeyToMember keysBase;
                     MapNameToRef refBase;
 
-                    for (RefractElements::iterator it = value.begin(); it != value.end(); ++it) {
-                        if (MemberElement* member = TypeQueryVisitor::as<MemberElement>(*it)) {
+                    for (auto& it : value) {
```
tjanc: Add a C assert on the pointer element. I imagine a `nullptr` here hard to debug as the below line would seem like it couldn't interpret something as a `MemberElement`.

src/ElementData.h
```diff
@@ -0,0 +1,53 @@
...
+#define DRAFTER_ELEMENTDATA_H
+
+#include "RefractSourceMap.h"
+
+namespace drafter
+{
+
+    template <typename T>
+    struct ElementData {
+        typedef T ElementType;
```
tjanc: Prefer aliases (`using`) over `typedef`s.

src/ElementData.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef DRAFTER_ELEMENTDATA_H
+#define DRAFTER_ELEMENTDATA_H
+
+#include "RefractSourceMap.h"
+
+namespace drafter
+{
+
+    template <typename T>
+    struct ElementData {
```
tjanc: If I understand this structure correctly, it holds a set of types and values that are compile-time dependent on a template parameter, `T`. Why are they encapsulated in a structure instead of a namespace?

klokane: It collect data related to element which does come from snowcrash and are lately used to create appropriate element with required meta/attributes

src/refract/Element.cc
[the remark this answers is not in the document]
klokane: You can thing about it as short if `item` :)
But ok, will change

src/refract/Element.cc
```diff
@@ -192,6 +187,21 @@ namespace refract
                 }
             };
 
+            template <typename T>
+            struct ValueMerge<T, IElement*> {
+                IElement*& value;
```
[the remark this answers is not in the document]
klokane: In current impl it is problem. All other cases count with reference (in meaning `Element::ValueType&`
If I remove ref, in this case, there is failing test.

src/refract/VisitorUtils.h
klokane: Especially in this case it is not error, it here not will be element. 
Enum element does not require to be "Enumerations" attribute filled or its existence

PR #502, opened 2017-08-29
chore: Polish mono repo build system

pksunkara:


vcbuild.bat
```diff
@@ -130,6 +129,8 @@ if errorlevel 1 goto exit
 @rem Run tests if requested.
 if "%test%"=="" goto intigration-test
 echo Running tests...
+.\build\%config%\test-libmarkdownparser.exe
+.\build\%config%\test-libsnowcrash.exe
```
kylef: Can you add the c-api tests too please? Seems that was never run.

pksunkara: Fixed.

PR #505, opened 2017-09-26
Fix #493 repeated item in JSON schema required

### LINKED ISSUE #493: Drafter allows duplicate attribute and generates JSON Schema with duplicate required fields
Following document

```apib
# Data Structures

## Struct (object)
+ field (required)
+ field (required)

# test [GET /]

    + Attributes (object)
        + data (required, array[Struct], fixed-type)
```

will result in no user warning and in JSON Schema with `"required": ["field", "field"]`. I think this situation (basically user typo) should be handled more gracefully on parser level. See https://github.com/apiaryio/dredd/issues/840 reported by @basavind for more details.

klokane:
Use internally `std::set` to catch all "required" members and avoid repeat value


src/refract/JSONSchemaVisitor.cc
```diff
@@ -575,40 +578,43 @@ namespace refract
...
         RefractElements& oneOfMembers,
         ObjectElement* o)
     {
-        for (auto it = begin; it != end; ++it) {
-            if (!*it) {
+        std::set<std::string> required;
+
+        for (const auto& member: members) {
+            assert(member);
+            if (!member) {
```
tjanc: This conditional is always entered. Otherwise line 590 would assert.

src/refract/JSONSchemaVisitor.cc
```diff
@@ -637,5 +643,10 @@ namespace refract
                     throw LogicError("Invalid member type of object in MSON definition");
             }
         }
+
+        std::for_each(required.begin(), required.end(),
```
tjanc: Use std::transform with std::inserter.

src/refract/JSONSchemaVisitor.cc
klokane: Agree, but I do not want to crash drafter with `nullptr` on production

tjanc: Drafter wouldn't necessarily crash. It would yield to undefined behaviour, which is even worse. Nevertheless, the assert makes the branch impossible to unit test, so either choose an assert, or a conditional.

tjanc: And if you use a conditional, that path is definitely untested without unit tests for JSONSchemaVisitor.

klokane: In general, there can be nullptr.
Then I will remove assert

klokane: There is problem with testing, it is private function. We need bigger refactoring for this.

klokane: At first look, it seem it is independent function and it can be isolated out of class

klokane: So it use members `fixed`, `fixedType`, `pDefs`.
Will try to fix it, but not sure if able finish before sprint end

PR #506, opened 2017-10-11
refactor: Change enum element format in parameters

pksunkara:


test/fixtures/api/resource-parameters.json
```diff
@@ -84,44 +84,37 @@
                         "value": {
                           "element": "enum",
                           "attributes": {
-                            "samples": {
+                            "default": {
                               "element": "array",
```
kylef: The default value here is still being wrapped in an array which will need to be removed.

pksunkara:
Fixed.

src/RefractAPI.cc
```diff
@@ -142,23 +142,25 @@ namespace drafter
...
         }
 
         // Add default value
         if (!parameter.node->defaultValue.empty()) {
             element->attributes[SerializeKey::Default]
-                = CreateArrayElement(LiteralToRefract<std::string>(MAKE_NODE_INFO(parameter, defaultValue), context));
+                = LiteralToRefract<std::string>(MAKE_NODE_INFO(parameter, defaultValue), context);
         }
 
+        // Add enumeraitons
```
kylef: This comment has a typo (`enumeraitons`).

I'm not entirely sure that this comment adds any value as it is obvious this code does what the comment says. Generally commenting why something is done is valuable, but what something done doesn't seem useful.

src/RefractAPI.cc
```diff
@@ -142,23 +142,25 @@ namespace drafter
     refract::IElement* ParameterValuesToRefract(
         const NodeInfo<snowcrash::Parameter>& parameter, ConversionContext& context)
     {
-        refract::ArrayElement* element = CollectionToRefract<refract::ArrayElement>(
-            MAKE_NODE_INFO(parameter, values), context, LiteralToRefract<std::string>, SerializeKey::Enum);
+        refract::EnumElement* element = new refract::EnumElement;
+
+        refract::ArrayElement* enumerations = CollectionToRefract<refract::ArrayElement>(
```
kylef: Just personal preference, but I think it would be more readable of the `enumerations` was defined next to where they are set below on line 161. They are not used until that location.

```c++
refract::ArrayElement* enumerations = CollectionToRefract<refract::ArrayElement>(
    MAKE_NODE_INFO(parameter, values), context, LiteralToRefract<std::string>);
element->attributes[SerializeKey::Enumerations] = enumerations;
```

pksunkara: Done.

PR #507, opened 2017-10-17
fix: Infinite loop issue

pksunkara:
Snowcrash should give out a circular reference error when an attirbutes
section is reference itself even when a description header node is present
before it.

This is actually a better fix for #328 than the one we did in #329

### DIFF
--- a/ext/snowcrash/src/BlueprintParser.h
+++ b/ext/snowcrash/src/BlueprintParser.h
@@ -181,7 +181,8 @@ namespace snowcrash
 
                         contextSectionType = sectionType;
                         contextCur = cur;
-                    } else if (contextSectionType != DataStructureGroupSectionType) {
+                    } else if (contextSectionType != DataStructureGroupSectionType
+                        && SectionProcessor<Action>::sectionType(cur) == ActionSectionType) {
 
                         contextSectionType = UndefinedSectionType;
                     }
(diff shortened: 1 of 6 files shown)


ext/snowcrash/src/BlueprintParser.h
```diff
@@ -181,7 +181,8 @@ namespace snowcrash
 
                         contextSectionType = sectionType;
                         contextCur = cur;
-                    } else if (contextSectionType != DataStructureGroupSectionType) {
+                    } else if (contextSectionType != DataStructureGroupSectionType
+                        && SectionProcessor<Action>::sectionType(cur) == ActionSectionType) {
```
kylef: I'm wondering what the difference would be between `SectionProcessor<Action>::sectionType(cur) == ActionSectionType` and `sectionType == ActionSectionType`?

pksunkara: `sectionType` will be `UndefinedSectionType` because in BlueprintParser's `nestedSectionType()`, we don't check for `ActionType` since it is not a direct child of Blueprint.

kylef: The test is still passing with this though: https://github.com/apiaryio/drafter/commit/d92137c7aa30af978fe96cecd90b136701781a11

pksunkara: I remember other tests failing with that. It is the first thing I tried too.

PR #513, opened 2017-10-23
Fix crash - accessing RefElement via non resolved member

klokane:
Please do not merge. Fixed but still isolating test case

### DIFF
--- a/src/refract/Element.cc
+++ b/src/refract/Element.cc
@@ -293,7 +293,7 @@ namespace refract
                             }
                         } else if (RefElement* ref = TypeQueryVisitor::as<RefElement>(element)) {
                             if (refBase.find(ref->value) == refBase.end()) {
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Master
 
+## Bug Fixes
```
kylef: There is already a `Bug Fixes` section below for this release.

Also, since this is fixing a bug there was introduced in master I don't think it makes sense to put it in the CHANGELOG. It wasn't present in the previous release either.

klokane: So you propose to remove CHANGELOG record? Do I understand correctly?

kylef: Yes, I don't think it makes sense to have a CHANGELOG entry for something that isn't a change since the last release.

klokane: rollback finished && push forced

PR #514, opened 2017-10-24
chore: Add clang-format to CI

pksunkara:


circle.yml
```diff
@@ -24,6 +25,7 @@ test:
     - env CXXFLAGS="-Werror" make test
     - valgrind --leak-check=full --show-leak-kinds=definite,indirect,possible --error-exitcode=1 ./bin/test-libdrafter 2> memcheck.log
     - valgrind --leak-check=full --show-leak-kinds=definite,indirect,possible --error-exitcode=1 ./bin/test-capi 2> memcheck-capi.log
+    - ./tools/clang-format-check.sh
```
tjanc: I'd rather have clang-format reject my changes before the build, not after. It would be frustrating to wait for a successful build just to get an error message about formatting at the end.

tools/clang-format-check.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+clang-format-5.0 -style=file -output-replacements-xml src/* src/refract/* test/* 2>/dev/null | grep -c "<replacement "
```
tjanc: There are three problems with how relevant files are passed to clang-format.
1. Not all of drafter's source files live directly in `src/*`, `src/refract/*` or `test/*`. Our former dependencies in `ext/` for example.
2. Every file in `src/*`, `src/refract/*` or `test/*` gets checked. A file not conforming to C++ syntax will necessarily fail (e.g. a grammar for a code generator, a README, etc.).
3. Sources in new subdirectories, such as `src/refract/data`, will not be checked even if added to the build configuration. It would be easy to forget adding a new subdirectory to some script in `tools` and getting a false positive from CI.

I can see two ways how to fix this:
1. Change this line to something in the form of `find src test -iname '*.h' -o -iname '*.cc' -not -path 'test/vendor' -not -path 'test/fixtures' | xargs clang-format ...`. Add similar lines for SC and friends.
2. Extract source names from the gyp file.

kylef: If it makes it any simpler, `git ls-files | grep  '\.cc\|\.h$'` can be used to get all the `h` and `cc` files tracked via git. This includes everything in `ext` (snowcrash and markdown-parser) but it does not include the submodules that are out of our control.

pksunkara:
Updated.

circle.yml
```diff
@@ -15,12 +15,14 @@ checkout:
...
   pre:
-    - sudo apt-get update && sudo apt-get install valgrind
+    - sudo add-apt-repository 'deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-5.0 main'
+    - sudo apt-get update && sudo apt-get install valgrind clang-format-5.0
 
 test:
   pre:
     - ./configure -i
   override:
+    - ./tools/clang-format-check.sh
```
tjanc: Does this get executed before the build now?

pksunkara: Yes

PR #520, opened 2017-11-22
Introduce interfaces for content types and fix memory ownership

tjanc:
Temporary PR for review


src/ElementData.h
```diff
@@ -9,27 +9,70 @@
 #define DRAFTER_ELEMENTDATA_H
 
 #include "RefractSourceMap.h"
+#include <deque>
 
 namespace drafter
 {
+    template <typename ElementT, typename = std::enable_if<std::is_base_of<refract::IElement, ElementT>::value> >
+    constexpr bool is_primitive = false;
```
klokane: not sure about introduce new naming conventions which does not conform to current conventions.
I personally prefer std c++ naming conventions, but this is team project

src/RefractAPI.cc
```diff
@@ -19,551 +21,537 @@
...
-        // Add sample value
-        if (!parameter.node->exampleValue.empty()) {
-            refract::ArrayElement* samples = new refract::ArrayElement;
-            samples->push_back(
-                CreateArrayElement(LiteralToRefract<std::string>(MAKE_NODE_INFO(parameter, exampleValue), context)));
-            element->attributes[SerializeKey::Samples] = samples;
+    // Add default value
+    if (!parameter.node->defaultValue.empty()) {
+        element->attributes().set(SerializeKey::Default,
+            CreateArrayElement(LiteralToRefract(MAKE_NODE_INFO(parameter, defaultValue), context)));
```
klokane: You have replace on line 103 `CreateArrayElement()` by `generate_element<ArrayElement>()`
We shoulde decide to replace all of them (not just there), or leave it untouched for this "turn"

src/RefractAPI.cc
```diff
@@ -19,551 +21,537 @@
...
+    if (asset.node->empty()) {
+        return nullptr;
+    }
 
-        element->set(refract::IElement::Create(header.node->first), refract::IElement::Create(header.node->second));
+    auto element = PrimitiveToRefract(asset);
 
-        AttachSourceMap(element, header);
+    element->element(SerializeKey::Asset);
+    element->meta().set(SerializeKey::Classes, CreateArrayElement(make_primitive(metaClass)));
```
klokane: In general, I understand, it will create "primitive element", but It is little bit confusing to have `make_primitive()` and `PrimitiveToRefract()`
I propose something like `from_primitive()`

And there is still problem with naming Convention (see `is_primitive()`)

src/RefractDataStructure.cc
```diff
@@ -225,48 +227,49 @@ namespace drafter
...
-            ElementInfo operator()(const NodeInfo<mson::TypeSection>& typeSection,
+        struct Fetch<U, true> {
+            ElementInfo<T> operator()(const NodeInfo<mson::TypeSection>& typeSection,
                 ConversionContext& context,
                 const mson::BaseTypeName& defaultNestedType)
             {
 
                 snowcrash::SourceMap<ValueType> sourceMap = FetchSourceMap<ValueType>()(typeSection);
-                return std::make_tuple(typeSection.node->content.value, sourceMap);
+                return ElementInfo<T>{ typeSection.node->content.value, sourceMap };
```
klokane: :+1:

src/RefractDataStructure.cc
```diff
@@ -1278,25 +1230,27 @@ namespace drafter
     }
 
     template <typename T>
-    refract::IElement* RefractElementFromMSON(const NodeInfo<snowcrash::DataStructure>& ds, ConversionContext& context)
+    std::unique_ptr<IElement> RefractElementFromMSON(
+        const NodeInfo<snowcrash::DataStructure>& ds, ConversionContext& context)
     {
         using namespace refract;
         typedef T ElementType;
```
klokane: using?

src/RefractDataStructure.cc
```diff
@@ -1278,25 +1230,27 @@ namespace drafter
...
+        AttachSourceMap(*element, MakeNodeInfo(ds.node, ds.sourceMap));
 
         // there is no source map for attributes
-        if (refract::IElement* attributes = MsonTypeAttributesToRefract(ds.node->typeDefinition.attributes)) {
-            element->attributes[SerializeKey::TypeAttributes] = attributes;
+        if (auto attributes = MsonTypeAttributesToRefract(ds.node->typeDefinition.attributes)) {
+            element->attributes().set(SerializeKey::TypeAttributes, std::move(attributes));
         }
 
         ElementData<T> data;
```
klokane: `ElementData<ElementType>`

src/RefractDataStructure.h
```diff
@@ -16,11 +16,11 @@ namespace drafter
 
     class ConversionContext;
 
-    refract::IElement* MSONToRefract(
+    std::unique_ptr<refract::IElement> MSONToRefract(
         const NodeInfo<snowcrash::DataStructure>& dataStructure, ConversionContext& context);
-    refract::IElement* ExpandRefract(refract::IElement* element, ConversionContext& context);
+    std::unique_ptr<refract::IElement> ExpandRefract(std::unique_ptr<refract::IElement> element, ConversionContext& context);
```
klokane: Probably not conform to ClangFormat spec, pls reindent

src/RefractElementFactory.cc
```diff
@@ -7,88 +7,90 @@
...
-                    if (value.first) {
-                        samples->push_back(refract::IElement::Create(value.second));
-                    }
-                    element->attributes[SerializeKey::Samples] = samples;
-                } break;
+
+                    auto element = make_empty<E>();
+
+                    element->attributes().set(SerializeKey::Samples,
+                        value.first ? CreateArrayElement(make_element<E>(value.second)) : make_empty<ArrayElement>());
```
klokane: `generate_element<>()`

src/RefractElementFactory.cc
```diff
@@ -7,88 +7,90 @@
...
+
+                    return std::move(element);
+                }
 
                 case eValue: {
                     std::pair<bool, ValueType> value = LiteralTo<ValueType>(literal);
                     if (value.first) {
-                        element->set(value.second);
+                        return make_element<E>(value.second);
```
klokane: what about
```
return value.first
  ? make_element<A>(value.second)
  : make_empty<A>();
```

src/RefractElementFactory.cc
```diff
@@ -7,88 +7,90 @@
...
+                case eElement: {
+                    auto element = make_empty<E>();
                     element->element(literal);
-                    break;
+                    return std::move(element);
+                }
             }
 
-            return element;
+            assert(false);
```
klokane: :+1:

src/RefractSourceMap.cc
```diff
@@ -1,27 +1,43 @@
 #include "RefractSourceMap.h"
 
-static refract::IElement* CharacterRangeToRefract(const mdp::CharactersRange& sourceMap)
+using namespace refract;
+
+namespace
 {
-    refract::ArrayElement* range = new refract::ArrayElement;
+    std::unique_ptr<IElement> CharacterRangeToRefract(const mdp::CharactersRange& sourceMap)
+    {
```
klokane:
```
return generate_element<ArrayElement>([&sourceMap](auto& data) {
  data.push_back(make_primitive(static_cast<double>(sourceMap.location))
  data.push_back(make_primitive(static_cast<double>(sourceMap.length)));
}      
```

src/Serialize.h
```diff
@@ -184,20 +180,13 @@ namespace drafter
...
 
-    refract::ArrayElement* CreateArrayElement(refract::IElement* value);
-
-    template <typename T>
-    refract::ArrayElement* CreateArrayElement(const T& content)
-    {
-        refract::IElement* value = refract::IElement::Create(content);
-        return CreateArrayElement(value);
-    }
+    std::unique_ptr<refract::ArrayElement> CreateArrayElement(std::unique_ptr<refract::IElement> value);
```
klokane: Remove because `generate_element<>()`?

src/refract/Element.h
```diff
@@ -2,798 +2,186 @@
...
+        using DataT = typename ElementT::ValueType;
+        return std::make_unique<ElementT>(DataT{ std::forward<Args>(args)... });
+    }
 
-        SelectElement(const TraitType::ValueType& value) : Type()
-        {
-            set(value);
-        }
+    template <typename Primitive, typename DataT = typename data::data_of<Primitive>::type>
+    auto make_primitive(const Primitive& p)
```
klokane: In earlier comment i suggest to rename `from_primitive` 
Reason is existence of `RefractToPrimitive()` and it is confising "what primitive"
I understand it is supplementary function to `is_primitive` but i thing it is confusing (at least for me)

klokane: And another one proposal. to avoid calling like this
```
make_primitive(static_cast<double>(annotation.code));
```
Add into DSD definition of supported "conversion" like:
* number -> std::is_numeric
* string -> std::string, const char*

src/refract/ElementFwd.h
```diff
@@ -2,37 +2,52 @@
...
-    struct ArrayElement;
-    struct EnumElement;
-    struct ObjectElement;
-
-    struct RefElement;
-    struct ExtendElement;
-
-    struct OptionElement;
-    struct SelectElement;
+    namespace data
```
klokane: maybe rename to `dsd`, it is presented with given name in comments, why not use it in namespace?

src/refract/InfoElements.h
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        const_iterator find(const std::string& name) const;
+        iterator find(const std::string& name);
+
+        IElement& set(const std::string& key, std::unique_ptr<IElement> value);
+        IElement& set(const std::string& key, const IElement& value);
+
+        /// clone elements from `other` to `this`
+        void clone(const InfoElements& other);
```
klokane: why are `beging/end` methods imlementation in header, while `erase/clear/empty/size` are in `cc` file?
All of them just forwarding call to member `elements`.

src/refract/JSONSchemaVisitor.cc
```diff
@@ -7,646 +7,664 @@
...
+            throw LogicError("Invalid member type of object in MSON definition");
+    }
+}
 
-                        if (!o1->value.empty()) {
-                            MemberElement* m1 = TypeQueryVisitor::as<MemberElement>(o1->value[0]->clone());
+std::string refract::renderJsonSchema(const IElement& e)
+{
+    auto pDefs = make_element<ObjectElement>();
+    JSONSchemaVisitor gen_schema(*pDefs);
```
klokane: While naming in new parts is possible to discuss, in old code base I'm absolutely against.

src/refract/JSONSchemaVisitor.cc
```diff
@@ -7,646 +7,664 @@
...
-            fixedType = true;
-        }
+    if (IsTypeAttribute(e, "fixed")) {
+        fixed = true;
+    }
 
-        if (fixed || fixedType) {
-            ArrayElement::ValueType av;
-            bool allEmpty = allItemsEmpty(val);
+    if (IsTypeAttribute(e, "fixedType")) {
```
klokane: We probably should rename this to `HasTypeAttribute()`
I know it is not "your" code, but while review I found it :)

src/refract/JSONSchemaVisitor.cc
```diff
@@ -7,646 +7,664 @@
...
-            addMember(str->value, renderer.getOwnership());
-        } else {
-            throw LogicError("A property's key in the object is not of type string");
         }
 
-        if (ext && str) {
-            delete str; // if Extend -> remove resolved (merged) element
-        }
+        addMember(keyStr->empty() ? std::string{} : std::string{ keyStr->get() }, renderer.getOwnership());
+    } else {
```
klokane: What about remove this condition and replace it by `assert(keyStr)` in precondition

src/refract/data/Object.cc
```diff
@@ -0,0 +1,72 @@
...
+
+    return elements_.insert(it, std::move(el));
+}
+
+object_t::iterator object_t::erase(object_t::const_iterator b, object_t::const_iterator e)
+{
+    return elements_.erase(b, e);
+}
+
+object_t::iterator object_t::add_member(std::string name, std::unique_ptr<IElement> value)
```
klokane: Naming convention `addMember()`
 
I'm ok with members 'member_' it is good to see diff between members and other variables

test/test-ApplyVisitorTest.cc
```diff
@@ -88,14 +85,16 @@ TEST_CASE("It should invoke Functor for member of container elements", "[Visitor
...
+
     v.visit(*e);
 
     REQUIRE(f.GCounter == 1); // just array
     REQUIRE(f.SCounter == 0);
-
-    delete e;
 }
 
 TEST_CASE("It should recognize Element Type by `Is` type operand", "[Visitor]")
```
klokane: By blame, my shame :)
We should implement or remove it completly
test report non-existent behavior....

test/refract/data/test-String.cc
```diff
@@ -0,0 +1,132 @@
...
+            THEN("its data is B")
+            {
+                REQUIRE(string2.get() == B);
+            }
+            THEN("its value is B")
+            {
+                const std::string P = string2;
+                REQUIRE(B == P);
+            }
+        }
```
klokane: add test when value of copy is changed do not change value of original
and vice versa for move c-tor

test/refract/data/test-Select.cc
```diff
@@ -0,0 +1,15 @@
...
+//  Copyright (c) 2017 Apiary Inc. All rights reserved.
+//
+
+#include "catch.hpp"
+
+#include "refract/data/Select.h"
+
+using namespace refract;
+using namespace data;
+
```
klokane: Missing test implementation.
Same for:
- ref
- option
- object
- null
- holder
- enum

test/refract/data/test-Number.cc
```diff
@@ -0,0 +1,114 @@
...
+            THEN("its data is N")
+            {
+                REQUIRE(number2.get() == N);
+            }
+            THEN("its value is N")
+            {
+                const double P = number2;
+                REQUIRE(N == P);
+            }
+        }
```
klokane: add test when value of copy is changed do not change value of original
and vice versa for move c-tor

test/refract/data/test-Member.cc
```diff
@@ -0,0 +1,215 @@
...
+            {
+                REQUIRE(mock->_total_ctx == 0);
+            }
+
+            THEN("no further ElementMocks are constructed")
+            {
+                REQUIRE(test::ElementMock::instances().size() == 2);
+            }
+        }
+    }
```
klokane: It seems (by interface) we support change of key value
add test for this case
if not supported, we should add test case to avoid

test/refract/data/test-InfoElements.cc
```diff
@@ -0,0 +1,218 @@
...
+
+SCENARIO("Elements can be added and erased from InfoElements", "[InfoElements]")
+{
+    GIVEN("An empty InfoElements")
+    {
+        InfoElements collection;
+
+        WHEN("an ElementMock pointer is added")
+        {
+            const char* key_ = "id";
```
klokane: why underscore suffix?

test/refract/data/test-Extend.cc
```diff
@@ -0,0 +1,209 @@
...
+                                const IElement* el2 = samplesArray->get().begin()[1].get();
+                                REQUIRE(dynamic_cast<const StringElement*>(el2)->get() == "zul");
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
```
klokane: - missing test cases for merging other Elements (just content testing is enough for this cases)
- missing test cases to merge incompatible Elements (e.g. StringElement and NumberElement)hink about impl check
- add incompatible elements while `push_back` 

(yes i know it technical debt)

test/refract/data/test-Array.cc
```diff
@@ -0,0 +1,512 @@
...
+                    }
+                    THEN("the original is not empty")
+                    {
+                        REQUIRE(!array.empty());
+                    }
+                    THEN("its size equals the original's size")
+                    {
+                        REQUIRE(array.size() == array2.size());
+                    }
+                    THEN("it is iterable")
```
klokane: DRY
It is third repetition of  **is iterable** test case with simillar implementation, what about come macro/function?

src/RefractAPI.cc
tjanc: Agreed. I've already implemented `make_element<ArrayElement>(...)` to construct from child elements, e. g. `make_element<ArrayElement>(foo(), bar())` where `foo()` and `bar` return `std::unique_ptr<IElement>`. I will modify this during cleanup.

src/RefractDataStructure.cc
tjanc: +1 will remove

src/RefractElementFactory.cc
tjanc: `make_element<ArrayElement>(...)`

src/RefractSourceMap.cc
tjanc: +1

tjanc: +1 even better:
```
return make_element<ArrayElement(
```

tjanc: +1 even better:
```
return make_element<ArrayElement(
  make_primitive(static_cast<double>(sourceMap.location)),
  make_primitive(static_cast<double>(sourceMap.length))
);
```

src/refract/Element.h
tjanc: I intended `make_primitive` to be a variant of `make_element` where the type of the element is deduced because it has a C++ primitive equivalent; e. g. `make_primitive(std::string{})`, `make_primitive("hello world")`, `make_primitive(42)` instead of `make_element<StringElement>(std::string{})`, et cetera. This construct already existed and unfortunately is used generically at some places in the code base.
Where do you see `RefractToPrimitive` in this branch?

src/refract/InfoElements.h
tjanc: I agree there is inconsistency. What about the other option, putting everything out of the header? Introducing interfaces.

src/refract/JSONSchemaVisitor.cc
tjanc: `gen_schema` to `generate`?

test/test-ApplyVisitorTest.cc
tjanc: What do you mean?

test/refract/data/test-Extend.cc
tjanc: It's technical debt, so let's shoot it down.

src/RefractSourceMap.cc
klokane: Not sure if this kind od c-tor there were while i made review :)
But agree

src/refract/Element.h
klokane: Sry mistake it is named `PrimitiveToRefract`

I understand to argue, but `make_primitive` evoke something diferent than "create element from primitive". At least for me :)

src/refract/InfoElements.h
klokane: Have not problem with move to cc. Choose one of them by your feeling :)

src/refract/JSONSchemaVisitor.cc
klokane: Still it is local variable. We have naming conventions for it

test/test-ApplyVisitorTest.cc
klokane: There are tests with body commented out by `#if 0` 
we should (sorted by prefered solution :)
- make implementation
- remove it from tests
- commentout full test not just body

But for this branche probably will be most effective just comment out related test

We do not want to have repoting functionality of test which is not implemeted

src/ElementData.h
tjanc: @kylef @pksunkara what do you think?

src/refract/Element.h
tjanc: I've looked at the code after renaming to `from_primitive`, and it reads alright.

tjanc: I can implement number from `int`, but I think we should use some big number representation that implicitly constructs from `int`, `double`, etc.

test/refract/data/test-Extend.cc
klokane: Ok, let it pass
At least, we should add issue

src/refract/JSONSchemaVisitor.cc
klokane: `generate` is ok, sorry

src/refract/Element.h
klokane: Probably best solution is hold it as string internal.
There are other problem like parsing of localized numbers, thousand separators, decimal point, transformation to different format etc.

tjanc: On second thought, I don't think it helps to bury the static_cast in a helper. I think the user of `make_primitive` should take care of providing a `double` as long as we don't provide a bignum constructor.

tjanc: Imagine `make_primitive` accepts an `int`. The user of that function has no clue there are restrictions on the `int` because it is cast into a `double` somewhere in the guts.

test/refract/data/test-String.cc
tjanc: One can't write a test like that for move, because the original object is not in defined state after it. But with the copy test I agree.

tjanc: Actually, I can't even write it for the copy. Currently, `string_t` doesn't provide an interface for changing the `std::string` it contains.

test/refract/data/test-Number.cc
tjanc: Same as `string_t`, `number_t` is immutable, so this test can't be written.

test/refract/data/test-Member.cc
tjanc: removed from implementation, wasn't used

test/refract/data/test-Array.cc
tjanc: I don't like to generalize in tests. If an assertion fails, I want it to be stated where is belongs. But I will think of a way.

src/refract/Registry.h
```diff
@@ -10,32 +10,31 @@
...
-    // Forward declarations of IElement
-    struct IElement;
-
     class Registry
     {
         // FIXME: potentionally dangerous,
         // if element is deleted and not removed from registry
         // solution: std::shared_ptr<> || std::weak_ptr<>
-        typedef std::map<std::string, IElement*> Map;
+        typedef std::map<std::string, std::unique_ptr<IElement> > Map;
```
klokane: We probably can remove comment above. Is it correct?

src/refract/Element.cc
```diff
@@ -11,424 +11,42 @@
...
-    {
-        for (MemberElement* e : elements)
-            delete e;
-    }
-
-    StringElement* IElement::Create(const char* value)
-    {
-        return Create(std::string(value));
+    constexpr std::array<const char*, 13> reserved_ = {
+        "array",   //
```
klokane: Why this empty comments?

klokane:
file `ElementInserter.h` is not used anymore, we can remove it

klokane:
From my point of view it seem ok. Just a few tiny things.

src/refract/Registry.h
tjanc: yeah this can be removed

src/refract/Element.cc
tjanc: to enforce better clang-format formatting

PR #521, opened 2017-12-19
Update tests according to enum sample/default specification changes

klokane:


src/RefractDataStructure.cc
```diff
@@ -873,30 +875,121 @@ namespace drafter
...
+            if (values.empty()) {
+                return;
+            }
+
+            auto const& first = std::get<0>(*values.begin());
+            // we are sure `first` is valid because precondition check `values.empty()`
+            if ((values.size() > 1) || (first.size() > 1)) {
+
+                mdp::CharactersRangeSet location;
+                for (auto const& item : values) {
```
tjanc: Use exclusively either `const T&` or `T const&` to increase readability. I suggest `const T&` because the const keyword is right in front of what it is bound to.

src/RefractDataStructure.cc
```diff
@@ -873,30 +875,121 @@ namespace drafter
...
+                    snowcrash::Warning("multiple definitions of 'default' value", snowcrash::MSONError, location));
+            }
+        }
+
+        template <>
+        struct LastElementToAttribute<refract::EnumElement> {
+            using T = refract::EnumElement;
+
+            template <typename U>
+            void operator()(
```
tjanc: This operator should be `const`.

test/fixtures/mson/enum-multiple-default.json
```diff
@@ -0,0 +1,315 @@
...
+                  {
+                    "element": "member",
+                    "content": {
+                      "key": {
+                        "element": "string",
+                        "content": "e3"
+                      },
+                      "value": {
+                        "element": "enum",
+                        "attributes": {
```
tjanc: Are `enumerations` missing here?

test/fixtures/schema/sample-inline-attribute.json
```diff
@@ -121,20 +121,16 @@
...
-                                                        "element": "string",
-                                                        "content": "3"
-                                                      },
-                                                      {
-                                                        "element": "string",
-                                                        "content": "4"
-                                                      }
-                                                    ]
-                                                  }
+                                                "content": {
```
tjanc: Why are `enumerations` missing now?

test/fixtures/schema/sample.json
```diff
@@ -121,20 +121,16 @@
...
-                                                        "element": "string",
-                                                        "content": "3"
-                                                      },
-                                                      {
-                                                        "element": "string",
-                                                        "content": "4"
-                                                      }
-                                                    ]
-                                                  }
+                                                "content": {
```
tjanc: Why are `enumerations` missing now?

test/fixtures/schema/sample.json
```diff
@@ -121,20 +121,16 @@
...
+                                                  "element": "string",
+                                                  "content": "3"
+                                                }
+                                              },
+                                              {
+                                                "element": "enum",
+                                                "content": {
```
tjanc: Why are `enumerations` missing now?

test/fixtures/schema/sample-inline-attribute.json
```diff
@@ -121,20 +121,16 @@
...
+                                                  "element": "string",
+                                                  "content": "3"
+                                                }
+                                              },
+                                              {
+                                                "element": "enum",
+                                                "content": {
```
tjanc: Why are `enumerations` missing now?

test/fixtures/mson/enum-multiple-default.json
kylef: There are no enumerations defined in the MSON line: `- e3: a,b,c (enum, default)`. Since there is the `default` type attribute, all a, b, and c are treated as the default value(s). Enum doesn't support multiple defaults so there is a warning annotation below: 

> multiple definitions of 'default' value

tjanc: According to the [MSON Spec](https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#212-structure-types) an `enum`
> Specifies an exclusive list of possible Values or Types for values.

From my understanding, this list is what we put into our `enumerations` attribute.

So if the value of the member element in `- e3: a,b,c (enum, default)` is parsed as an MSON enum, it is _defined_ by an exclusive list of its possible MSON Values or MSON Types. In this case, I noticed we don't error out, so I expect this list is well defined. What is it in this case? `[a,b,c]`? `[]`?

The idea of leaving out the `enumerations` attribute is limited to `samples` and `default` _inside_ the definition of an enum element, because the `enumerations` of a value always match the `enumerations` of its type.

PR #525, opened 2018-01-24
Sum type for JSON and YAML simple object notations

tjanc:
- `std::variant`-like tagged union type with unit tests
- added `boost::containers`, enabling recursive data structures and opening the door to `boost::pmr`
- definitions and aliases for JSON/YAML values
- outline for JSON serialization


test/utils/test-Variant.cc
```diff
@@ -0,0 +1,895 @@
...
+            {
+                REQUIRE(Foo::record().copy_constructor == 0);
+                REQUIRE(Foo::record().copy_assignment == 0);
+            }
+
+            THEN("it was not moved")
+            {
+                REQUIRE(Foo::record().copy_constructor == 0);
+                REQUIRE(Foo::record().copy_assignment == 0);
+            }
```
pksunkara: I don't see the above 2 `THEN` anywhere else in similar tests below. Did you forget to remove these?

test/utils/test-Variant.cc
```diff
@@ -0,0 +1,895 @@
...
+        Bar::reset_record();
+        Baz::reset_record();
+
+        tested v(std::move(bar));
+
+        THEN("one Bar and one Foo exist")
+        {
+            REQUIRE(Foo::instances().size() == 0);
+            REQUIRE(Bar::instances().size() == 2);
+            REQUIRE(Baz::instances().size() == 0);
```
pksunkara: The tests and the message is contradicting.

test/utils/test-Variant.cc
tjanc: 👍 actually `move_contructor` and `move_assignment` should be tested there

test/utils/test-Variant.cc
tjanc: :+1 wrongly copy-pasted message

test/utils/test-Variant.cc
pksunkara: Shouldn't you be testing `move_assignment` and `copy_assignment` in other tests below too?

tjanc: 👍 added assertions where missing

klokane:
What is source `test/fixtures/utf-8` there is missing license or something like this.

klokane:
Missing tests for variant `visit`, `emplace`

klokane:
Do we need `boost` libs at this time?

klokane:
Missing tests for src/utils/Utils.h

tjanc:
- `test/fixtures/utf-8` is not added as part of this PR but #527 ; however, it's just a sequence of all utf-8 characters, so I don't really think licensing is an issue
- tests for visit are missing and should definitely be included; emplace is already tested by testing `operator=` but I can copy-paste the tests
- we need boost to represent recursive data structures; boost containers specifically allow us to do that; that's also in the description of this PR
- `static_assert`s for `src/utils/Utils.h` are missing and should be included

tjanc:
I will fix this in a separate PR, ok?

klokane:
It is ok fix it other PR.
About `test/fixtures/utf-8` I mean if it come from someone other (you did not create it by yourself) then you will need some kind of license

PR #526, opened 2018-02-01
Klokane/enum samples on master

klokane:
rebase of #521 on master

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -841,6 +841,27 @@ namespace
         element.attributes().set(key, std::move(a));
     }
 
+    template <>
+    void AllElementsToAtribute<refract::EnumElement, ElementInfoContainer<refract::EnumElement> >(
+        ElementInfoContainer<refract::EnumElement> values, const std::string& key, IElement& element)
+    {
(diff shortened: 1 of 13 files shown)


pksunkara:
Do you want to close the #521?

src/RefractDataStructure.cc
```diff
@@ -853,6 +874,49 @@ namespace
...
+        }
+
+        auto info = Merge<T>()(std::move(values));
+
+        if (info.value.empty()) {
+            element.attributes().set(key, make_empty<EnumElement>());
+            return;
+        }
+
+        element.attributes().set(key, make_element<EnumElement>(dsd::Enum{ std::move(info.value.back()) }));
```
tjanc: Should we pop the last element after the move? Its value is `nullptr`.

klokane: We can, but in this case rest is dropped, so we can omit pop in this case
It is same like nonspecialized case. see:
https://github.com/apiaryio/drafter/blob/ce0edbd63e9ef3e8d9249929d7031666cba74b24/src/RefractDataStructure.cc#L874

klokane:
Changelog updated

PR #528, opened 2018-02-15
(util) simple object YAML serialization

tjanc:
Implements YAML serialization according to: http://www.yaml.org/spec/1.2/spec.html

Utf-8 fixtures were generated using the implementation and diff`ed against PyYAML and Js-YAML. Both unfortunately escape some printable symbols, which SHOULD not happen; that's why the result of string escaping in this implementation differs syntactically from PyYAML and Js-YAML. However, the implementation doesn't differ semantically.

Note that the implementation comes with a reasonably well tested standalone utf-8 input iterator implementation.


test/utils/test-Utf8.cc
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+    }
+
+    // TODO @tjanc@ test ending prematurely 3-byte
+    // TODO @tjanc@ test ending prematurely 4-byte
+    // TODO @tjanc@ test oversized 2-byte
+    // TODO @tjanc@ test oversized 3-byte
+    // TODO @tjanc@ test oversized 4-byte
+    // TODO @tjanc@ test surrogates as codepoints (3-byte)
```
pksunkara: What are these about?

pksunkara: Does the PR need to wait for them?

src/utils/so/YamlIo.cc
```diff
@@ -9,13 +9,257 @@
...
+
+            serialize_yaml(out, value.data);
+        }
+
+        void operator()(const Number& value, std::ostream& out, int indent = 0) const
+        {
+            if (indent > 0)
+                out << ' ';
+
+            out << value.data; // TODO @tjanc@ probably not correct
```
pksunkara: What does the YAML spec say? JSON needed the scientific stuff IIRC

tjanc: From what I gathered YAML accepts default C++ `double` serialization. Will remove TODO.

test/utils/test-Utf8.cc
tjanc: This SHOULD be implemented to make the tests, although such a condition CANNOT occur as long as the parser implements utf-8 parsing correctly. Because that's not the case, I shall implement those tests. It's an oversight on my side, I should have grep'ed for `TODO`s in my source code.

pksunkara:
I need this merged now, so @tjanc please make a new PR removing that TODO comment.

PR #531, opened 2018-02-21
(utils) yaml serialization review addressed

tjanc:
- implemented missing test
- removed irrelevant TODO

- enriched comments a bit
- minor renaming

### DIFF
--- a/src/utils/Utf8.h
+++ b/src/utils/Utf8.h
@@ -30,72 +30,72 @@ namespace drafter
             ///
             /// Decode utf8 codepoint from octet sequence
             ///
-            /// @param  b   iterator to begin of octet sequence to be decoded from
-            /// @param  e   iterator to end of octet sequence to be decoded from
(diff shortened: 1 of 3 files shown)


src/utils/Utf8.h
```diff
@@ -20,11 +21,95 @@ namespace drafter
...
+
             using codepoint = std::uint32_t;
 
+            constexpr codepoint replacement = 0xFFFD;
+            constexpr codepoint invalid = 0xFFFFFFFF;
+
+            ///
+            /// Decode utf8 codepoint from octet sequence
+            ///
+            /// @param  b   iterator to begin of octet sequence to be decoded from
```
kylef: Perhaps it would be nicer to call these variables a more explanatory name? `begin` and `end` vs `b` and `e`?

tjanc: Will rename to `first` and `last`.

PR #532, opened 2018-03-02
Enumeration implementation

klokane:


klokane:
I personally do not like differences between:
```
 e2.1: 1,2,3 (enum) - E:[1,2,3]                                                                                                                                                                                                    
                                                                                                                                                                                                                                    
- e2.2 (enum) - E: [1,2,3]
    - 1
    - 2
   -  3
```
and 
```
- e1.1: 1 (enum) - C: 1, E: [1]

- e1.2 (enum) - E: [1]
    - 1
```

It is same definition (it is different in "one" vs "many") but it leads into diferent result.

@apiaryio/adt what do you mean about it?

tjanc:
I think it makes _most_ sense like this, although it's not perfect.
How would you like the enum listed in-line to behave? What samples/default/content would you like to see for `- e2.1: 1,2,3 (enum)`?

It's a feature by exception. We can deprecate and purge it from existence. Nevertheless, I see value in the inline notation for being compact.

Personally, I parse it as an enum defined by the inline list, a pure type definition without content/default/samples.

klokane:
I'm ok with `- e2.1: 1,2,3 (enum) - E:[1,2,3]`.
I personaly prefer change behavior for one value.
possible options:
```
- e1.1: 1 (enum) - C: 1, E: [1]

- e1.2 (enum) - C:1, E: [1]
    - 1
```

or 

```
- e1.1: 1 (enum) - E: [1]

- e1.2 (enum) - E: [1]
    - 1
```

But second one will leave idea of `content` for enums. There not will be way to put anything into content

tjanc:
Your first suggestion is intriguing. The only thing coming to mind is this:
```
- e1.1: 1 (enum) - C: 1, E: [1]

- e1.2 (enum) - C:1, E: [1]
    - 1

- e1.3 (enum) - C:?, E: [1,2,3]
    - 1
    - 2
    - 3

- e1.4: 2 (enum) - C:2, E: [1,2,3]
    - 1
    - 2
    - 3
```
As you can see above, extrapolating `e1.2` to `e1.3` by just adding more entries to the sum type will prove inconsistent anyway.

klokane:
By extending origin idea "first value is winner", I mean `e1.3` should leads to 
```
- e1.3 (enum) - C:1, E: [1,2,3]
    - 1
    - 2
    - 3
```

klokane:
with `e1.4` i agree. First is `2` so content will be `2` it seems legit

tjanc:
I think we reached this point:



We feel like we can't leave it like this because we are programmers. But the exception of specifying the enum entry list in place of content **is** an inconsistency; by specification.
The "misplaced" block is really just syntax sugar. It's best to think of `e: a,b,c,... (enum)` in the same way as
```
- e (enum)
    - a
    - b
    - c
    - ...
```
Can we agree on that?

klokane:
Yes, we can. That is point which I mean as starting :)

tjanc:
Ah ok I didn't get that. In that case, I'd suggest
```
- e (enum) - E[1,2,3]
    - 1
    - 2
    - 3
```
Just because I want to be able to specify content-less (type only) enums with MSON. Furthermore,
```
- e: 2 (enum) - C: 2; E[1,2,3]
    - 1
    - 2
    - 3
```
And for completeness,
```
- e: 1,2,3 (enum) - E[1,2,3]
```
What's your thinking?

klokane:
It is according to this PR implementation.
From my point of view there is inconsistency for one element `e1.1` vs `e1.2` but if all agree with this behavior I will accept it.

klokane:
@apiaryio/adt ready for review

test/fixtures/api/attributes-named-type-enum-reference.json
```diff
@@ -227,7 +238,7 @@
                                   "content": "application/schema+json"
                                 }
                               },
-                              "content": "{\n  \"$schema\": \"http://json-schema.org/draft-04/schema#\",\n  \"type\": \"object\",\n  \"properties\": {\n    \"a\": {\n      \"type\": \"string\",\n      \"enum\": [\n        \"valueB\",\n        \"valueA\"\n      ]\n    }\n  },\n  \"required\": [\n    \"a\"\n  ]\n}"
+                              "content": "{\n  \"$schema\": \"http://json-schema.org/draft-04/schema#\",\n  \"type\": \"object\",\n  \"properties\": {\n    \"a\": {\n      \"type\": \"string\",\n      \"enum\": [\n        \"valueB\",\n        \"valueA\",\n        \"valueA\"\n      ]\n    }\n  },\n  \"required\": [\n    \"a\"\n  ]\n}"
```
kylef: Is this a regression? the JSON Schema enum now contains duplicate `valueA`, perhaps this is going to be solved by @tjanc work on the JSON Schema visitor?

src/ElementInfoUtils.h
```diff
@@ -0,0 +1,74 @@
...
+
+    template <typename T, bool IsPrimitive = is_primitive<T>()>
+    struct Merge;
+
+    template <typename T>
+    struct Merge<T, true> {
+        using Info = ElementInfo<T>;
+        using Container = std::deque<Info>;
+
+        ElementInfo<T> operator()(Container container) const
```
pksunkara: You have defined `Info` above. Why not use here and below?

test/fixtures/mson/enum-sample.json
```diff
@@ -197,6 +223,26 @@
...
+                                "element": "number",
+                                "content": 5
+                              },
+                              {
+                                "element": "number",
+                                "content": 6
+                              },
+                              {
+                                "element": "number",
+                                "content": 1
```
pksunkara: The `e7` in this file is wrong. https://github.com/apiaryio/drafter/pull/532/files#diff-882a86245349acaecd3ba383d5503251R350

test/fixtures/mson/enum-variants.json
```diff
@@ -128,26 +139,22 @@
...
                               {
                                 "element": "string",
                                 "content": "XL"
                               }
                             ]
                           }
-                        },
-                        "content": {
-                          "element": "string",
-                          "content": "L"
```
pksunkara: This shouldn't be removed.

test/fixtures/mson/enum-variants.json
```diff
@@ -194,10 +205,6 @@
...
```
pksunkara: This shouldn't be removed.

test/fixtures/mson/enum-variants.json
```diff
@@ -352,10 +385,6 @@
                               }
                             ]
                           }
-                        },
-                        "content": {
-                          "element": "string",
-                          "content": "S"
```
pksunkara: This shouldn't be removed.

test/fixtures/mson/enum-variants.json
```diff
@@ -704,10 +777,6 @@
                               "content": "K"
                             }
                           }
-                        },
-                        "content": {
-                          "element": "string",
-                          "content": "A"
```
pksunkara: This shouldn't be removed.

test/fixtures/mson/enumerations.json
```diff
@@ -0,0 +1,372 @@
...
+                            "element": "array",
+                            "content": [
+                              {
+                                "element": "string",
+                                "content": "1"
+                              }
+                            ]
+                          },
+                          "default": {
+                            "element": "enum"
```
pksunkara: Missing `1` here.

test/test-ElementComparator.cc
```diff
@@ -0,0 +1,195 @@
...
+
+    GIVEN("Two empty primitive elements of diferent type")
+    {
+
+        auto first = make_empty<S>();
+        auto second = make_empty<N>();
+
+        WHEN("it is compared")
+        {
+            THEN("it is recognized as equal")
```
pksunkara: it is **not** recognized as equal

test/fixtures/api/attributes-named-type-enum-reference.json
klokane: Yes it probably need to be resolved in JSON schema visitor. I did not to want to touch it because Thomas is working on.

src/ElementInfoUtils.h
klokane: It is header, I want to avoid to dirty global space. Local naming is better in this case

test/fixtures/mson/enum-sample.json
klokane: Due to there is no change against older code let make separated issue.

klokane:
@pksunkara updated

test/fixtures/api/attributes-named-type-enum-reference.json
tjanc: This must be fixed earlier on, it's a regression.

pksunkara:
There are still [this comment](https://github.com/apiaryio/drafter/pull/532#discussion_r172560064) that needs to be addressed.

test/fixtures/mson/enumerations.json
klokane: It is wrongly indented in source. sry.

klokane:
pls do not merge, working on fix json schema

src/RefractDataStructure.cc
```diff
@@ -801,23 +730,47 @@ namespace
...
-            if (!enumInfo.value.empty()) {
-                auto enumsElement = make_element<ArrayElement>();
-                std::move(enumInfo.value.begin(), enumInfo.value.end(), std::back_inserter(enumsElement->get()));
+            auto valuesInfo = Merge<EnumElement>()(std::move(data.values));
+            auto samplesInfo = Merge<EnumElement>()(std::move(CloneElementInfoContainer<T>(data.samples)));
+            auto defaultInfo = Merge<EnumElement>()(std::move(CloneElementInfoContainer<T>(data.defaults)));
+
+            dsd::Array enums;
+
+            auto addToEnumerations = [&enums, &context](auto& info, const auto& sourceMap, const bool reportDuplicity) {
```
tjanc: Why does this lambda capture references instead of taking them as argument? Is there a reason to be un-pure?

src/RefractDataStructure.cc
```diff
@@ -801,23 +730,47 @@ namespace
...
+                    enums.push_back(std::move(info));
+                } else if (reportDuplicity) {
+                    context.warn(
+                        snowcrash::Warning("duplicit value in enumeration", snowcrash::MSONError, sourceMap.sourceMap));
+                }
+            };
+
+            std::for_each(valuesInfo.value.begin(),
+                valuesInfo.value.end(),
+                std::bind(addToEnumerations, std::placeholders::_1, valuesInfo.sourceMap, true));
```
tjanc: Prefer primitive lambdas with captures to `std::bind`.

src/ElementInfoUtils.h
```diff
@@ -0,0 +1,74 @@
+//
+//  MergeElementInfo.h
```
tjanc: File name doesn't match.

test/test-ElementComparator.cc
```diff
@@ -0,0 +1,195 @@
+#include "ElementComparator.h"
+#include "catch.hpp"
+#include "refract/Element.h"
+
+using namespace drafter;
+using S = refract::StringElement;
+using N = refract::NumberElement;
+
+SCENARIO("Compare equality of elements", "[Element][comparate][equal]")
```
tjanc: `comparate` -> `compare` || `comparator`

src/ElementComparator.h
```diff
@@ -0,0 +1,84 @@
...
+                        auto key = info.first;
+                        if (ignore.find(key) != ignore.end()) {
+                            return false;
+                        }
+
+                        if (keys.find(key) != ignore.end()) {
+                            return false;
+                        }
+
+                        return true;
```
tjanc: Is this covered in unit tests?

src/ElementComparator.h
```diff
@@ -0,0 +1,84 @@
...
+                    })
+                != rhs.end()) {
+                return false;
```
tjanc: Same as above. `gcov` excerpt:
```
     2448:   45:            if (std::find_if(rhs.begin(),
        -:   46:                    rhs.end(),
      846:   47:                    [&keys, &lhs, &ignore](const auto& info) {
      560:   48:                        auto key = info.first;
      560:   49:                        if (ignore.find(key) != ignore.end()) {
      277:   50:                            return false;
        -:   51:                        }
        -:   52:
        6:   53:                        if (keys.find(key) != ignore.end()) {
        3:   54:                            return false;
        -:   55:                        }
        -:   56:
    #####:   57:                        return true;
      816:   58:                    })
     2448:   59:                != rhs.end()) {
    #####:   60:                return false;
        -:   61:            }

```

src/ElementComparator.h
```diff
@@ -0,0 +1,84 @@
...
+#include <algorithm>
+
+namespace drafter
+{
+    using namespace refract;
+    struct InfoElementsComparator {
+        bool operator()(const InfoElements& rhs, const InfoElements& lhs)
+        {
+
+            const std::set<std::string> ignore = { "sourceMap" };
```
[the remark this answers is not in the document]
klokane: Actually it is useless, but idea is allow inject configurable set of "keys" to ignore while compare

src/ElementInfoUtils.h
```diff
@@ -0,0 +1,74 @@
...
+        Info operator()(Container container) const
+        {
+            Info result{};
+
+            for (auto& info : container) {
+                std::move(info.value.begin(), info.value.end(), std::back_inserter(result.value));
+                result.sourceMap.sourceMap.append(info.sourceMap.sourceMap);
+            }
+
+            return std::move(result);
```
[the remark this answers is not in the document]
klokane: I just add line number 50 and move it into separated file.
into return move semantic it was switched by you.
See https://github.com/apiaryio/drafter/blame/master/src/RefractDataStructure.cc#L716

Or do you mean by adding `result.sourceMap.sourceMap.append(info.sourceMap.sourceMap);`
It is now good adept to move?

src/RefractDataStructure.cc
```diff
@@ -787,9 +716,9 @@ namespace
 
     template <typename T>
     struct SaveValue<T, false> {
-        void operator()(std::deque<ElementInfo<T> > values, std::deque<ElementInfo<T> > /* enumerations*/, T& element) const
+        void operator()(ElementData<T>& data, T& element, ConversionContext& /* context */) const
         {
-            auto info = Merge<T>()(std::move(values));
+            auto info = Merge<T>()(std::move(data.values));
```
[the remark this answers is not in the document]
klokane: Will look into

src/RefractDataStructure.cc
klokane: They are hold full time, all others param are changed, but i can move it into params

src/RefractDataStructure.cc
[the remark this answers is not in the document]
klokane: I mean there is possible way be leave values untouched and clone it.
In `SaveValue` for enumeration it is required access to other `data.samples` and `data.defalut` so we need to use full `ElementData`

src/ElementComparator.h
```diff
@@ -0,0 +1,84 @@
...
+            std::set<std::string> keys;
+
+            if (std::find_if(lhs.begin(),
+                    lhs.end(),
+                    [&keys, &rhs, &ignore](const auto& info) {
+                        auto key = info.first;
+                        if (ignore.find(key) != ignore.end()) {
+                            return false;
+                        }
+                        keys.insert(key);
```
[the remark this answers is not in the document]
klokane: We can not compare ranges. For example, first element can contain  `sourceMap` while second one not. For this case we have no same count of attrs/meta while elements can still be recognized as equal.

For this reason we compare key by key. And while comparation happen we save set of already compred `keys` into set, to avoid compare them again on second element.

So in general I do not modify stack, just save some temporary info.

test/fixtures/mson/enum-multiple-default.json
```diff
@@ -54,6 +54,19 @@
...
+                          "enumerations": {
+                            "element": "array",
+                            "content": [
+                              {
+                                "element": "string",
+                                "content": "J"
+                              },
+                              {
+                                "element": "string",
+                                "content": "K"
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-multiple-default.json
```diff
@@ -75,6 +88,19 @@
...
+                              }
+                            ]
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-multiple-default.json
```diff
@@ -96,6 +122,23 @@
...
+                              },
+                              {
+                                "element": "string",
+                                "content": "b"
+                              },
+                              {
+                                "element": "string",
+                                "content": "c"
+                              }
+                            ]
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-sample.json
```diff
@@ -81,6 +81,19 @@
...
+                            "content": [
+                              {
+                                "element": "string",
+                                "content": "3"
+                              },
+                              {
+                                "element": "string",
+                                "content": "4"
+                              }
+                            ]
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-sample.json
```diff
@@ -153,6 +166,19 @@
...
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -59,6 +59,17 @@
...
                         "element": "enum",
+                        "attributes": {
+                          "enumerations": {
+                            "element": "array",
+                            "content": [
+                              {
+                                "element": "string",
+                                "content": "L"
+                              }
+                            ]
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -128,15 +139,15 @@
...
                               },
                               {
                                 "element": "string",
-                                "content": "M"
+                                "content": "S"
                               },
                               {
                                 "element": "string",
-                                "content": "L"
+                                "content": "M"
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -218,6 +233,15 @@
...
                         "element": "enum",
                         "attributes": {
+                          "enumerations": {
+                            "element": "array",
+                            "content": [
+                              {
+                                "element": "string",
+                                "content": "L"
+                              }
+                            ]
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -250,6 +274,15 @@
...
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -334,9 +367,17 @@
...
+                                "element": "string",
+                                "content": "S"
+                              },
                               {
                                 "element": "string",
                                 "content": "T"
+                              },
+                              {
+                                "element": "string",
+                                "content": "U"
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

test/fixtures/mson/enum-variants.json
```diff
@@ -386,6 +427,10 @@
                               {
                                 "element": "string",
                                 "content": "N"
+                              },
+                              {
+                                "element": "string",
+                                "content": "L"
```
tjanc: This matches any `string`. Are you missing `fixed` attribute on the `string` elements above?

klokane:
@apiaryio/adt guys there are no responses on my comments it this PR. Can you please add reactions on?

klokane:
I'm not sure if it is good idea to mark it as fixed.
what about
```
- e (enum)
  - 1 (fixed)
  - 2
```

Once added attribute `fixed` you are not able to remove

test/fixtures/mson/enum-multiple-default.json
klokane: see general comment https://github.com/apiaryio/drafter/pull/532#issuecomment-374215268

src/ElementComparator.h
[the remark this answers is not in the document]
tjanc: `keys` is living on the stack and you are calling `insert` on it, which is non-`const`. That's why the lambda modifies the stack. But that's actually not the point.

Quoting cppreference, `find_if` suggests you are searching for something:
```
find finds the first element satisfying specific criteria
```
Something described by the **predicate** you fed it, specifically:
```
find_if searches for an element for which predicate p returns true
```
That's how every other C++ programmer will try to read this code. But the lambda you feed `find_if` is nothing like a predicate, and is modifying values captured from the stack. Yes it works, I'm just saying, it's not easy to read because it abuses find_if.

src/ElementInfoUtils.h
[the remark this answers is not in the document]
tjanc: Yep, seems like I introduced that while modifying ownership semantics with that commit. Can you please fix it anyway? I think it even gives a warning in LLVM.

src/ElementComparator.h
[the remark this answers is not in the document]
tjanc: What about a custom iterator skipping `sourceMap` entries to make it more readable?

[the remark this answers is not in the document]
klokane: If you look inside, it is predicate. 
1. look if it is in ignored 
2. look if there is same key in rhs
3. check if given(lhs) and found(rhs) values are same.

I save already comparated keys to avoid check them in rhs.

Similar algo is used to check all rhs.

But ok. I rewrite it

src/RefractDataStructure.cc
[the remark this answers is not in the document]
klokane: Another optional solution is send "can move" hinting. Some helper structure like
```
struct CanMove {
  bool values;
  bool samples;
  bool defaults;
}
//...
// calling
SaveValue<T>()(data, { true, false, false }m element, context);

// and then inside of SaveValue
auto valuesInfo = canMove.values
    : Merge<EnumElement>()(std::move(data.values))
    ? Merge<EnumElement>()(std::move(CloneElementInfoContainer<T>(data.values))
```

src/ElementComparator.h
klokane: Non relevant according to change algorihm

src/RefractDataStructure.cc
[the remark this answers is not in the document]
tjanc: Yeah, let's avoid more complexity and leave it as is.

src/refract/VisitorUtils.cc
```diff
@@ -66,3 +66,22 @@ MemberElement* refract::FindMemberByKey(const ObjectElement& e, const std::strin
...
+    switch (elementType) {
+        case TypeQueryVisitor::String:
+        case TypeQueryVisitor::Number:
+        case TypeQueryVisitor::Boolean:
+            if (!e.empty())
+                return true;
+        default:
+            return HasTypeAttribute(e, "fixed");
+            ;
+    };
```
tjanc: Please make an if condition out of this switch.

src/refract/InfoElementsUtils.h
```diff
@@ -0,0 +1,51 @@
...
+
+#include "InfoElements.h"
+#include "TypeQueryVisitor.h"
+#include "Element.h"
+
+namespace refract
+{
+
+    template <typename ValueElementType, typename DSDType>
+    void AppendInfoElement(InfoElements& ie, const std::string& key, DSDType&& value)
```
tjanc: I find it very hard to see what this function is doing by looking at it's name, signature and template arguments. Please either add a full doxygen-style comment, or modify the name/signature significantly.

src/refract/VisitorUtils.cc
tjanc: Also, what about `NullElement`?

tjanc: I mean, the function is called `IsLiteral`. I think `IsLiteral` should behave the same for all four primitive Elements.
There is no such concept as `fixed` on `Null Element`, because there is just one value in the Null universe. For that reason I'd just always use `{"element":"null"}` for Null Element specifically.

klokane: Oh you mean, null should not be fixed under any condition. Agree. Will fix

klokane:
There is something wrong with **continuous-integration/appveyor/pr** with message *Expected — Waiting for status to be reported*
I tried to invoke rebuild on appveyor but wthout success.

klokane:
After rebase on current master it was build successfully

PR #533, opened 2018-03-06
Allow resource to be directly nested under the api element

pksunkara:


pksunkara:
One of you need to check the 2nd commit closely. It had almost rewritten test fixtures. I could have added a resource group.

pksunkara:
I added resource group to most of the fixtures because otherwise the whole JSON output of a fixture will change and it lead to a lot of noise in the commit.

src/utils/so/JsonIo.cc
```diff
@@ -85,7 +85,7 @@ namespace
                     break;
                 default: {
                     // unsigned int representation
-                    std::uint8_t c = byte < 0x80 ? byte : byte + 0x80;
+                    std::uint8_t c = static_cast<std::uint8_t>(byte);
```
klokane: Why this?

klokane:
All other seems ok

src/utils/so/JsonIo.cc
pksunkara: It's not compiling on mac without the changes here.

PR #536, opened 2018-03-16
(json-schema) primitive types, array, enum

tjanc:
Implements a subset of JSON schema generation from API Elements,
  + NullElement,
  + BooleanElement,
  + StringElement,
  + ArrayElement,
  + EnumElement,

without any support for named types.

Note that `fixed` and `fixedType` apply to some of the above elements. `optional`, `required` and `nullable` do not. `samples` and `default` are generally irrelevant when generating schemas.

This PR does not change production code paths.

### DIFF
--- a/src/refract/JsonSchema.cc
+++ b/src/refract/JsonSchema.cc
@@ -13,9 +13,8 @@
 #include "Element.h"
 #include "../ElementData.h"
 #include "Utils.h"
+#include "VisitorUtils.h"
 #include "../utils/log/Trivial.h"
-
-#include "VisitorUtils.h" // GetValue
 
 using namespace refract;
 using namespace schema;
@@ -24,19 +23,19 @@ using namespace drafter::utils::log;
 
 namespace
 { // API Elements tools
-    so::String instanciate(const StringElement& e)
+    so::String instantiate(const StringElement& e)
     {
         assert(!e.empty());
         return so::String{ e.get().get() };
     }
 
-    so::Number instanciate(const NumberElement& e)
+    so::Number instantiate(const NumberElement& e)
     {
         assert(!e.empty());
         return so::Number{ e.get().get() };
     }
 
-    so::Value instanciate(const BooleanElement& e)
+    so::Value instantiate(const BooleanElement& e)
     {
         assert(!e.empty());
         return e.get() ? //
@@ -44,26 +43,17 @@ namespace
             so::Value{ in_place_type<so::False>{} };
     }
 
-    bool has_fixed_attr(const IElement& e)
+    bool hasFixedAttr(const IElement& e)
     {
         return e.attributes().end() != e.attributes().find("fixed");
     }
 
-    bool has_fixed_type_attr(const IElement& e)
+    bool hasFixedTypeAttr(const IElement& e)
     {
         return e.attributes().end() != e.attributes().find("fixedType");
     }
 
-    template <typename T>
-    const T* get_if_default_attr(const T& e)
-    {
-        auto it = e.attributes().find("default");
-        if (e.attributes().end() != it)
-            return TypeQueryVisitor::as<const T>(it->second.get());
-        return nullptr;
-    }
-
-    const dsd::Array& get_enumerations(const EnumElement& e)
+    const dsd::Array& getEnumerations(const EnumElement& e)
     {
         auto resultIt = e.attributes().find("enumerations");
         assert(e.attributes().end() != resultIt);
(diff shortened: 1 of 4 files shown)


src/refract/JsonSchema.cc
```diff
@@ -0,0 +1,290 @@
...
+{
+    using flags = std::bitset<1>;
+    constexpr std::size_t fixed_flag = 0;
+}
+
+namespace
+{ // JSON Schema tools
+    so::Object& add_schema_version(so::Object& schema)
+    {
+        schema.data.emplace_back("$schema", so::String{ "http://json-schema.org/draft-04/schema#" });
```
kylef: 💭 We should start thinking about JSON Schema draft 7 😬. Since we don't delve into validations and other features of JSON Schema, perhaps our generated schemas would already be compatible.

tjanc: Agreed. We could do that after `JsonSchema.cc` is integrated and changes in fixtures are reviewed, to keep diffs as sane as possible.
:thought_balloon:  A commit in the future changing both the URI to `draft-06` and `"enum":[foo]` to `"const": foo`.

src/refract/JsonSchema.cc
```diff
@@ -0,0 +1,290 @@
...
+
+        } else if (options.test(fixed_flag)) { // tuple of constants/types
+
+            if (e.empty() || e.get().empty())
+                add_enum(schema, so::Array{});
+            else {
+                so::Array items{};
+                for (const auto& item : e.get())
+                    items.data.emplace_back(make_schema(*item, options));
+                add_items(schema, std::move(items));
```
tjanc: Missing `"minItems":<N>' and `"additionalItems":false` here. Will be fixed in following PR.

src/refract/JsonSchema.cc
```diff
@@ -0,0 +1,290 @@
...
+
+    so::Value instanciate(const BooleanElement& e)
+    {
+        assert(!e.empty());
+        return e.get() ? //
+            so::Value{ in_place_type<so::True>{} } :
+            so::Value{ in_place_type<so::False>{} };
+    }
+
+    bool has_fixed_attr(const IElement& e)
```
klokane: why snake_naming? Full project with few exceptions used cameCase

src/refract/JsonSchema.cc
```diff
@@ -0,0 +1,290 @@
...
+#include "VisitorUtils.h" // GetValue
+
+using namespace refract;
+using namespace schema;
+using namespace drafter::utils;
+using namespace drafter::utils::log;
+
+namespace
+{ // API Elements tools
+    so::String instanciate(const StringElement& e)
```
klokane: instanciate - misspell?

test/refract/test-JsonSchema.cc
```diff
@@ -0,0 +1,706 @@
...
+            auto result = schema::generateJsonSchema(*el);
+
+            THEN("the schema matches any boolean")
+            {
+                REQUIRE(to_string(result) == R"({"$schema":"http://json-schema.org/draft-04/schema#","type":"null"})");
+            }
+        }
+    }
+
+    GIVEN("A BooleanElement with content")
```
klokane: BooleanElement? There is NullElement

test/refract/test-JsonSchema.cc
```diff
@@ -0,0 +1,706 @@
...
+            auto result = schema::generateJsonSchema(*el);
+
+            THEN("the schema matches arrays of strings")
+            {
+                REQUIRE(to_string(result) == R"({"$schema":"http://json-schema.org/draft-04/schema#","type":"array","items":{"type":"string"}})");
+            }
+        }
+    }
+
+    GIVEN("An ArrayElement with fixed string as content")
```
klokane: Fixed string?

src/refract/JsonSchema.cc
tjanc: How is that a misspell?

test/refract/test-JsonSchema.cc
tjanc: I have this fixed in tjanc/schema-definitions. Is it enough to fix it in the next PR?

test/refract/test-JsonSchema.cc
tjanc: A string with fixed attribute true. Should I be more verbose?

src/refract/JsonSchema.cc
tjanc: Because I find it way more readable, even though it is different from the rest of the project. Also, it's not a public interface.
Should I change this in the whole commit?

klokane: my personal preference is too snake_naming, but projects is in camelCase. 
So I mean we should continue with. Yes in whole commit.

src/refract/JsonSchema.cc
klokane: Not sure, I just asking. I mean it should be "instantiate"

test/refract/test-JsonSchema.cc
klokane: ok

test/refract/test-JsonSchema.cc
klokane: I'm not sure what do you mean by "fixed string".
It seems like "normal" StringElement. I do not see fixed there.
e.g. for NumberElement with content you do not use "fixed number"

Maybe i just wrong understand to meaning

src/refract/JsonSchema.cc
tjanc: https://en.wiktionary.org/wiki/instanciate
You're right.

test/refract/test-JsonSchema.cc
tjanc: No, you're right. I mistakenly associated your comment with the expressions on 346-347.

PR #537, opened 2018-03-20
(json-schema) refs & named types in `definitions` 

tjanc:
+ schemas of named types render into the `definitions` property
+ schemas of named types are referenced via `$ref` properties
+ array element with fixed attributed renders minItems and
additionalItems
+ minor fixes (spelling, unused typedefs etc.)

### DIFF
--- a/src/refract/JsonSchema.cc
+++ b/src/refract/JsonSchema.cc
@@ -15,6 +15,7 @@
 #include "Element.h"
 #include <bitset>
 #include <cassert>
+#include <algorithm>
 
 using namespace refract;
 using namespace schema;
@@ -43,14 +44,30 @@ namespace
             so::Value{ in_place_type<so::False>{} };
     }
 
+    bool hasTypeAttr(const IElement& e, const char* name)
+    {
+        auto typeAttrIt = e.attributes().find("typeAttributes");
+
+        if (typeAttrIt != e.attributes().end())
+            if (const auto* typeAttrs = TypeQueryVisitor::as<const ArrayElement>(typeAttrIt->second.get())) {
+                const auto b = typeAttrs->get().begin();
+                const auto e = typeAttrs->get().end();
+                return e != std::find_if(b, e, [&name](const auto& el) { //
+                    const auto* entry = TypeQueryVisitor::as<const StringElement>(el.get());
+                    return entry && !entry->empty() && (entry->get().get() == name);
+                });
+            }
+        return false;
+    }
+
     bool hasFixedAttr(const IElement& e)
     {
-        return e.attributes().end() != e.attributes().find("fixed");
+        return hasTypeAttr(e, "fixed");
     }
 
     bool hasFixedTypeAttr(const IElement& e)
     {
-        return e.attributes().end() != e.attributes().find("fixedType");
+        return hasTypeAttr(e, "fixedType");
     }
 
     const dsd::Array& getEnumerations(const EnumElement& e)
--- a/src/refract/JsonSchema.cc
+++ b/src/refract/JsonSchema.cc
@@ -8,13 +8,13 @@
 
 #include "JsonSchema.h"
 
-#include <cassert>
-#include <bitset>
-#include "Element.h"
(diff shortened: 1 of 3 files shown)


src/refract/JsonSchema.cc
```diff
@@ -108,93 +99,152 @@ namespace
...
+    so::Object make_schema(const IElement& e, so::Object& defs, flags options = 0);
+
+    so::Object& render_schema(so::Object& schema, so::Object& defs, const ObjectElement& e, flags options)
     {
         assert(false); // TODO @tjanc@ not implemented
         return schema;
     }
 
-    so::Object& render_schema(so::Object& schema, const ArrayElement& e, flags options)
+    so::Object& render_schema(so::Object& s, so::Object& defs, const ArrayElement& e, flags options)
```
kylef: Why change from a nice readable name like schema to s?

test/refract/test-JsonSchema.cc
```diff
@@ -194,6 +227,23 @@ SCENARIO("JSON Schema serialization of NumberElement", "[json-schema]")
             }
         }
     }
+
+    GIVEN("A named NumberElement with content and fixed attribute true")
+    {
+        auto el = from_primitive(42.0);
+        el->element("Baz");
+        el->attributes().set("fixed", from_primitive(true));
```
kylef: Shouldn’t fixed as string element be placed inside the typeAttributes (array element) inside attributes of the element?

test/refract/test-JsonSchema.cc
```diff
@@ -259,6 +309,23 @@ SCENARIO("JSON Schema serialization of StringElement", "[json-schema]")
...
+        el->element("Flip");
+        el->attributes().set("fixed", from_primitive(true));
+
+        WHEN("a JSON Schema is generated from it")
+        {
+            auto result = schema::generateJsonSchema(*el);
+
+            THEN("the schema matches the string \"foo\"")
+            {
+                REQUIRE(to_string(result) == R"({"$schema":"http://json-schema.org/draft-04/schema#","definitions":{"Flip":{"type":"string","enum":["foo"]}},"$ref":"#/definitions/Flip"})");
```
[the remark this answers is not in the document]
tjanc: I think it depends how you read the standard: http://json-schema.org/latest/json-schema-core.html#rfc.section.8
```
An object schema with a "$ref" property MUST be interpreted as a "$ref" reference. The value of the "$ref" property MUST be a URI Reference. Resolved against the current URI base, it identifies the URI of a schema to use. All other properties in a "$ref" object MUST be ignored.
```
You might have a point. I don't think it's invalid JSON Schema, although it's quite possible a parser might ignore the `schema` property. The URI is still valid, the fact a JSON Schema parser should ignore properties coinciding with `$ref` doesn't change the fact. Also, I've tried all examples against at least one linter without issues and this idiom seems to be common place on discussion forums.

We could wrap the root in an `allOf` though, just to be on the safe side. What do you think?

src/refract/JsonSchema.cc
tjanc: Because with this commit the implementation needs to choose whether to write into the schema directly or whether to write into `definitions`. See calls to `choose_target`. I could rename `s` to `schema`, but then all occurences of `schema` below need to be replaced with something... e.g. `target`?

test/refract/test-JsonSchema.cc
tjanc: Yes, you're right. I completely missed this, thanks :+1:

test/refract/test-JsonSchema.cc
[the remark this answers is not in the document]
kylef: I was able to get the following schema to work correctly on various validator implementations I tested. So seems it could be allowed.

```json
{
  "$schema":"http://json-schema.org/draft-04/schema#",
  "definitions":{
    "Items":{"type":"array", "items": {"$ref":"#/definitions/Item"}},
    "Item": {"type": "object"}
  },
  "$ref":"#/definitions/Items"
}
```

Issue #542, opened 2018-05-08
[v4.0.0-pre.0] No rule to make target `out/Release/obj.target/libsos/ext/sos/src/sos.o', needed by `out/Release/obj.target/libsos.a'


chris48s:
Hi. I'm trying to compile `v4.0.0-pre.0` from https://github.com/apiaryio/drafter/archive/v4.0.0-pre.0.tar.gz as a shared library on Ubuntu linux. I'm able to get `v3.2.7` from https://github.com/apiaryio/drafter/releases/download/v3.2.7/drafter-v3.2.7.tar.gz working OK, but `v4.0.0-pre.0` is giving me:

```
$ ./configure --shared
creating ./config.gypi
creating ./config.mk
creating makefiles
All OK.

$ make libdrafter
make -C ./build V=1 libdrafter
make[1]: Entering directory `/path/to/stuff/drafter-4.0.0-pre.0/build'
make[1]: *** No rule to make target `out/Release/obj.target/libsos/ext/sos/src/sos.o', needed by `out/Release/obj.target/libsos.a'.  Stop.
make[1]: Leaving directory `/path/to/stuff/drafter-4.0.0-pre.0/build'
make: *** [libdrafter] Error 2
```

The release notes specify GCC 5.3 or higher, but here's my GCC details, just in case:

```
$ gcc --version
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.9) 5.4.0 20160609
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

I'm afraid I'm a bit of a n00b when it comes to C, so it is possible I am missing something obvious. Any advice would be appreciated

kylef:
Can you try the link https://github.com/apiaryio/drafter/releases/download/v4.0.0-pre.0/drafter-4.0.0-pre.0.tar.gz to download the source. The one you grabbed doesn't include some of the git submodule dependencies as GitHub computes the tarball without them. I've just uploaded the complete 4.0.0-pre.0 tarball.

chris48s:
Hi. Thanks for updating that, but now the new archive seems to be missing the `/tools` dir. I'm getting

```
$ ./configure --shared
creating ./config.gypi
creating ./config.mk
creating makefiles
[...]

PR #545, opened 2018-05-25
Fix tarball generation to include tools/gyp in tarball

kylef:
In Drafter master (and 4 pre release) we moved gyp to be in tools/gyp (previously in ext) which causes problems with the generated release tar balls since gyp wouldn't be included.

This was discovered in https://github.com/apiaryio/drafter/issues/542#issuecomment-387804944. This fix makes it so that the tools/gyp directory is not deleted.

### DIFF
--- a/tools/make-tarball.sh
+++ b/tools/make-tarball.sh
@@ -29,11 +29,11 @@ $(pwd)/tools/make-tarball.sh checkout $TMPDIR/$DIR/
 (
   # Remove unnessecery files
   cd $TMPDIR/$DIR
-  rm -fr appveyor.yml circle.yml .npmignore tools/ package.json emcc/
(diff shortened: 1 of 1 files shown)


tools/make-tarball.sh
```diff
@@ -29,11 +29,11 @@ $(pwd)/tools/make-tarball.sh checkout $TMPDIR/$DIR/
...
   # Remove unnessecery files
   cd $TMPDIR/$DIR
-  rm -fr appveyor.yml circle.yml .npmignore tools/ package.json emcc/
+  rm -fr appveyor.yml circle.yml .npmignore tools/*.{sh,py} tools/homebrew package.json emcc/
 )
 
 (
   cd $TMPDIR
-  echo "Creating tarball"
+  echo "Creating tarball $TARBALL"
```
kylef: This makes it so that the message tells you where the tarball was created otherwise you're left guessing.

Issue #547, opened 2018-05-29
Bad Error Message


goganchic:
I've opened MSON specification and found such example https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#232-value-member-types so I've created such piece of code:

```
# Tesla

## Users [POST /users]

+ Response 200 (application/json)
    + Attributes
        + (array)
            + foo
            + bar
```

I use drafter from master and it reports me such an error:

> drafter: ../src/refract/Element.h:76: const DataType& refract::Element< <template-parameter-1-1> >::get() const [with DataType = refract::dsd::String]: Assertion `hasValue_' failed.

Is it bug or I've done something wrong?

P.s. such example works fine:

```
# Tesla

## Users [POST /users]

+ Response 200 (application/json)
    + Attributes (array)
        + foo
        + bar
```

pksunkara:
You have done something wrong. `Attributes` is an object, but you gave it a value member with `(array)`. The error should have a better message though.

goganchic:
ok, thanks!

pksunkara:
No, we need to keep this open to have a better error message.

goganchic:
ok

kylef:
The following patch will restore Drafter 3.2.7 behaviour. Although I do not think this is correct and we should likely catch the problem earlier. I think `GetKeyAsString()` should require an actual key to be there as it does in master. @tjanc what do you think? 

```diff
diff --git a/src/refract/VisitorUtils.cc b/src/refract/VisitorUtils.cc
index 8192a8c9..6a1d2b08 100644
--- a/src/refract/VisitorUtils.cc
+++ b/src/refract/VisitorUtils.cc
@@ -27,7 +27,9 @@ std::string refract::GetKeyAsString(const MemberElement& e)
     auto element = e.get().key();

     if (auto str = TypeQueryVisitor::as<const StringElement>(element)) {
-        return str->get();
+        if (!str->empty()) {
+            return str->get();
+        }
     }

     if (auto ext = TypeQueryVisitor::as<const ExtendElement>(element)) {
```

PR #551, opened 2018-06-07
Fix isssue #547

klokane:


test/fixtures/render/issue-547.json
```diff
@@ -0,0 +1,212 @@
...
+                        }
+                      },
+                      "content": [
+                        {
+                          "element": "dataStructure",
+                          "content": {
+                            "element": "object",
+                            "content": [
+                              {
+                                "element": "member",
```
pksunkara: This member shouldn't exist.

klokane: I have try to generate output from drafter v2.3.1, there is same member like this one.

So my PR do not change previous behavior. In my opinion it is not part of this bugfix.
So if you like to change this behavior let open new issue and discuss it.

PR #552, opened 2018-06-08
Finished JSON Schema serialization

tjanc:
+ adds support for Object
+ adds support for nullable
+ fixes some inconsistencies
+ hooks in new serialization


src/refract/dsd/Object.cc
```diff
@@ -58,6 +59,23 @@ Object& Object::operator=(Object rhs)
...
+
+    if (!empty())
+        if (auto mbr = TypeQueryVisitor::as<const MemberElement>(el.get()))
+            if (!mbr->empty() && mbr->get().key())
+                if (auto str = TypeQueryVisitor::as<const StringElement>(mbr->get().key()))
+                    if (!str->empty()) {
+                        auto it = find(str->get().get());
+                        if (it != end())
+                            erase(it);
+                    }
```
klokane: This cascade is little bit confusing while reading

tjanc: It just checks whether we are adding a MemberElement with StringElement key so we ensure uniqueness - if the key is already encountered, it shall be removed. Specific suggestions how to rewrite this?

goganchic:
Great work! But would you please split all changes into small commits? E.g. you can extract refactoring staff like this https://github.com/apiaryio/drafter/pull/552/files#diff-b3cd98f0ba049e7bdc9ae6c68e9e1345L17 into separate commit. I think small commits will be much more easier to check and understand.

ext/snowcrash/ext/markdown-parser/src/MarkdownParser.cc
```diff
@@ -57,10 +57,6 @@ void MarkdownParser::parse(const ByteBuffer& source, MarkdownNode& ast)
     m_source = NULL;
     m_sourceLength = 0;
     m_listBlockContext = false;
-
-#ifdef DEBUG
```
pksunkara: Why is DEBUG removed?

kylef:
@tjanc can you please rebase this on master so it will trigger the new CircleCI tests.

ext/snowcrash/ext/markdown-parser/src/MarkdownParser.cc
pksunkara: Unless we have an alternative, please don't remove it. It's pretty useful when debugging.

tjanc: I'm still convinced writing debug information to standard output out of nowhere is a bit too hacky. You can always log like this:

```
LOG(debug) << "Hello " << 42 << " or some other value";
```

This amounts to the same thing but can be dynamically turned on with a flag: `drafter -L` or `drafter --enable-log`. Also, it logs into the file `drafter.log` instead of polluting standard output. If needed, the `ENABLE_LOGGING` macro can be used to force logging while debugging.

tjanc: This shall be solved in a future PR, for now I'm reverting the change.

PR #554, opened 2018-06-18
Move to Circle CI 2.0

kylef:
This pull request moves to Circle CI 2.0 and adds a lot more compilers into our CI mix. Previously we only tested on a single version of gcc and now we test on all major supported versions alongside macOS and clang(s). I was on the edge if we should test each major gcc other than min and major but given the build is parallelised we are not loosing any time doing so. Infact the whole CI matrix for Circle CI 2 complex quicker than the single Windows build on AppVeyor.

The PR decouples the integration and valgrind tests as installing Ruby on each image complicates things and I don't believe running those tests on each compiler is warranted as I don't think it is remotely likely to ever produce different results.

### Jobs

- [x] lint
- [x] test
    - [x] gcc (min)
    - [x] gcc (latest)
    - [x] clang (min)
    - [x] clang (latest)
    - [x] macOS (min)
    - [x] macOS (latest)
- [x] integration tests
- [x] valgrind (memory) tests
- [x] release


.circleci/config.yml
```diff
@@ -0,0 +1,131 @@
...
+            ./configure
+            make
+      - run:
+          name: Test
+          command: make test
+
+  - &test-clang-base
+    <<: *test-base
+    environment:
+      CXXFLAGS: '-Werror -Wno-error=unused-function -Wno-error=pessimizing-move -Wno-error=missing-braces'
```
kylef: Clang (both on macOS and Linux) produce further warnings, we should probably fix `pessimizing-move` although the `missing-braces` warning isn't really worth doing anything about and seems later versions of Clang remove this particular warning.

pksunkara:
What do you think about supporting other clang versions?

tools/clang-format-check.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-clang-format-5.0 -style=file -output-replacements-xml $(git ls-files | grep '\.cc\|\.h$') | grep -c "<replacement "
+set -e
```
kylef: Because of the missing `-e` and we're piping this would exit 0 even if it failed such as command not failed. Which clang-format-5.0 never existed so this silently passes.

As you can see, master isn't even running these and silently passing:

screen shot 2018-06-19 at 13 41 48

The clang-format tests are failing and that's because the code doesn't follow the rules. It's outside of this PR to fix so as the lint is not "required" we can continue and merge PR and fix files not confirming to clang-format separately.

kylef:
This PR is now ready for review.

tools/clang-format-check.sh
kylef: Fixing this will likely conflict with #552 so we should likely wait until thats merged.

PR #560, opened 2018-07-09
(fix) fixed issue 556

### LINKED ISSUE #556: Strange error for incorrect document
I'm looking throw function [parseTypeSpecification](https://github.com/apiaryio/snowcrash/blob/master/src/MSONUtility.h#L187) and try to find out how Drafter process such invalid document:

```
# My API

# Data Structures

# Foo[Hello]

+ a
+ b

# GET /foo

    + Attributes (array[Foo[Hello]], fixed-type)
```

When I start Drafter I get such error:

```
./bin/drafter test.apib
Assertion failed: (hasValue_), function get, file ../src/refract/Element.h, line 70.
[1]    55413 abort      ./bin/drafter test.apib
```

So I have several questions:

* Is it ok to use brackets in `Named Type` name?
* Is it a valid error for such document?
* What is formal definition of such error? (e.g. I should not use brackets in type definition or something like this)

Used Drafter version: master branch, commit 2318c78d

tjanc:
fixes #556

### DIFF
--- a/src/refract/ExpandVisitor.cc
+++ b/src/refract/ExpandVisitor.cc
@@ -211,6 +211,9 @@ namespace refract
 
             auto origin = ExpandMembers(e);
             origin->meta().erase("id");
+
+            if(extend->empty())
+                extend->set();
             extend->get().push_back(std::move(origin));
 
             return std::move(extend);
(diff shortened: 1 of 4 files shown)


tjanc:
Added integration test.

PR #563, opened 2018-07-16
Remove SOS library; use utils/so instead

tjanc:
- hooks in the new "simple object" code into all json/yaml rendering
- implements API Elements -> "simple object" transformation

Fixes JSON/YAML escaping.


test/fixtures/schema/variable-property.json
```diff
@@ -175,7 +175,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"_links\": {\n    \"self\": {\n      \"href\": \"a URI\"\n    }\n  }\n}"
+                              "content": "{\n  \"_links\": {}\n}"
```
kylef: Given the APIB, the "self" link example should be present given it is variable property with defined key?

```apib
# Group Example
# Resource [/example]

## Action [GET]

    + Attributes
        - _links
            - *self*
                - href: a URI
```

test/fixtures/schema/string-literal.json
```diff
@@ -164,7 +164,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"street\": [\n    \"bye bye\",\n    \"san francisco\"\n  ]\n}"
+                              "content": "{\n  \"street\": [\n    \"\",\n    \"bye bye\",\n    \"san francisco\"\n  ]\n}"
```
kylef: This is an interesting one:

```apib
# Group Example
# Resource [/example]

## Action [GET]

    + Attributes
        - street (array[string], fixed)
            - bye bye
            - san francisco
```

The empty `string` which the user intended as a "type hint" as `array[string]` is becoming a fixed value in the output. The current JSON Schema is adding it two. So the matching JSON Schema requires this change to validate properly.

HOWEVER, the JSON Schema and example in Drafter 3.2.7 stable doesn't do this.

test/fixtures/schema/variable-property.json
tjanc: I think I've misinterpreted variable properties. Isn't the token `self` in `*self*` just a variable name?

test/fixtures/schema/sample-complex.json
```diff
@@ -169,7 +169,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"colors\": [\n    \"red\"\n  ]\n}"
+                              "content": "{\n  \"colors\": []\n}"
```
kylef: Given the APIB, I'd expect the sample value to be used:

```apib
# Group Example
# Resource [/example]

## Action [GET]

  + Attributes
      - colors (array)
          - Sample: red
          - Sample
              - blue
              - green
```

test/fixtures/schema/optional.json
```diff
@@ -151,7 +151,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{}"
+                              "content": "{\n  \"street\": \"\"\n}"
```
kylef: Optional without any possible value being rendered. I don't think we should render it in the case it has no values.

```apib
# Group Example
# Resource [/example]

## Action [GET]

    + Attributes
        - street (optional)
```

test/fixtures/schema/string-literal.json
tjanc: That's because the JSON Schema and JSON Value renderers get some API Elements, and they cannot know that the API Blueprint was what it was. Check out the corresponding API Elements:https://github.com/apiaryio/drafter/blob/f17705e4b74a134db5848bf3d40cbc4eebb07f12/test/fixtures/schema/string-literal.json#L129

test/fixtures/schema/optional.json
kylef: This may also introduce regressions for dredd users as when an optional property is present in request body it MAY have additional validation. As we're using empty string it might not meet any validation on server side and thus cause existing Dredd to start failing using request bodies generated with this logic.

test/fixtures/schema/string-literal.json
tjanc: So, this might be a bug, but it's a bug during AST -> API Elements transformation, not API Elements -> JSON Schema/Value.

test/fixtures/schema/variable-property.json
kylef: https://apiblueprint.org/documentation/mson/specification.html#322-variable-property-name

> `*rel (Custom String)* (object)`
> Where rel is a sample value for the arbitrary Property Name of a Property Member Declaration.

I believe it matches the arbitrary property name sample value section of spec.

The example itself would imply that was the intention as in the HAL spec (HAL being the content format being used) `self` is a common relation identifier.

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    so::Value renderValue(const BooleanElement& e, TypeAttributes options)
+    {
+        LOG(debug) << "rendering BooleanElement to JSON Value";
+
+        options = updateTypeAttributes(e, options);
+
+        if (e.empty()) {
+            if (options.test(NULLABLE_FLAG))
```
kylef: Given the implementation for all primitive types are the same, can we make this re-usable instead of duplicating the business logic on value precedence in each renderValue?

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,499 @@
...
+        options.set(NULLABLE_FLAG, hasNullableTypeAttr(e));
+
+        if (hasRequiredTypeAttr(e))
+            options.set(REQUIRED_FLAG);
+
+        if (hasOptionalTypeAttr(e))
+            options.reset(REQUIRED_FLAG);
+
+        const auto* k = e.get().key();
+        const auto* v = e.get().value();
```
kylef: IMO would be nicer to use clearer variable names instead of `k`, `v`, `e`, `s`, `a`. Descriptive variable names go a long way in making this more readable than trying to remember which character represents which things. We're humans not machines, readability counts.

test/fixtures/circular/array-in-object.json
```diff
@@ -126,7 +126,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"id\": \"1\",\n  \"data\": []\n}"
+                              "content": "{\n  \"id\": \"1\",\n  \"data\": [\n    {}\n  ]\n}"
```
kylef: In this case, I do not think the object should be rendered in the array. As the object can have required values that wouldn't be rendered. In that case the new renderer can produce JSON examples that wouldn't pass validation under the generated JSON Schema.

This can be demonstrated if this circular test fixture was updated to have a required parameter (and use fixed-type) on the array:

```apib
# Group Example
# GET /
    + Attributes (A)

# Data Structures

# A (object)
- id: 1 (required)
- data (array[A], fixed-type)
```

Which would produce the following:

```json
{
  "id": "1",
  "data": [
    {}
  ]
}
```

Unfortunately, it seems the JSON Schema isn't generated as I expected and thus the example I was trying to share doesn't work. As it doesn't use `definitions` to support circular schema generation (I though that was added in recent changes).

```json
{
  "type": "object",
  "properties": {
    "id": {
      "type": "string"
    },
    "data": {
      "type": "array",
      "items": {
        "anyOf": [
          {
            "type": "object"
          }
        ]
      }
    }
  },
  "required": [
    "id"
  ]
}
```

test/fixtures/extend/circular.json
```diff
@@ -134,7 +134,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "[\n  [],\n  []\n]"
+                              "content": "[\n  null,\n  null\n]"
```
kylef: I cannot see how this example can be correct. There isn't a `nullable` type attributes on any structures in the source document so cannot be any `null` values present.

test/fixtures/extend/complex.json
```diff
@@ -126,7 +126,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"d\": {\n    \"c\": [\n      {\n        \"f\": \"\"\n      }\n    ]\n  }\n}"
+                              "content": "{\n  \"d\": {\n    \"c\": [\n      {\n        \"f\": null\n      }\n    ]\n  }\n}"
```
kylef: Ditto the above comment.

> I cannot see how this example can be correct. There isn't a `nullable` type attributes on any structures in the source document so cannot be any `null` values present.

test/fixtures/render/enum-default-multiple.json
```diff
@@ -199,7 +199,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"enumField\": \"g\"\n}"
+                              "content": "{\n  \"enumField\": \"e\"\n}"
```
kylef: Wouldn't this be a regression? Previously the last default value took precedence and now the first one does.

```apib
# Group Example
# GET /

    + Attributes
        + enumField (enum)
            + Default: e
            + Default: f
            + Default: g
```

Although there is no warning saying I am shadowing the previous default values, but we are generally preferring the last one before and in other code paths so I think we should continue to do so for consistency instead of changing the behaviour during refactoring.

test/fixtures/render/enum-default.json
```diff
@@ -199,7 +199,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"enumField\": \"c\"\n}"
+                              "content": "{\n  \"enumField\": \"a\"\n}"
```
kylef: I think this is a regression. Before we was using the default value `c` and now we're using the first enumeration value. I believe the default should take precedence.

test/fixtures/render/enum-sample.json
```diff
@@ -204,7 +204,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"enumField\": \"c\"\n}"
+                              "content": "{\n  \"enumField\": \"a\"\n}"
```
kylef: > I think this is a regression. Before we was using the default value `c` and now we're using the first enumeration value. I believe the default should take precedence.

Similar to above, I think the sample value should take precedence over the enumeration members.

test/fixtures/render/issue-547.json
```diff
@@ -131,7 +131,7 @@
                               "content": "application/json"
                             }
                           },
-                          "content": "{}"
+                          "content": "{\n  \"\": [\n    \"foo\",\n    \"bar\"\n  ]\n}"
```
kylef: I am wondering if this is correct. The underlying ADD is the following:

```apib
## GET /

    + Attributes
        + (array)
            + foo
            + bar

```

There is a warning as the user has defined a property without a name and thus previously it was discarded. This logic is now proceeding with empty string.

test/fixtures/render/mixin-array-sample.json
```diff
@@ -146,7 +146,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "[\n  \"x\",\n  \"a\",\n  \"b\",\n  \"y\"\n]"
+                              "content": "[\n  \"x\",\n  \"y\"\n]"
```
kylef: I think this is a regression. The document has an `include` which brings in the `a` and `b` sample values which are being removed.

test/fixtures/render/mixin-override.json
```diff
@@ -221,7 +221,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"tasks\": [\n    {\n      \"status\": \"CREATED\",\n      \"name\": \"one\"\n    },\n    {\n      \"status\": \"CREATED\",\n      \"name\": \"two\"\n    }\n  ]\n}"
+                              "content": "{\n  \"tasks\": [\n    {\n      \"status\": \"\",\n      \"name\": \"one\"\n    },\n    {\n      \"status\": \"\",\n      \"name\": \"two\"\n    }\n  ]\n}"
```
kylef: I think this is a regression. Previously the following enum produce the rendered value `CREATED` whereas it is now an empty string.

```apib
## Status
+ status (enum[string])
    + `CREATED`
    + `QUEUED`
    + `RUNNING`
    + `FINISHED`
    + `ERROR
```

test/fixtures/render/object-array-string.json
```diff
@@ -184,7 +184,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"a\": [\n    \"b\",\n    \"c\"\n  ]\n}"
+                              "content": "{\n  \"a\": [\n    \"\",\n    \"b\",\n    \"c\"\n  ]\n}"
```
kylef: This introduces an empty string into the rendered array which I do not believe should be there. This is being done based on the type-hint of the array (`array[string]`).

test/fixtures/schema/array-fixed-inline-samples.json
```diff
@@ -185,7 +185,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"address\": [\n    \"city\"\n  ]\n}"
+                              "content": "{\n  \"address\": [\n    \"street\",\n    \"city\",\n    \"state\"\n  ]\n}"
```
kylef: Outside of the scope of this PR to fix. While reviewing this fixture. I believe the generated JSON Schema is incorrect.

Given what the `**` represent in fixed array is that the value is no longer required although the JSON Schema does require a min of 3 items. Shouldn't it be possible that the existing value of `["city"]` would validate given the MSON syntax as a valid value?

```json
{
  "type": "object",
  "properties": {
    "address": {
      "type": "array",
      "minItems": 3,
      "items": [
        {
          "type": "string"
        },
        {
          "type": "string",
          "enum": [
            "city"
          ]
        },
        {
          "type": "string"
        }
      ],
      "additionalItems": false
    }
  }
}
```

screen shot 2018-07-17 at 18 54 12

test/fixtures/schema/array-with-nested-type.json
```diff
@@ -157,7 +157,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"address\": [\n    4,\n    2,\n    42\n  ]\n}"
+                              "content": "{\n  \"address\": [\n    0,\n    4,\n    2,\n    42\n  ]\n}"
```
kylef: This `0` is being introduced, similar to another comment I made about array type-hints above.

test/fixtures/schema/array-with-nested-types.json
```diff
@@ -164,7 +164,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"address\": [\n    \"4\",\n    \"2\",\n    \"hello\",\n    \"42\"\n  ]\n}"
+                              "content": "{\n  \"address\": [\n    0,\n    \"\",\n    \"4\",\n    \"2\",\n    \"hello\",\n    \"42\"\n  ]\n}"
```
kylef: This `0` is being introduced, similar to another comment I made about array type-hints above.

test/fixtures/schema/boolean-literal.json
```diff
@@ -164,7 +164,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"street\": [\n    true,\n    false\n  ]\n}"
+                              "content": "{\n  \"street\": [\n    false,\n    true,\n    false\n  ]\n}"
```
kylef: Another array type-hint being rendered.

test/fixtures/schema/default-attribute.json
```diff
@@ -184,7 +184,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"list\": \"4\"\n}"
+                              "content": "{\n  \"list\": \"3\"\n}"
```
kylef: This is rendering the first member of an enum instead of the default value 4.

test/fixtures/schema/default-section.json
```diff
@@ -184,7 +184,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"list\": \"4\"\n}"
+                              "content": "{\n  \"list\": \"3\"\n}"
```
kylef: Same problem as the `test/fixtures/schema/default-attribute.json` fixture.

test/fixtures/schema/enum-nullable.json
```diff
@@ -176,7 +176,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "{\n  \"address\": 4\n}"
+                              "content": "{\n  \"address\": 0\n}"
```
kylef: The value `4` is not being rendered.

```apib
# Group Example
# Resource [/example]

## Action [GET]

    + Attributes
        + address (enum[number], nullable)
            + 4
```

test/test-CAPI.c
```diff
@@ -34,7 +34,7 @@ int test_parse_and_serialize()
 
     size_t len = strlen(expected);
 
-    assert(strncmp(out, expected, len) == 0);
+    assert(strncmp(out, expected, len) <= 0);
```
kylef: Is this returning negative because there is a missing new line at the end of the `out` file? Can we fix this instead of working around the problem in tests?

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,499 @@
...
+
+    so::Object renderValue(const ObjectElement& e, TypeAttributes options)
+    {
+        LOG(debug) << "rendering ObjectElement to JSON Value";
+        so::Object result{};
+
+        options = updateTypeAttributes(e, options);
+
+        if (e.empty()) {
+            if (const auto* smple = findSample(e)) {
```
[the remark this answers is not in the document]
tjanc: Consistency with `default` being `deflt` (default is reserved); same character width. But I have no problem with changing it to `sample`.

features/fixtures/refract.sourcemap.yaml
```diff
@@ -1,2161 +1,2160 @@
-element: "parseResult"
-content:
+"element": "parseResult"
```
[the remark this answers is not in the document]
tjanc: I'm surprised this didn't come up when YAML serialization was implemented.
I will modify the logic there to scan for a column (`:`) and only then quote it.

test/fixtures/schema/boolean-literal.json
tjanc: Again, this is not a bug in this branch. It's a bug in master. Should I fix it here?

kylef: The diff is coming from master > this branch. So the bug would be introduced in this branch?

tjanc: The bug surfaces in this branch, because my implementation doesn't try and work around bugs in the API Elements tree. It expects API Elements according to the spec and interprets it accordingly.

The former implementation worked around some of these bugs (i.e. it interpreted too much), and because of that we had wrong API Elements and correct JSON Value.

I can fix the whole type-hint thing in this branch, but I'm not sure how trivial the solution will be (haven't seen the AST -> API Elements transformation code for a while).

src/Render.cc
```diff
@@ -90,13 +91,12 @@ namespace drafter
         // One of this will always execute since we have a catch above for not having render format
         switch (renderFormat) {
             case JSONRenderFormat: {
-                refract::RenderJSONVisitor renderer;
-                refract::Visit(renderer, *expanded);
-                return std::make_pair(renderer.getString(), NodeInfo<Asset>::NullSourceMap());
+                std::stringstream ss{};
+                utils::so::serialize_json(ss, refract::generateJsonValue(*expanded));
```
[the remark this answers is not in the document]
tjanc: Good observation. We could shave off a lot of memory & cache shuffling. Will look into it.

src/refract/Registry.cc
```diff
@@ -41,10 +40,6 @@ std::string Registry::getElementId(IElement& element)
         throw LogicError("Element has no ID");
     }
 
-    // FIXME: remove dependecy on SosSerializeCompactVisitor
-    SosSerializeCompactVisitor v;
-    VisitBy(*it->second, v);
-
```
pksunkara: PR looks good code wise, but I couldn't understand why this part is removed. Would appreciate a small reply explaining. Especially, why we were visiting in the first place if it's not needed.

tjanc: We didn't. This was dead code.

test/test-CAPI.c
tjanc: This is my mistake. I saw `len = strlen(out)` on line 35 for some reason. Why do we only compare with a prefix instead of the whole document here?

test/fixtures/schema/enum-nullable.json
tjanc: API Elements -> JSON Value is correct. Added #574 to allow for a better first value on type-hinted enums.

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    so::Value renderValue(const StringElement& e, TypeAttributes options)
+    {
+        LOG(debug) << "rendering StringElement to JSON Value";
+
+        options = updateTypeAttributes(e, options);
+
+        if (e.empty()) {
+            if (options.test(NULLABLE_FLAG))
```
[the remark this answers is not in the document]
klokane: Agree, it is fixed or some other result from discussion?

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,550 @@
...
+
+    TypeAttributes inherit_flags(TypeAttributes options) noexcept
+    {
+        options.reset(FIXED_TYPE_FLAG);
+        options.reset(NULLABLE_FLAG);
+        options.reset(REQUIRED_FLAG);
+        return options;
+    }
+
+    TypeAttributes inherit_or_pass_flags(TypeAttributes options, const IElement& e)
```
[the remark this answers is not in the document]
tjanc: Does this cause a readability challenge to you? It's non-interface code burried in an anonymous namespace in an implementation file, and I maintain `interitOrPassFlags` is less readable.

src/RefractDataStructure.cc
```diff
@@ -342,11 +366,13 @@ namespace
                     break;
 
                 case mson::TypeSection::SampleClass:
-                    data.samples.push_back(fetch(typeSection, context, defaultNestedType));
+                    if (!typeSection.node->content.value.empty() || !typeSection.node->content.elements().empty())
```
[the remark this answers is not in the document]
tjanc: How would you call the function?

src/refract/JsonValue.cc
```diff
@@ -0,0 +1,550 @@
...
+    so::Object renderValueSpecific(const ObjectElement& e, TypeAttributes options)
+    {
+        LOG(debug) << "rendering ObjectElement to JSON Value";
+        so::Object result{};
+
+        options = updateTypeAttributes(e, options);
+
+        if (e.empty()) {
+            if (const auto* sample = findFirstSample(e)) {
+                if (!sample->empty())
```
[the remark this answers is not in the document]
tjanc: Agreed, I somewhere have a commit fixing this; there already is a function for that: `renderSampleOrDefault`.

src/refract/JsonValue.cc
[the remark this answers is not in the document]
tjanc: This is questionable. Of course, if there are no samples or default or content or whatnot present, we will fall back to rendering `null`. If the user specifies `nullable`, I thought `null` is actually a really interesting value for them (i.e. to test against). There is no harm in changing this behaviour to ignore `nullable` alltogether (I would keep a LOG warning then if we fall back to `null` without `nullable` present). @kylef @klokane what are your thoughts?

src/refract/JsonValue.cc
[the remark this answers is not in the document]
tjanc: It's not a rule, is it? It's a guideline. We have this guideline to enhance readability.

I don't like the mechanic review, that's all. We are discussing the casing in the name of a private implementation function, more than a month after the branch was first reviewed; at this point I explicitly asked for review of semantics, not syntax (i.e. whether the fixture diffs are to your liking).

The reviewer always has the last call, and that actually is a rule. The function shall be renamed to java-talk.

src/RefractAPI.cc
```diff
@@ -73,7 +79,48 @@ namespace
...
         return std::move(element);
     }
-}
+
+    bool isRequest(const NodeInfo<snowcrash::Action>& action)
+    {
+        return !action.isNull() && !action.node->method.empty();
+    }
+
+    NodeInfoByValue<snowcrash::Asset> renderPayloadBody(
```
pksunkara: Isn't it better to move these two functions to Render.cpp?

tjanc: We don't need these functions anywhere else, why would we have it declared in a header (`Render.h`)?

src/refract/JsonValue.cc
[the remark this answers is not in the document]
tjanc: I'm leaving out braces if they are not necessary. If they are there, probably it's an artifact from debugging messages being embedded there before.

src/refract/JsonValue.cc
tjanc: Agreed

test/fixtures/schema/array-fixed-inline-samples.json
pksunkara: I actually don't see anything wrong here either in the body or the schema.

tools/clang-format-check.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash
```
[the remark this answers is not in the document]
tjanc: Source: https://github.com/apiaryio/drafter/commit/269ca6a00cbd845ff7628c0057124bad0d830269

test/fixtures/render/issue-246.json
```diff
@@ -126,7 +126,7 @@
                                   "content": "application/json"
                                 }
                               },
-                              "content": "[\n  {\n    \"key1\": \"value1\"\n  },\n  {\n    \"a\": []\n  },\n  []\n]"
+                              "content": "[\n  {\n    \"key1\": \"value1\"\n  },\n  {},\n  {\n    \"a\": [\n      {}\n    ]\n  },\n  [\n    {}\n  ]\n]"
```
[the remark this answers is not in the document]
pksunkara: Decided that this is correct and good.

PR #564, opened 2018-07-17
add PROPPATCH and SEARCH to HTTP_REQUEST_METHOD

butonic:
I am trying to find an api tool that can deal with WebDAV. Adding these two methods would be a start.


butonic:
suporseded by https://github.com/apiaryio/snowcrash/pull/441

pksunkara:
@butonic, we have deprecated snowcrash. Your PR here is correct.

PR #572, opened 2018-07-24
Release 4.0.0-pre1

kylef:
I've done some minor reworking of the CHANGELOG to make the changes clearer and link to the API Elements spec which contains more details.


pksunkara:
@kylef Is this the only changelog needed?

PR #576, opened 2018-07-31
update formating according to clang-format 5.0

klokane:


pksunkara:
Merged kyle's PR, which means the first commit here is not needed.

pksunkara:
Looks like this have been merged into a previous branch we merged.

Issue #519, opened 2017-11-08
JSON Visitor should use `sample` value for rendering JSON in array


kylef:
For example, if I want to reference another data structure inside an array of fixed-type. I'd like to provide multiple examples to be rendered in the JSON array example.

```apib
# API

## GET /users

+ Response 200 (application/json)
    + Attributes
        + users (array[User], fixed-type)
            + Sample
                + (User)
                    + username: doe
                + (User)
                    + username: tim

## Data Structures

### User

+ username: kyle
```

The `Sample` section is ignored when it comes to rendering a JSON body and I think the sample should be included.

### Current Behaviour

```json
{
  "users": [
    {
      "username": "kyle"
    }
  ]
}
```

### Expected Behaviour

```json
{
  "users": [
    {
      "username": "doe"
    },
    {
      "username": "tim"
    }
  ]
}
```

kylef:
The same problem exists for objects as well as array.

kylef:
Another case to reproduce:

```apib
+ Response 200 (application/json)
    + Attributes
        + location (array, fixed)
            + `43.3` (number, sample)
            + `-8.41` (number, sample)
```

PR #587, opened 2018-08-13
Fix issue #519

tjanc:
## Context
1. A `fixedType` Array element should be treated as the type of lists of _any of_ the elements in its content.
2. An Enum element should be treated as the type of _any of_ the elements in its `enumerations` attribute.

One could argue both _any of_ combinators are expected to behave similarly.

## Current behaviour
In (1) the JSON value rendering precedence is **content** - **sample** - **default**.
In (2) the JSON value rendering precedence is **content** - **sample** - **default** - **enumerations**.

This is unified only syntactically. In (1) what seems to be _content_ actually is a `fixedType`'d Array element's way to define its `enumerations`.

## Fix
Issue #519 comes with this suggestion:
In (1) the JSON value rendering precedence SHOULD BE **sample** - **default** - **content**.

This PR does just this, it changes the precedence of value source for `fixedType` Array elements in the JSON value renderer.

### DIFF
--- a/src/refract/JsonValue.cc
+++ b/src/refract/JsonValue.cc
@@ -219,16 +219,26 @@ namespace
 
         so::Array result{};
         if (e.empty()) {
+            if (options.test(FIXED_TYPE_FLAG))
+                LOG(warning) << "empty fixedType ArrayElement in backend";
+
             auto alt = renderSampleOrDefault(e, inherit_flags(options));
             if (alt.first)
                 return std::move(alt.second);
-        } else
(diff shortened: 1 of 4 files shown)


kylef:
Discussed with @tjanc, we will close this PR as we should find a way to solve this more generically instead of a work-around on fixed-type specifically. Closing but leaving branch for future reference.

kylef:
We'll likely have to change MSON / API Elements to be able to solve this problem.

PR #593, opened 2018-08-22
(fix) variable properties APIE and JSON rendering

tjanc:


test/fixtures/mson/variable-property-name.json
```diff
@@ -300,6 +300,24 @@
...
+                    "element": "member",
+                    "attributes": {
+                      "variable": {
+                        "element": "boolean",
+                        "content": true
+                      }
+                    },
+                    "content": {
+                      "key": {
+                        "element": "StrSample"
```
pksunkara: Shouldn't this line be expanded too?

PR #601, opened 2018-09-05
(fix) removed file logging from production build

tjanc:
* mainly to allow wasm build


drafter.gyp
```diff
@@ -428,6 +428,7 @@
       "conditions" : [
         [ 'libdrafter_type=="static_library"', { 'defines' : [ 'DRAFTER_BUILD_STATIC' ] }],
       ],
+      "defines": ["LOGGING"],
```
pksunkara: I don't think this should be here.

tjanc: Why?

pksunkara: Wrong message at wrong place. Ignore it.

PR #602, opened 2018-09-06
chore: Make sure tag gets tested on circleci

pksunkara:
> CircleCI does not run workflows for tags unless you explicitly specify tag filters. Additionally, if a job requires any other jobs (directly or indirectly), you must use regular expressions to specify tag filters for those jobs.

Also an example from the same page on circleci documentation

```yml
workflows:
  version: 2
  build-n-deploy:
    jobs:
      - build:
          filters:  # required since `deploy` has tag filters AND requires `build`
            tags:
              only: /.*/
      - deploy:
          requires:
            - build
          filters:
            tags:
              only: /^v.*/
            branches:
              ignore: /.*/
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,22 +23,27 @@ aliases:
     environment:
       CXXFLAGS: '-Werror -Wno-error=unused-function -Wno-error=missing-braces'
 
+  - &tag-filter
+    filters:
+      tags:
+        only: /^v.*/
+
 workflows:
   version: 2
 
   drafter:
     jobs:
       - lint
-      - test-gcc5.4
-      - test-gcc6
-      - test-gcc7
-      #- test-gcc8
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -29,16 +29,43 @@ workflows:
   drafter:
     jobs:
       - lint
-      - test-gcc5.4
-      - test-gcc6
-      - test-gcc7
+      - test-gcc5.4:
+          filters:
```
kylef: We can do something like the following with YAML references to de-duplicate the tag filter a little bit:

```yaml
aliases:
  - &tag-filter
    filters:
      tags:
        only: /^v.*/
```

Then:

```yaml
- test-gcc6: *tag-filter
- test-gcc7: *tag-filter
```

PR #603, opened 2018-09-10
JSON value s11n unit tests

tjanc:
Includes minor fixes; not all corner cases are covered by integration tests.

### DIFF
--- a/test/refract/test-JsonValue.cc
+++ b/test/refract/test-JsonValue.cc
@@ -1,0 +1,1477 @@
+//
+//  test/refract/test-JsonValue.cc
+//  test-librefract
+//
+//  Created by Thomas Jandecka on 17/02/2018
+//  Copyright (c) 2018 Apiary Inc. All rights reserved.
+//
+
+#include "catch.hpp"
+
+#include "refract/JsonValue.h"
+#include "refract/Element.h"
+#include "refract/ElementUtils.h"
+#include "utils/so/JsonIo.h"
+
+#include <sstream>
+
+using namespace drafter::utils;
+using namespace so;
+using namespace refract;
+
+namespace
+{
+    std::string to_string(const so::Value& v)
+    {
+        std::ostringstream ss{};
+        so::serialize_json(ss, v, so::packed{});
+        return ss.str();
+    }
+
+    template <typename Element>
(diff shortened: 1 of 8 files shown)


test/refract/test-JsonValue.cc
```diff
@@ -0,0 +1,1477 @@
...
+        {
+            auto result = generateJsonValue(*el);
+
+            THEN("the result is `42.1930e-39`")
+            {
+                REQUIRE(to_string(result) == "42.1930e-39");
+            }
+        }
+    }
+}
```
klokane: Please add tests for `Nan` and `Inf`

klokane: For NumberElement

klokane:
:+1: great coverage of testcases

test/refract/test-JsonValue.cc
tjanc: We use the JSON definition of number, `Nan` and `Inf` are not numbers.

klokane: But you can set value of NumberElement to Nan/Inf, isn't it?
There should be, at least, defined bahavior for this case

tjanc: You can theoretically construct a NumberElement with any string now, as the check for JSON Number is in `Serialize.cc`. But it will have undefined behaviour, so I'm not sure what to test against. Finding something non-numberish in a NumberElement could be handled as an `assert` in `JsonValue.cc`.

klokane: On other side refract is not strictly for JSON, so we should define what happens. e.g. conversion to zero?
Not sure, just idea

tjanc: There is the relevant issue on the MSON spec not being spec enough: https://github.com/apiaryio/mson/issues/93
In API Elements ("refract") we are pretty clear: https://github.com/apiaryio/api-elements/blob/master/docs/element-definitions.md#values
We could add a debug build run time check during JSON serialization (not JSON _value_ serialization), although I don't deem that necessary. Here are the tests currently covering your worries: https://github.com/apiaryio/drafter/blob/master/test/test-Serialize.cc#L931

klokane: Ok. Let it pass for now and create issue to resolve this

PR #653, opened 2018-10-02
Klokane/annotation with column line info

klokane:
Source code ready for review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Drafter Changelog
 
+## master
+
+* Add column/line info to anotations source maps 
+* change start column index from 1 to 0 in reporting. (while `-l` switch is used)
+
 ## 4.0.0-pre.1
 
 ### Breaking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,6 @@
 ## master
 
 * Add column/line info to anotations source maps 
-* change start column index from 1 to 0 in reporting. (while `-l` switch is used)
 
 ## 4.0.0-pre.1
 
--- a/src/ConversionContext.h
+++ b/src/ConversionContext.h
@@ -20,7 +20,7 @@ namespace drafter
     class ConversionContext
     {
         refract::Registry registry;
-        NewLinesIndex newLinesIndex;
+        const NewLinesIndex newLinesIndex;
 
     public:
         ConversionContext(const char* source, const WrapperOptions& options);
--- a/src/ConversionContext.h
+++ b/src/ConversionContext.h
@@ -10,6 +10,7 @@
 
 #include "refract/Registry.h"
 #include "snowcrash.h"
+#include "SourceMapUtils.h"
 
 namespace drafter
 {
@@ -19,8 +20,11 @@ namespace drafter
     class ConversionContext
     {
         refract::Registry registry;
+        NewLinesIndex newLinesIndex;
 
     public:
+        ConversionContext(const char* source, const WrapperOptions& options);
+
         const WrapperOptions& options;
         std::vector<snowcrash::Warning> warnings;
 
@@ -34,7 +38,10 @@ namespace drafter
             return registry;
         }
 
-        ConversionContext(const WrapperOptions& options) : options(options) {}
+        inline const NewLinesIndex& GetNewLinesIndex() const
+        {
+            return newLinesIndex;
+        }
 
         void warn(const snowcrash::Warning& warning);
     };
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -26,6 +26,8 @@
 #include "NamedTypesRegistry.h"
 #include "ConversionContext.h"
 
+#include "reporting.h"
+
 using namespace drafter;
 using namespace refract;
 using namespace drafter::utils::log;
@@ -613,7 +615,7 @@ std::unique_ptr<IElement> drafter::BlueprintToRefract(
 }
 
 std::unique_ptr<IElement> drafter::AnnotationToRefract(
-    const snowcrash::SourceAnnotation& annotation, const std::string& key)
+    const snowcrash::SourceAnnotation& annotation, const std::string& key, ConversionContext& context)
 {
     auto element = from_primitive(annotation.message);
 
@@ -622,7 +624,8 @@ std::unique_ptr<IElement> drafter::AnnotationToRefract(
     element->meta().set(SerializeKey::Classes, make_element<ArrayElement>(from_primitive(key)));
 
     element->attributes().set(SerializeKey::AnnotationCode, from_primitive(annotation.code));
-    element->attributes().set(SerializeKey::SourceMap, SourceMapToRefract(annotation.location));
+    element->attributes().set(
+        SerializeKey::SourceMap, SourceMapToRefractWithColumnLineInfo(annotation.location, context));
 
     return std::move(element);
 }
--- a/src/RefractAPI.cc
+++ b/src/RefractAPI.cc
@@ -26,8 +26,6 @@
 #include "NamedTypesRegistry.h"
 #include "ConversionContext.h"
 
-#include "reporting.h"
-
 using namespace drafter;
 using namespace refract;
 using namespace drafter::utils::log;
--- a/src/RefractSourceMap.cc
+++ b/src/RefractSourceMap.cc
@@ -1,4 +1,6 @@
 #include "RefractSourceMap.h"
+#include "reporting.h"
+#include "ConversionContext.h"
 
 using namespace refract;
 
@@ -10,6 +12,7 @@ namespace
             from_primitive(sourceMap.location),
             from_primitive(sourceMap.length));
     }
+
 } // namespace
 
 std::unique_ptr<IElement> drafter::SourceMapToRefract(const mdp::CharactersRangeSet& sourceMap)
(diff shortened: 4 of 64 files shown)


src/RefractAPI.cc
```diff
@@ -26,6 +26,8 @@
 #include "NamedTypesRegistry.h"
 #include "ConversionContext.h"
 
+#include "reporting.h"
```
klokane: remove - restful

src/RefractSourceMap.cc
```diff
@@ -1,4 +1,6 @@
 #include "RefractSourceMap.h"
+#include "reporting.h"
```
klokane: remove - restful

src/SourceMapUtils.h
```diff
@@ -0,0 +1,39 @@
...
+//
+//  Created by Jiri Kratochvil on 2018-10-02
+//  Copyright (c) 2018 Apiary Inc. All rights reserved.
+//
+
+#ifndef DRAFTER_SOURCEMAPUTILS_H
+#define DRAFTER_SOURCEMAPUTILS_H
+
+#include "SourceAnnotation.h"
+
```
klokane: add `namespace drafter`

src/reporting.h
```diff
@@ -11,6 +11,7 @@
 
 #include "drafter.h"
 #include "SourceAnnotation.h"
+#include "reporting.h"
```
klokane: remove - mistake

test/draftertest.h
```diff
@@ -103,10 +103,11 @@ namespace draftertest
 
     struct FixtureHelper {
 
-        static std::unique_ptr<refract::IElement> parseAndSerialize(
-            snowcrash::ParseResult<snowcrash::Blueprint>& blueprint, const drafter::WrapperOptions& options)
+        static std::unique_ptr<refract::IElement> parseAndSerialize(const std::string& source,
```
klokane: remove - place inside `handleResultJSON()`

tools/update-fixtures.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+
+for i in $(./bin/test-libdrafter | grep Filename | sed -e 's/^  Filename: //' | sed -e 's/\x1B\[[0-9;]*[JKmsu]//g')
```
kylef: You can already do `env GENERATE=1 make test` to update fixtures, was you unaware of that?

klokane: Oh, I remember there were some discussion related it is not working.
So I probably miss info it is working again.

Ok, so i will remove this script.

klokane:
@apiaryio/adt ready for review

kylef:
> The line and column info should be optional (a new option to drafter_parse) and it should be defaulted to false. @kylef Can you confirm?

In the Swagger 2 parser line and number is always there. I don't see any benefit of it being optional because this is something that most consumers want. Do you have some use-case in mind where you would want to disable line and column numbers for annotation source maps?

test/fixtures/syntax/issue-350.json
```diff
@@ -214,10 +254,30 @@
...
                     {
                       "element": "number",
+                      "attributes": {
+                        "line": {
+                          "element": "number",
+                          "content": 9
+                        },
+                        "column": {
+                          "element": "number",
+                          "content": 1
```
kylef: As per the spec, the column should be zero-indexed. Since this is empty line this should be 0?

https://api-elements.readthedocs.io/en/latest/element-definitions.html#source-map


---

Also, given the ADD I think the source map should not end on the start of line 9, but end of line 8? Maybe larger issue to deal with than in the scope of this PR though.

test/test-RefractDataStructureTest.cc
```diff
@@ -2,16 +2,6 @@
 
 using namespace draftertest;
 
-static drafter::WrapperOptions MSONTestOptions(false, true);
```
tjanc: How come this was removed?

test/fixtures/syntax/issue-350.json
klokane: I use algorithm which was already there for `-l` switch. So it is wrong in original code.
I will look at it

src/reporting.cc
```diff
@@ -134,7 +70,7 @@ namespace
                 if (useLineNumbers) {
 
                     AnnotationPosition annotationPosition;
-                    GetLineFromMap(linesEndIndex, *it, annotationPosition);
+                    ::GetLineFromMap(linesEndIndex, *it, annotationPosition);
```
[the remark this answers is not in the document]
klokane: Can be removed, that is rest after some "experiments" :)

test/test-RefractDataStructureTest.cc
klokane: it is moved into `test/draftertest.h`.
I made some refactoring of tests and remove restful parameter for macro and move similar macro definition into one file. But if you mean it will be better hold `TEST_MSON()` macro in this file, we can do it.

src/SourceMapUtils.h
```diff
@@ -0,0 +1,44 @@
+//
```
tjanc: I'm wondering why this code is not buried somewhere in an anonymous namespace in a source file.

klokane: Is is part of code whi was used in reporting. I did not wrote it jost move from another file to public space.
Reason to be publish it, is it is used in two parts - first is `reporting.cc` and second one is used in annotation line/column info

pksunkara:
> In the Swagger 2 parser line and number is always there. I don't see any benefit of it being optional because this is something that most consumers want. Do you have some use-case in mind where you would want to disable line and column numbers for annotation source maps?

When people need sourcemaps to manipulate stuff but it's not a user interface so they don't need line and column info.

klokane:
Oh, I realize code will not working correctly with `utf-8` because it count byte positions instead of char positions. Will going to fix it.

src/ConversionContext.h
```diff
@@ -34,7 +38,10 @@ namespace drafter
             return registry;
         }
 
-        ConversionContext(const WrapperOptions& options) : options(options) {}
+        inline const NewLinesIndex& GetNewLinesIndex() const
```
tjanc: Why `inline`? Fun fact: `inline` causes GCC to produce a weak symbol, that's all.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Drafter Changelog
 
+## master
+* Add column/line info to anotations source maps
```
tjanc: Suggesting to add change in indexing source maps (from 1 -> from 0).

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Drafter Changelog
 
+## master
+
+* Add column/line info to anotations source maps 
+* change start column index from 1 to 0 in reporting. (while `-l` switch is used)
```
pksunkara: Why is this done? This is UI. We need to use index starting from 1. @kylef What do you think?

pksunkara:
We still haven't made the column/line info optional

klokane:
Yes, we did not it optional.
There were discussion related to this and there was no decision make it optional

CHANGELOG.md
klokane: https://github.com/apiaryio/drafter/pull/653/files/3217ac2c099085b7bfd2640e1ad91751287697e2#r223059516

kylef: I think you mean the `-u` flag? 

>   -u, --use-line-num    use line and row number instead of character index when printing annotation

I don't think this is nessecerily done in the UI, it is an explanation that `-u` flag will now behave differently than before because the same logic is used for column/line numbers.

test/fixtures/syntax/issue-350.json
tjanc: Actually, I don't see where in the spec the line or column is said to be zero-indexed. Only the byte offset of the source map is said to be zero-indexed. The `column` and `line` attributes are actually very undefined per spec.

kylef: Hmm, you are right. We should clarify in the spec. Should we go for 1-indexed or 0-indexed?

CHANGELOG.md
tjanc: > https://github.com/apiaryio/drafter/pull/653/files/3217ac2c099085b7bfd2640e1ad91751287697e2#r223059516

Based on my reading of the spec Pavan raised a valid point. There is actually no need to change to zero-based indexing for column and line attributes of source maps; I initially also misread the spec, because it talks about zero-based indexing - of the byte offset however, not column and line numbers. All in all, column and line numbers are good old natural numbers, and I would keep that. Natural numbers, in my world, start from 1.

test/fixtures/syntax/issue-350.json
tjanc: https://github.com/apiaryio/drafter/pull/653#discussion_r232318730
> column and line numbers are good old natural numbers, and I would keep that. Natural numbers, in my world, start from 1.

CHANGELOG.md
klokane: Ok, so do you want to return back implementation to one-indexed like in original code?

I made unit tests related to indexing 25 days ago. It was clear from this code how indexing will be implemented. I asked for review, nobody commented it...
https://github.com/apiaryio/drafter/pull/653/commits/f255865e30db7ef40c1ef55d7681858b690e1f02

PR #657, opened 2018-10-31
Initial CMake configuration

tjanc:
Adds support for CMake. 

The following should work across platforms:
```
git clone --recursive https://github.com/apiaryio/drafter tjanc/cmake-targets
cd drafter
mkdir build
cd build
cmake .. -DCMAKE_INSTALL_PREFIX=dir/you/install/to
make
make test
make install
```
`CMAKE_INSTALL_PREFIX` defaults to your root when not set.
The resulting install should look like this:
```
dir/you/install/to
├── bin
│   └── drafter
├── include
│   └── drafter.h
└── lib
    ├── cmake
    │   └── drafter
    │       ├── drafter-config.cmake
    │       ├── drafter-config-version.cmake
    │       ├── drafter-targets.cmake
    │       └── drafter-targets-noconfig.cmake
    ├── libdrafter.a
    ├── libdrafter-pic.a
    └── libdrafter.so

```

A few notes:
- da7df98, 59782dd: Catch2 exports its headers from the `single_header` directory, which was not what we were using; I fixed our GYP config to match Catch2's CMake-exported include directories and updated headers in our tests
- 54c2f33: some warnings popped up after fixing headers, fixed those
- currently, the location of drafter's CMake config files are in `src` and `test`; that's not ideal, but until libdrafter has its own directory to live in it's the best we can do
- updating CI to use cmake is not in scope of this PR
- `CMakeLists.txt` in root just includes subprojects and ensures invocations of `find_package` in subprojects are not resolved into system paths

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,48 @@
+cmake_minimum_required(VERSION 3.4 FATAL_ERROR)
+
+message(STATUS "Testing dependencies ${BUILD_TESTING_DEPENDENCIES}.")
+
+set(CMAKE_PREFIX_PATH "/prefix")
+set(CMAKE_SKIP_INSTALL_ALL_DEPENDENCY TRUE)
+set(as_subproject
+    markdown-parser
+    snowcrash
+    Sundown
+    cmdline
+    Drafter
+    Catch2
+    dtl
+    BoostContainer)
+
+macro(find_package)
+    if(NOT "${ARGV0}" IN_LIST as_subproject)
+        _find_package(${ARGV})
+    endif()
+endmacro()
+
+# add external dependencies living in `ext`
+add_subdirectory(ext/snowcrash/ext/markdown-parser/ext/sundown EXCLUDE_FROM_ALL)
+add_subdirectory(ext/cmdline EXCLUDE_FROM_ALL)
+add_subdirectory(ext/boost_1_66_0 EXCLUDE_FROM_ALL)
+
+message(STATUS "Setting up tests for production...")
+set(BUILD_TESTING ON)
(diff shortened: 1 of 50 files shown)


tjanc:
Had to touch last commit to restart CI due to SSL failure in appveyor.

kylef:
I'm seeing a 5x slowdown running the tests on this branch via Cmake. Do you have any idea what the cause could be? Do you see the same on Linux?

Cmake:

(ran multiple times seeing similiar timing)

```
$ make test
100% tests passed, 0 tests failed out of 874

Total Test time (real) =  15.15 sec
```

GYP:

```
$ time make test
...

./bin/test-libmarkdownparser
===============================================================================
All tests passed (448 assertions in 23 test cases)

./bin/test-libsnowcrash
===============================================================================
All tests passed (5449 assertions in 478 test cases)

./bin/test-libdrafter
===============================================================================
All tests passed (2445 assertions in 373 test cases)

./bin/test-capi
        2.55 real         2.25 user         0.18 sys
```

tjanc:
You are right, my configuration ran every test in a separate process, incurring visible overhead before and after running each test.

1. 79682a2 addresses this by running the three test executables as a whole instead of running each unit test in isolation. On my system the result is slightly faster than with Gyp.
2. 98c87d3 optimizes time even further, decoupling the tests and allowing parallel execution of the three through `ctest -j<jobs>`; on my system that shaves off another 10-20%

Tips:
- `cmake .. -DCMAKE_BUILD_TYPE=Release` gets you an optimized build
- `ctest -V` runs tests showing the test's output(s), given you are in the build directory

PR #658, opened 2018-11-01
fix: Default in enum parameters should be enum element

pksunkara:


kylef:
> The default does not have to be wrapped into an enum, and probably it shouldn't.

Yes, I though we moved away from wrapping in enum. The content, sample (value) and default value should be an element themselves. Such as `{"element": "string", "content": "hello"}`.

tjanc:
@pksunkara I'm OK to merge this to unify parameters and MSON behavior if you think the real fix would be too complicated at this stage; but only if we file a bug report and list it as a known bug in the release notes. What do you think?

kylef:
> We haven't. The MSON currently wraps it as enum and minim serialization depends on it. In this line, you wouldn't need [defaultValue.content] if we are not relying on the wrap. You could have done [defaultValue]

The linked source is for 0.6 serialisation (converting 1.0 into 0.6), in 1.0, the default value should be an Element as per https://api-elements.readthedocs.io/en/latest/element-definitions.html#enum-element:

> default (Element) - Default value for this Element; type of default MUST match the type this Element describes

kylef:
Perhaps we can get https://github.com/apiaryio/drafter/pull/653 merged before this so the release contains both changes.

PR #661, opened 2018-11-16
Fix untested error.apib fixture

kylef:
The fixture `error.apib` was untested and incorrect which this PR resolves.

### DIFF
--- a/test/fixtures/parse-result/error.json
+++ b/test/fixtures/parse-result/error.json
@@ -30,10 +30,30 @@
                   "content": [
                     {
                       "element": "number",
+                      "attributes": {
+                        "line": {
+                          "element": "number",
+                          "content": 4
+                        },
+                        "column": {
+                          "element": "number",
+                          "content": 1
+                        }
+                      },
                       "content": 24
                     },
                     {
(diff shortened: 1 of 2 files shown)


test/fixtures/parse-result/error.json
```diff
@@ -30,10 +30,30 @@
...
                     {
                       "element": "number",
+                      "attributes": {
+                        "line": {
+                          "element": "number",
+                          "content": 4
+                        },
+                        "column": {
+                          "element": "number",
+                          "content": 1
```
kylef: The byte range is `start 24, length 1` but the column and line number are indentical which seems incorrect. I'd expect this to end on on column 2 given the source map /c @klokane

klokane: We should make a clear column addressing. In my opinion it is ok. We are addressing one character right? 

It is typical problem of open interval vs closed interval.

There is question "what is last character in addressed space?" Should we point on to last character, or should we point after last character? It is similar to addressing index in array. I mean we are not addressing in meaning of range, but we are addressing first character and last character.

We probably should clear this in spec

PR #674, opened 2019-03-19
Disable integration tests on windows CI

kylef:
These are currently failing on master, attempting to resolve them in https://github.com/apiaryio/drafter/pull/673. Disabling integration tests in interim so we're not blocking development while we fix Aruba.


tjanc:
Or we could just use plain old bash scripts to run drafter instead of taking in dependencies on ruby.

PR #679, opened 2019-04-03
Replace `variant` implementation

tjanc:
C++17 introduces `std::variant`. We have made the effort to implement this data structure ourselves, supporting C++14.

Now that we are forced to go back to C++11, we replace our implementation with mpark/variant, which supports said standard and provides the exact same interface and behavior as `std::variant` from C++17. Once drafter allows C++17, this dependency can therefore be easily dropped.


src/refract/JsonSchema.cc
```diff
@@ -655,7 +654,7 @@ namespace
     so::Array* findAnyOf(so::Object& schema)
     {
         if (so::Value* anys = so::find(schema, "anyOf"))
-            return &drafter::utils::get<so::Array>(*anys);
+            return mpark::get_if<so::Array>(anys);
```
opichals: `::get_if` vs `::get` is that ok?

PR #680, opened 2019-04-03
Go back to c++11

tjanc:
To support some old compilers, we are forced to go back to C++11.

Notably, this branch compiles with gcc>=4.8.3 and runs with respective libc; tested on docker image `lambci/lambda:build-nodejs8.10`.


tjanc:
> There is question if we should not provide test to REGEX replacement.

The changed passages are already unit tested, no?

Issue #682, opened 2019-04-10
Drafter aborts while rendering JSON value for a non-object inside an array of objects


kylef:
The included API Blueprint caused the assertion at https://github.com/apiaryio/drafter/blob/bb00117c0f96fecc07a89dfc33ebbdd002164eba/src/refract/JsonValue.cc#L95 to be hit:

```
$ ./bin/drafter <document>
Assertion failed: (false), function renderProperty, file ../src/refract/JsonValue.cc, line 95.
fish: './bin/drafter <document>' terminated by signal SIGABRT (Abort)
```

```apib
## GET /

+ Response 200 (application/json)
    + Attributes (array[object])
        + Thing
```

Drafter 4.0.0-pre.3

kylef:
Drafter 3.2.7 succeeds and returns the following JSON body:

```json
[
  {
    "": "Thing"
  }
]
```

It does not produce any error.

kylef:
Another case, perhaps not same bug but it seems related:

```apib
# GET /

+ Response 200 (application/json)
    + Attributes (Array)
        + (object)

## Data Structures

### Array (array)
```

```
Assertion failed: (hasValue_), function get, file ../src/refract/Element.h, line 75.
fish: './bin/drafter 682-1.apib' terminated by signal SIGABRT (Abort)
```

tjanc:
> Another case, perhaps not same bug but it seems related:
> 
> ```
> # GET /
> 
> + Response 200 (application/json)
>     + Attributes (Array)
>         + (object)
> 
> ## Data Structures
> 
> ### Array (array)
> ```
> 
> ```
> Assertion failed: (hasValue_), function get, file ../src/refract/Element.h, line 75.
> fish: './bin/drafter 682-1.apib' terminated by signal SIGABRT (Abort)
> ```

It's independent. See #683

PR #684, opened 2019-04-24
Set NDEBUG on release builds

kylef:
Verified using example from #682.

Before:

```
$ ./bin/drafter 682.apib -l
Assertion failed: (false), function renderProperty, file ../src/refract/JsonValue.cc, line 95.
fish: './bin/drafter bb.apib -l' terminated by signal SIGABRT (Abort)
```

After:

```shell
$ ./bin/drafter 682.apib -l

OK.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
     + MSON string now interpreted as String Element instead of Object Element
   ```
 
+* Release builds of Drafter will no longer include assertions, this prevents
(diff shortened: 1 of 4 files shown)


kylef:
@tjanc this will cause variable not used warnings (and in CI errors) due to things such as:

https://github.com/apiaryio/drafter/blob/bb00117c0f96fecc07a89dfc33ebbdd002164eba/src/refract/dsd/Traits.h#L198

Wondering what the best way to solve that might be

PR #690, opened 2019-05-07
Fix: entries in EnumElement are expanded

### LINKED ISSUE #689: Unresolved reference in EnumElement after expansion
The following blueprint is not interpreted correctly during expansion:
```
# API name

# Group Visits

## Visits [/visits]

### Create a Visit [POST]

    + Attributes (Visit)

# Data Structures

## ContactInfo (object)

+ email: june@email.com (string)

## Visit (object)

+ monitor (enum)
    + (object)
        + Include ContactInfo
```
An unresolved RefElement `ContactInfo` is still present in the backend, not matching assumptions and causing a crash at best.

tjanc:
Fixes #689


test/fixtures/mson/issue-689.apib
```diff
@@ -0,0 +1,22 @@
...
+### Create a Visit [POST]
+
++ Response 200 (application/json)
+    + Attributes (Visit)
+
+# Data Structures
+
+## ContactInfo (object)
+
++ email: june@email.com (string)
```
kylef: Please use `example.com`, it's a reserved domain name, `june@email.com` could be a legit persons email address which we probably don't want to use in our tests.

[from RFC2606](https://tools.ietf.org/html/rfc2606)

```suggestion
+ email: user@example.com (string)
```

PR #695, opened 2019-05-08
Test on CI with gyp on GCC 4.8

kylef:
GCC 4.8 is the oldest compiler we support with gyp (in Protagonist) and we've introduced compiler support regressions such as https://github.com/apiaryio/drafter/pull/691. This can be avoided in the future if we CI against this version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -105,7 +105,7 @@ jobs:
   test-gyp:
     <<: *test-base-legacy
     docker:
-      - image: gcc:5.4
+      - image: gcc:4.8.5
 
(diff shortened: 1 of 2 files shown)


kylef:
@tjanc any ideas why this would fail on CI? Yet the same code compiles on GCC 4.8.4 elsewhere with Protagonist on Travis CI.

```
../src/refract/JsonSchema.cc:678:55: error: missing initializer for member '{anonymous}::ObjectSchema::properties' [-Werror=missing-field-initializers]
                             ObjectSchema optionSchema{};
```

PR #700, opened 2019-05-13
Fix formatting of JsonSchema

kylef:
The lack of `{` `}` around an if-else statement caused clang-format to get confused and incorrectly indent subsequent functions which makes the file hard to follow.

I was a little confused as to why `renderSchema` for object had the logic for array and other types embedded until I realised that it does not, just the code is formatted strangely.

### DIFF
--- a/src/refract/JsonSchema.cc
+++ b/src/refract/JsonSchema.cc
@@ -400,394 +400,393 @@ namespace
         addType(schema, TYPE_NAME);
 
         ObjectSchema result{};
+        if (e.empty()) {
+            LOG(warning) << "empty data structure element in backend";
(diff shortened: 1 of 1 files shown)


src/refract/JsonSchema.cc
```diff
@@ -400,394 +400,393 @@ namespace
         addType(schema, TYPE_NAME);
 
         ObjectSchema result{};
-        if (e.empty())
+        if (e.empty()) {
```
kylef: The only change I made was adding `{` here, replacing the `else` with `} else {`, and then adding `}` on line 414 and then running clang-format on the file.

PR #701, opened 2019-05-13
Fix "fixed" type attribute not inheriting for empty objects

kylef:
For some reason, the logic for inheriting "fixed" requires an object to have a value. An empty object should inherit fixed too.

### DIFF
--- a/src/refract/ElementUtils.cc
+++ b/src/refract/ElementUtils.cc
@@ -38,7 +38,7 @@ bool refract::inheritsFixed(const IElement& e)
 
 bool refract::inheritsFixed(const ObjectElement& e)
 {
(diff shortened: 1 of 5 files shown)


src/refract/ElementUtils.cc
```diff
@@ -38,7 +38,7 @@ bool refract::inheritsFixed(const IElement& e)
 
 bool refract::inheritsFixed(const ObjectElement& e)
 {
-    return definesValue(e);
```
kylef: @tjanc is there any reason why this logic requires fixed? I cannot find any cases this change would break, there are no unit/integration tests which cover the case of empty objects and inheriting fixed type attribute.

PR #712, opened 2019-06-03
Add asan & lsan to CI, fix reported issues

tjanc:


ext/snowcrash/ext/markdown-parser/src/MarkdownNode.h
```diff
@@ -82,13 +82,15 @@ namespace mdp
             const Data& data_ = Data());
 
         /** Copy constructor */
-        MarkdownNode(const MarkdownNode& rhs);
+        MarkdownNode(const MarkdownNode& rhs) = delete;
+        MarkdownNode(MarkdownNode&& rhs) = default;
 
         /** Assignment operator */
-        MarkdownNode& operator=(const MarkdownNode& rhs);
+        MarkdownNode& operator=(const MarkdownNode& rhs) = delete;
```
pksunkara: What does this `= delete` and `= default` do?

ext/snowcrash/ext/markdown-parser/src/MarkdownParser.cc
```diff
@@ -258,8 +254,8 @@ void MarkdownParser::renderHTML(const ByteBuffer& text)
     if (!m_workingNode)
         throw NO_WORKING_NODE_ERR;
 
-    MarkdownNode node(HTMLMarkdownNodeType, m_workingNode, text);
-    m_workingNode->children().push_back(node);
+    MarkdownNode node();
```
pksunkara: You don't need this line.

ext/snowcrash/ext/markdown-parser/src/MarkdownNode.h
tjanc: * `MarkdownNode(MarkdownNode&& rhs) = default;` makes the compiler implement the default move constructor for `MarkdownNode`. The default move constructor for structures is what you would expect: move every member of `rhs` to the respective member of the constructed structure.
* `MarkdownNode(const MarkdownNode& rhs)` explicitly denies any declaration of said function; effectively prohibiting copy construction

I figured we don't need to deep-copy our MarkdownNode tree, it made work with ASAN hard because it propagated invalid values from uninitialized memory. Also no reason to deep-copy in our code base.

tjanc:
> I don't see lsan. Otherwise looks good.

LSAN is run by ASAN.

PR #714, opened 2019-06-18
Fix duplicity in generated scheme with swapped order of member object wit `fixed-type`

### LINKED ISSUE #664: De-duplicate equal object entries in schema "anyOf" field for fixed-type array
Let's assume, I have the next apib:
```
# Group Example
## GET /

   + Attributes (array, fixed-type)
       + (object)
           + msg: 'hello' (string)
           + id: 2 (number)
       + (object)
           + msg: 'bye' (string)
           + id: 3 (number)
```

And for this input I get such JSON Schema:
```
$ ./build/Release/drafter.exe -f json simple.apib | ./tools/refract-filter.py
{
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "object",
        "properties": {
          "msg": {
            "type": "string"
          },
          "id": {
            "type": "number"
          }
        }
      }
    ]
  }
}
```

This is fine. I have equal structures as array's nested members and field "anyOf" contains only one member.

The next step: I swap the places of the second object fields:
```
# Group Example
## GET /

   + Attributes (array, fixed-type)
       + (object)
           + msg: 'hello' (string)
           + id: 2 (number)
       + (object)
           + id: 3 (number)
           + msg: 'bye' (string)
```

AFAIK, in objects, there's no difference in fields order, these two objects should be equal.
But `emplace_unique` (or whatever doing the job) cannot handle this case:
```
{
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "object",
        "properties": {
          "msg": {
            "type": "string"
          },
          "id": {
            "type": "number"
          }
        }
      },
      {
        "type": "object",
        "properties": {
          "id": {
            "type": "number"
          },
          "msg": {
            "type": "string"
          }
        }
      }
    ]
  }
}
```

What I expected to see: the same output as the first JSON Schema.

FYI, I have the latest version (4.0.0-pre2)

I think this has some relation to #566

klokane:
Closes #664

### DIFF
--- a/src/utils/so/Value.cc
+++ b/src/utils/so/Value.cc
@@ -37,9 +37,26 @@ bool drafter::utils::so::operator==(const Number& lhs, const Number& rhs)
     return lhs.data == rhs.data;
 }
 
+namespace
+{
+    bool hasEqualMembers(const Object& lhs, const Object& rhs)
+    {
+        using itemType = typename Object::container_type::const_reference;
+        for (itemType item : lhs.data) {
+
+            if (std::find_if(rhs.data.begin(), rhs.data.end(), [&item](itemType rItem) { return item == rItem; })
+                == rhs.data.end()) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+}
+
 bool drafter::utils::so::operator==(const Object& lhs, const Object& rhs)
 {
-    return lhs.data.size() == rhs.data.size() && std::equal(lhs.data.begin(), lhs.data.end(), rhs.data.begin());
+    return lhs.data.size() == rhs.data.size() && hasEqualMembers(lhs, rhs);
 }
 
 bool drafter::utils::so::operator==(const Array& lhs, const Array& rhs)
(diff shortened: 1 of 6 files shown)


src/utils/so/Value.cc
```diff
@@ -37,9 +37,26 @@ bool drafter::utils::so::operator==(const Number& lhs, const Number& rhs)
     return lhs.data == rhs.data;
 }
 
+namespace
+{
+    bool hasEqualMembers(const Object& lhs, const Object& rhs)
+    {
+        using itemType = typename Object::container_type::const_reference;
+        for (itemType item : lhs.data) {
```
kylef: If I understand this logic correctly, if there is the following I think they'll be treated as equal.

LHS:

```
+ name: doe
```

RHS:

```
+ name: doe
+ additional: true
```

Since the logic loops over lhs members and verifies they are equal to rhs. Additional RHS members are ignored.

Perhaps a guard that the length of lhs and rhs are equal at the top of the function would solve this problem?

klokane: There is on upper level checked count of members is equal. 

https://github.com/apiaryio/drafter/pull/714/files#diff-c71b7d4a5846e44057ccc118bacbe934R59

So situation you describe should not happen. But if you like I can make it in one step if you like

kylef: @klokane you're right, I overlooked that part.

PR #715, opened 2019-06-19
Update boost to 1.70.0

kylef:
Firstly, I've moved boost into `ext/boost` instead of `ext/boost_1_66_0` so that we don't change directory everytime it's moved. There is a file `ext/boost/VERSION` which specifies the version used.

This will prevent extra diff noise, and having to change build configuration when we change boost version.

I've also included the LICENSE file from boost.


tjanc:
You used `bcp` in the end, correct?
```
bcp --boost=~/Downloads/boost boost/container/vector.hpp ext/boost
```

kylef:
That's right, so the following can be used in future:

```shell
$ rm -fr ext/boost/boost
$ bcp --boost=/Users/kyle/Downloads/boost_1_70_0 boost/container/vector.hpp ext/boost
$ git add ext/boost/boost
```

PR #717, opened 2019-06-21
(fix): referencing undefined named type in type definition

klokane:


src/RefractDataStructure.cc
```diff
@@ -456,24 +456,36 @@ namespace
...
 
                     if (typeName == mson::UndefinedTypeName && !type.symbol.literal.empty()) {
                         typeName = GetMsonTypeFromName(type.symbol.literal, context);
                         method = type.symbol.variable ? eSample : eElement;
+
+                        if (typeName == mson::UndefinedTypeName
+                            && method == eElement) { // undefined named element in hinting
+                            std::stringstream msg;
+                            msg << "Undefined named type '" << type.symbol.literal << "' referenced in type definition";
+                            context.warn(snowcrash::Warning(
```
kylef: Should this be an error instead? We're producing an invalid parse result otherwise because there will still be the reference to the symbol, from `test/fixtures/render/issue-556.json`:

```json
{
  "element": "Unknown"
}
```

klokane: Not sure if error, because type hinting is just info, content in array is  not mandatory. 
We can avoid creating element for referenced element

kylef: We treat the following as an error also:

```apib
# API
## GET /

    + Attributes
        + property (Unknown)
```

```
error: (4)  base type 'Unknown' is not defined in the document; line 6, column 11 - line 6, column 29
```

klokane: We can continue in parsing, so I vote for warning

kylef: Yes, can we make it so that `{ element: "Unknown" }` goes away from the parse result in the case of warning?

klokane: simple change, just `continue` in condition. We can discuss it in stand up. 
If "err && stop" or "warn && remove referenced named element"

klokane: Fixed

klokane: Remove undefined from result && produce warning

PR #726, opened 2019-07-04
Use variant for mson::Element

tjanc:
Refactors mson::Element to use `variant` instead of a structure containing members for every type that could be stored in it. There are at least three more opportunities for improvement like this in `MSON.h` to be refactored in the future, matching this PR.

## Principle
Let's say there is structure `Content` that can hold either `Foo` or `Bar` (but never both, i.e. a Σ-type). Currently, we use something like:
```
enum ContentKlass { UndefinedKlass, FooKlass, BarKlass };
struct Content {
  ContentKlass klass;
  Foo foo;
  Bar bar;
};
```
Much better, we can write:
```
class Content {
  std::variant<std::monostate, Foo, Bar> content;
public:
  const auto& foo() const noexcept { return std::get<Foo>(content); }
  const auto& bar() const noexcept { return std::get<Foo>(content); }
  template<typename F> void visit(F&& f) {
    std::visit(std::forward<F>(f), content);
  }
};
```

## Benefits
* better approximation of our model (MSON)
* avoid initialization of members that are not used
* avoid manual switches and conditionals in code working with the MSON AST, and
* let the compiler ensure every code path on the structure is covered
* more functional code
* faster dispatch (variant produces a vtable)


ext/snowcrash/src/MSON.h
```diff
@@ -29,6 +31,36 @@
...
+        const T& operator*() const noexcept
+        {
+            return content;
+        }
+
+        T& operator*() noexcept
+        {
+            return content;
+        }
+    };
```
tjanc: Both `OneOf` and `Group` sections are typedef'ed to a vector of `Element`, so the only way to disambiguate based on type (e. g. when visiting the variant) is by introducing seperate types (symbols). Above template achieves that.

ext/snowcrash/src/MSON.h
```diff
@@ -197,7 +229,7 @@ namespace mson
     struct Element;
 
     /** Collection of elements */
-    typedef std::vector<Element> Elements;
+    typedef boost::container::vector<Element> Elements;
```
tjanc: I changed this to `boost::container::vector`, just because it allows usage before a contained type is defined. In the past, `Element` was forced to hold a `std::vector<std::unique_ptr<Element>>`, now we hold `vector<Element>` directly.

src/RefractDataStructure.cc
```diff
@@ -1289,76 +1336,188 @@ namespace
...
+    {
+        auto option = make_element<OptionElement>();
+        MsonElementsToRefract( //
+            MakeNodeInfo(*section, sourceMap->elements()),
+            std::back_inserter(option->get()),
+            context);
+        return option;
+    }
+
+    struct MsonOneOfEntryToElementsLambda {
```
tjanc: This can be a much shorter lambda in C++14. In C++11 we don't have a choice but introducing a structure with a template operator().

src/RefractDataStructure.cc
```diff
@@ -1289,76 +1336,188 @@ namespace
...
+        ConversionContext& context,
+        const mson::BaseTypeName)
+    {
+        // TODO avoid throwing exceptions...
+        throw snowcrash::Error("unable to handle element group", snowcrash::ApplicationError);
+    }
 
-            default:
-                throw snowcrash::Error("unknown type of mson element", snowcrash::ApplicationError);
+    struct MsonTypeSectionToElementsLambda {
```
tjanc: This can be a much shorter lambda in C++14. In C++11 we don't have a choice but introducing a structure with a template operator().

src/NamedTypesRegistry.cc
```diff
@@ -22,336 +22,380 @@
...
-            {
-                return ds->name.symbol.literal;
-            }
+    void collectElementMember(const mson::Element::Empty& el, Members& result) {}
 
-            const std::string& name(const mson::TypeSpecification& ts) const
-            {
-                return ts.name.symbol.literal;
-            }
+    struct CollectElementMemberLambda {
```
tjanc: This can be a much shorter lambda in C++14. In C++11 we don't have a choice but introducing a structure with a template operator().

src/RefractDataStructure.cc
```diff
@@ -1289,76 +1336,188 @@ namespace
...
 
-            case mson::Element::MixinClass:
-                return MsonMixinToRefract(MakeNodeInfo(mse.node->content.mixin, mse.sourceMap->mixin));
+        for (const auto& oneOfInfo : oneOfNodeInfo) {
+            oneOfInfo.node->visit(MsonOneOfEntryToElementsLambda{ sourceMap, context, *select });
+        }
 
-            case mson::Element::OneOfClass:
-                return MsonOneofToRefract(MakeNodeInfo(mse.node->content.oneOf(), mse.sourceMap->oneOf()), context);
+        return std::move(select);
```
klokane: some special reason for `return std::move()` generally it is better to avoid because of copy elision

https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization

klokane: Oh it seem, it is rest of some older changes.
Please remove it separate commit in whole file

src/NamedTypesRegistry.cc
```diff
@@ -22,336 +22,380 @@
 #include <iostream>
 #endif /* DEBUG_DEPENDENCIES */
 
+using namespace drafter;
 using namespace refract;
 
-namespace drafter
+namespace
```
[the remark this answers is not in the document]
klokane: But code itself seems ok for me

src/RefractDataStructure.cc
tjanc: I believe gcc 4.8.5 complained (incomplete implementation of return value elision). Cool that you have noticed this though.

kylef: Maybe we can wrap it in a macro so that we can prevent move in modern compilers (and also prevent compiler warnings in newer compilers):

```c
#if GCC_VERSION > 4.9
#define MOVE(x) x
#else
#define MOVE(x) std::move(x)
#endif

return MOVE(x)
```

src/RefractDataStructure.cc
```diff
@@ -35,6 +35,57 @@ using namespace refract;
...
+        ConversionContext&,
+        const mson::BaseTypeName);
+
+    std::unique_ptr<IElement> MsonToElement( //
+        const mson::Element&,
+        const snowcrash::SourceMap<mson::Element>*,
+        ConversionContext&,
+        const mson::BaseTypeName);
+}
+
```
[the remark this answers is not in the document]
tjanc: I'm not convinced that would help readability. Those declarations are pretty straight forward, although they could live in a separate header/src.

src/RefractDataStructure.cc
```diff
@@ -1289,76 +1336,188 @@ namespace
...
+{
+    std::unique_ptr<OptionElement> MsonOneOfEntryToElements(
+        const mson::Element::Empty&, const snowcrash::SourceMap<mson::Element>*, ConversionContext&)
     {
+        throw snowcrash::Error("unknown type of mson element", snowcrash::ApplicationError);
+    }
 
-        NodeInfoCollection<mson::OneOf> oneOfNodeInfo(oneOf);
-        auto select = oneOfNodeInfo.empty() ? make_empty<SelectElement>() : make_element<SelectElement>();
+    std::unique_ptr<OptionElement> MsonOneOfEntryToElements(const mson::Element::PropertyMemberSection& section,
```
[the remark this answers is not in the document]
tjanc: I found the three lines in the bodies not to warrant an extra function. I can introduce it, but I don't know how to name it without confusion. Any idea?

pksunkara:
Looks good to me.

ext/snowcrash/src/Just.h
```diff
@@ -0,0 +1,48 @@
...
+//  Created by Thomas Jandecka on 11/7/19.
+//  Copyright (c) 2019 Apiary Inc. All rights reserved.
+//
+
+#ifndef SNOWCRASH_JUST_H
+#define SNOWCRASH_JUST_H
+
+#include <utility>
+
+namespace mson
```
klokane: Do you mean, `mson` ns is good place for `just`? I mean it is generalized concept so it should be in some  `util` ns or something similar.

ext/snowcrash/src/MSONParameterParser.h
```diff
@@ -118,17 +118,19 @@ namespace snowcrash
                             valueMemberSM = typeSection.sourceMap.elements().collection.at(i).value;
                         }
 
-                        if (valueMember)
-                            if (valueMember->valueDefinition.values.size() == 1) {
-                                out.node.values.push_back(valueMember->valueDefinition.values[0].literal);
+                        if (!valueMember)
+                            break;
```
klokane: `break`? I mean, there should be `continue`, because next element can be correctly *value member*
Also this should be after loop begin (nearly line 115)

src/MsonOneOfSectionToApie.cc
```diff
@@ -15,6 +15,16 @@
 using namespace refract;
 using namespace drafter;
 
+namespace
```
klokane: :+1:

src/RefractDataStructure.cc
tjanc: In hindsight - I was wrong. The compiler was right. A better way to avoid copies is to forbid them.

That's not in the scope of this PR though, so I'm removing the `move`s relying on optional RVO in compilers; it's around since C++98 and from C++17 onwards, RVO is even required in most of these cases.

PR #725, opened 2019-07-01
Print node debug to stderr

kylef:
Fixes #676

I've also removed the function unless we're in a debug build, the printNode callsite is wrapped in same check already.

PR #727, opened 2019-07-04
Remove markdown-parser debug logging

kylef:
As per https://github.com/apiaryio/drafter/pull/725#pullrequestreview-258008645, this probably isn't usful. Tools like lldb/gdb exist which can make debugging experiance better.

### DIFF
--- a/ext/snowcrash/ext/markdown-parser/src/MarkdownNode.cc
+++ b/ext/snowcrash/ext/markdown-parser/src/MarkdownNode.cc
@@ -6,6 +6,10 @@
 //  Copyright (c) 2014 Apiary Inc. All rights reserved.
 //
 
+#ifdef DEBUG
+#include <iostream>
+#endif
+
 #include "MarkdownNode.h"
 
 using namespace mdp;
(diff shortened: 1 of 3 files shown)


ext/snowcrash/ext/markdown-parser/src/MarkdownNode.cc
```diff
@@ -78,68 +78,3 @@ const MarkdownNodes& MarkdownNode::children() const
 
     return *m_children;
 }
-
-void MarkdownNode::printNode(size_t level) const
```
tjanc: I would keep the definition, but perhaps in DEBUG only.

ext/snowcrash/ext/markdown-parser/src/MarkdownNode.h
```diff
@@ -91,9 +90,6 @@ namespace mdp
         /** Destructor */
         ~MarkdownNode();
 
-        /** Prints the node to the stdout */
-        void printNode(size_t level = 0) const;
```
tjanc: The symbol can stay, always.

ext/snowcrash/ext/markdown-parser/src/MarkdownParser.cc
```diff
@@ -57,10 +57,6 @@ void MarkdownParser::parse(const ByteBuffer& source, MarkdownNode& ast)
     m_source = NULL;
     m_sourceLength = 0;
     m_listBlockContext = false;
-
-#ifdef DEBUG
-    ast.printNode();
-#endif
```
tjanc: Exactly, this is the only lines I would actually delete.

kylef:
@tjanc ok, I've updated the PR to just remove `printNode`, but I've used it from my prior PR so any printNode ends up on stderr instead.

PR #729, opened 2019-07-10
Windows cmake build

tjanc:
In Visual C++ 2015 Build Tools:
```
drafter> mkdir build
drafter> cd build
drafter/build> cmake .. -DCMAKE_INSTALL_PREFIX=..\install -DCMAKE_BUILD_TYPE=Release
drafter/build> cd ..
drafter> cmake --build build -t drafter-cli --config Release
drafter> cmake --build install -t drafter-cli --config Release
```
After a successful build, a standalone `drafter.exe` is located in `install/bin`.

### Notes
* can be improved without too much effort by generating a Windows installer via CPack
* introduction of a CTest dashboard script is not in the scope of this PR
* debug build fails to load the debug version of `ucrtbase.dll`, called `ucrtbased.dll`; that's expected, as Visual C++ 2015 Build Tools don't ship the debug version

### DIFF
--- a/src/utils/Utf8.h
+++ b/src/utils/Utf8.h
@@ -213,7 +213,8 @@ namespace drafter
 
                 template <typename Container>
                 explicit constexpr input_iterator(const Container& last) noexcept //
-                    : input_iterator(last.begin(), last.end())
+                    : e_(last.end()),
+		      cp_and_next_(decode_one(last.begin(), last.end()))
                 {
                 }
 
--- a/src/utils/Utf8.h
+++ b/src/utils/Utf8.h
@@ -214,7 +214,7 @@ namespace drafter
                 template <typename Container>
                 explicit constexpr input_iterator(const Container& last) noexcept //
                     : e_(last.end()),
-		      cp_and_next_(decode_one(last.begin(), last.end()))
+                      cp_and_next_(decode_one(last.begin(), last.end()))
                 {
                 }
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.6 FATAL_ERROR)
+project(DrafterProjects)
 
 include(DefaultBuildType.cmake)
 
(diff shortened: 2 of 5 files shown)


kylef:
I can confirm these changes fix the problems I was facing, I am able to build using:

```shell
$ mkdir build
$ cd build
$ cmake ..
$ cmake . --build
```

```shell
$ .\src\Debug\drafter.exe --help
usage: drafter [options] ... <input file>

API Blueprint Parser
If called without <input file>, 'drafter' will listen on stdin.

options:
  -o, --output          save output Parse Result into file (string [=])
  -f, --format          output format of the Parse Result (yaml|json) (string [=yaml])
  -s, --sourcemap       export sourcemap in the Parse Result
  -h, --help            display this help message
  -v, --version         print Drafter version
  -l, --validate        validate input only, do not output Parse Result
  -u, --use-line-num    use line and row number instead of character index when printing annotation
  -L, --enable-log      enable logging
```

```shell
.\src\Debug\drafter.exe --validate ..\features\fixtures\blueprint.apib

OK.
warning: (9)  the 'headers' section at this level is deprecated and will be removed in a future, use respective payload header section(s) instead :685:45
warning: (9)  the 'headers' section at this level is deprecated and will be removed in a future, use respective payload header section(s) instead :992:45
warning: (4)  invalid value format for 'number' type. please check mson specification for valid format :200:102
```

That levaes me with one question, why is the debug build happening by default? We have release by default on other platforms (https://github.com/apiaryio/drafter/blob/50cfec3f5659c97f72a29d6e42b932daf513c80e/DefaultBuildType.cmake#L1).

src/utils/Utf8.h
```diff
@@ -213,7 +213,8 @@ namespace drafter
 
                 template <typename Container>
                 explicit constexpr input_iterator(const Container& last) noexcept //
-                    : input_iterator(last.begin(), last.end())
+                    : e_(last.end()),
+		      cp_and_next_(decode_one(last.begin(), last.end()))
```
kylef: The indetation here appears to be wrong (but clang-format isn't complaining). Perhaps it's just GitHub rendering incorrectly.

PR #733, opened 2019-07-18
Refactor - replace some part of RefractDataStructure by functional way

klokane:
Little bi more functional way refactoring

- remove overused template parameter from `DescriptionToRefract()`
- remove `struct Join` and replace it by `accumulate` && lambda combo
- repace `ValueHasMembers()` for-loop by functional `find_if`

### DIFF
--- a/src/RefractDataStructure.cc
+++ b/src/RefractDataStructure.cc
@@ -33,6 +33,7 @@
 
 #include <fstream>
 #include <functional>
+#include <numeric> // accumulate
 
 using namespace refract;
 using namespace drafter;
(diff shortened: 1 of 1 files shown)


src/RefractDataStructure.cc
```diff
@@ -1094,12 +1081,9 @@ namespace
 
     bool ValueHasMembers(const mson::ValueMember* value)
     {
-        for (mson::TypeSections::const_iterator it = value->sections.begin(); it != value->sections.end(); ++it) {
-            if (it->klass == mson::TypeSection::MemberTypeClass) {
-                return true;
-            }
-        }
-        return false;
+        return std::find_if(value->sections.cbegin(), value->sections.cend(), [](typename mson::TypeSections::const_reference item) {
```
tjanc: `value` is const*, why not `value->sections.begin()` (resp. `value->sections.end()`)?

src/RefractDataStructure.cc
```diff
@@ -916,27 +899,31 @@ namespace
...
+    std::unique_ptr<IElement> DescriptionToRefract(const DescriptionInfoContainer& descriptions)
     {
         if (descriptions.empty()) {
             return nullptr;
         }
 
-        std::string description;
-        Join join(description);
-        snowcrash::SourceMap<std::string> sourceMap;
+        auto info = std::accumulate(descriptions.cbegin(), descriptions.cend(), DescriptionInfo(),
```
tjanc: nice

PR #735, opened 2019-07-19
Switch Windows CI from GYP to CMake

tjanc:
Remove Windows GYP CI and introduce CMake instead.
- also removed some redundant windows builds in the process

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,24 +1,19 @@
+version: '{build}'
+clone_folder: c:\projects\drafter
 image:
-  - Visual Studio 2015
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -1,24 +1,19 @@
...
-  - Visual Studio 2017
-
-environment:
-  matrix:
-    - VS_VERSION: 
-    - VS_VERSION: debug
-
+- Visual Studio 2015
+configuration:
+- Release
```
pksunkara: What about Debug and 2017?

PR #739, opened 2019-07-31
Update docker image to build with CMake

kylef:
I realised the docker image isn't tested nor updated from GYP to CMake, this pull request solves both. We are now building and running the image in CI and building using CMake.

I've refactored the drafter image a bit to use Alpine to shrink the size down, making use of separate environments for build and running so we can remove build dependency slices from the image. This makes the image around 1.8GB smaller.

```
$ docker images | grep drafter
drafter           96MB
apiaryio/drafter  1.86GB
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -78,6 +78,7 @@ workflows:
   drafter:
     jobs:
       - lint
+      - build: *tag-filter
       - test-gcc-49-debug: *tag-filter
       - test-gcc-49-release: *tag-filter
       - test-gcc-latest-debug: *tag-filter
@@ -95,6 +96,7 @@ workflows:
       - test-asan: *tag-filter
       - release:
           requires:
+            - build
             - test-gcc-49-debug
             - test-gcc-49-release
             - test-gcc-latest-debug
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -128,6 +130,18 @@ jobs:
           command: |
             ./tools/third-party-licenses-check.sh
 
+  build:
+    docker:
+      - image: circleci/python:3.7
```
kylef: I've used circleci's Python image here, the only thing we actually need is docker and git which is available on this image.

.circleci/config.yml
```diff
@@ -128,6 +130,17 @@ jobs:
...
+
+    steps:
+      - checkout
+      - run: git submodule update --init --recursive
+      - setup_remote_docker
+      - run: docker build -t drafter .
+      - run: echo '# My API' | docker run -i drafter
```
kylef: This is a primitive check to make sure that the image is capable of parsing the document. For example, all the dynamic libraries that Drafter depends on are available. If it fails it will exit 1 and fail the CI.

PR #746, opened 2019-09-11
feat: Added asset generating options in drafter_parse

pksunkara:
I have not added the c-api tests because they are not acutally running anyway I try to. Waiting on @tjanc to look into it.

Also, after a bit of thinking I have decided not to add these options to the CLI. What is your opinion @kylef? I don't mind either way.

---
This change is [Reviewable](https://reviewable.io/reviews/apiaryio/drafter/746)


test/test-CAPI.c
```diff
@@ -118,10 +122,81 @@ int test_parse_to_string_requiring_name()
...
+    assert_true(result);
+
+    assert_true(strstr(result, expected_without_name) != 0);
+
+    free(result);
+
+    return 0;
+};
+
+const char* source_mson = "# GET /\n+ Response 200 (application/json)\n    + Attributes\n        + name: pavan";
```
pksunkara: This is `C` code, which means I couldn't use the fixture loading functions we already had.

pksunkara: I could just test that `messageBody` string is contained or not. @kylef Would you prefer it that way? It just doesn't test completely.

src/config.cc
```diff
@@ -39,6 +41,8 @@ void PrepareCommanLineParser(cmdline::parser& parser)
         "yaml",
         cmdline::oneof<std::string>("yaml", "json"));
     parser.add(config::Sourcemap, 's', "export sourcemap in the Parse Result");
+    parser.add(config::MessageBody, 'b', "generate message bodies in the Parse Result");
+    parser.add(config::MessageBodySchema, 'c', "generate message body schemas in the Parse Result");
```
kylef: What does the `c` flag stand for?

src/drafter.h
```diff
@@ -54,6 +54,8 @@ typedef enum
  */
 typedef struct {
     bool requireBlueprintName;
+    bool generateMessageBody;
+    bool generateMessageBodySchema;
```
kylef: Every time we add an item to the options we are breaking the ABI. I wonder if it makes sense to redesign this in the future so we can introduce changes without breaking ABI stability?

Perhaps the internals of the structure can be hidden from the public interface are we can provide functions to set the values?

```c
drafter_parse_options *drafter_parse_options_alloc(void) {
    // FIXME - set all option defaults
    return (drafter_parse_options *)malloc(sizeof(drafter_parse_options));
}

void drafter_parse_options_dealloc(drafter_parse_options * options) {
    free(options);
}

drafter_parse_options *drafter_parse_options_set_generate_message_body(drafter_parse_options * options, bool generate_message_body) {
    options->generateMessageBody = generate_message_body;
}
```

@tjanc what do you think?

test/test-CAPI.c
kylef: Might be better approach, otherwise we risk other changes affecting the parse result making us have to alter the fixtures in `expected_without_assets`

src/config.cc
pksunkara: For schema, but `s` is already taken.

pksunkara: Suggestions welcome

src/drafter.h
kylef: Larger example: https://github.com/kylef-archive/geoffrey/blob/c197c9ed6ed550aec74e53ef2842149eb3520659/src/main.c#L7 // https://github.com/kylef-archive/geoffrey/blob/c197c9ed6ed550aec74e53ef2842149eb3520659/examples/echo.c#L41

pksunkara:
Updated the C API tests

src/drafter.h
kylef: Discussed in sprint planning, here is the output:

- We'll create an interface such as https://github.com/apiaryio/drafter/pull/746/files#r324681068 which allows us to add and change options without breaking ABI stability.

test/test-CAPI.c
tjanc: I think the way to test this feature is
1) integration tests running the `drafter` executable
2) unit tests acting on the C++ API

The C API itself isn't properly tested anyway. If we want to fix that, it's a task in itself; probably one where we need to implement a mocked version of our C++ interface (C API tests should ensure proper dispatch to the C++ interface, really).

src/config.cc
tjanc: what about capital `S`

src/drafter.h
tjanc: @pksunkara Have I missed something? Is this implemented as discussed?

pksunkara:
@tjanc No, I haven't pushed the changes yet. I am working on drafter.js integration to make sure everything's good before finalizing.

tjanc:
Diverged too far from master, expect a PR to replace this in near future.

Issue #752, opened 2019-10-01
Drafter backend crashes on unexpanded `Include (object)`


tjanc:
```
# My API

## GET /

+ Response 200 (application/json)
    + Attributes
        + One Of
            + Include (object)
                + key: value
```

kylef:
Fixed in https://github.com/apiaryio/drafter/commit/4028172d97a1c072c9e0b9d5ee9a700e7462dd6e

PR #753, opened 2019-10-01
Fix issue #752

tjanc:
## Fixed
- primitive type names are now added to type registry
- release build: unresolved references are ignored in backend
- debug build: unresolved references are asserted on in backend

## Not fixed
- snowcrash not recognizing nested sections in
a primitive mixin.

### DIFF
--- a/src/refract/Registry.h
+++ b/src/refract/Registry.h
@@ -18,17 +18,21 @@ namespace refract
 {
     class Registry
     {
-        typedef std::map<std::string, std::unique_ptr<IElement> > Map;
-        Map registrated;
+    public:
+        using type_map = std::map<std::string, std::unique_ptr<IElement> >;
 
-        std::string getElementId(IElement& element);
+    private:
+        type_map types_;
+
+    public:
+        Registry();
 
     public:
         const IElement* find(const std::string& name) const;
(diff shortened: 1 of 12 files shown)


src/refract/Registry.h
```diff
@@ -18,17 +18,21 @@ namespace refract
 {
     class Registry
     {
-        typedef std::map<std::string, std::unique_ptr<IElement> > Map;
-        Map registrated;
+    public:
+        using type_map = std::map<std::string, std::unique_ptr<IElement> >;
```
kylef: Does this need to be public? If I understand correctly it's only used in Registry.cc as an implementation detail of Registry on how it is internally stored.

kylef:
Can you add a changelog entry too? Just so the next release is sure to mention the change and it doesn't slip by, something like:

```markdown
## TBD

### Bug Fixes

* Prevents a potential segmentaiton fault crash during XYZ circumstances (possibly use an example in API Blueprint below).
```

src/refract/Registry.h
tjanc: It does not need to be public, but the implementation file can reference this type outside of member definitions if it is.

PR #754, opened 2019-10-08
fix issue 749

### LINKED ISSUE #749: Duplicit value in enumeration
I've been busy with other projects for the last few months but returned to work on my API docs and discovered I now have 16 semantic errors, even though I've not changed the document. The errors are all:

Duplicit value in enumeration
Line xxxx  API Blueprint

They all refer to enumerations declared in Data Structures section

```apib
# Data Structures
## WaveRecord (object)

+ status: `Processed` (enum, required) - The wave status.
    + Members
        + `Draft` - The wave has been created but not queued.
        + `Processed` - The wave has been processed.
        + `Queued` - The wave has been queued but not processed.
```

The error line xxxx in this example refers to the `+ status:` line. If I remove the `Processed` value I get an error indicating I've not supplied a value.

At the moment it's just a semantic error, but I'd like to know what's changed and what I can do to fix the warnings.

klokane:
closes #749 

Ignore enum duplicity if defined inline and later again in members
```
+ state: processed (enum)
+ Members
    + queued
    + processed
    + done
```

### DIFF
--- a/test/fixtures/mson/enum-variants.json
+++ b/test/fixtures/mson/enum-variants.json
@@ -189,7 +189,7 @@
                     "meta": {
                       "description": {
                         "element": "string",
-                        "content": "C: 'L', E: ['L','S','M','XL'] - first 'L' it in value"
+                        "content": "C: 'L', E: ['L','S','M','XL'] - ignore \"inline\" duplicity \"L\""
                       }
                     },
                     "content": {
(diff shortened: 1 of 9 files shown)


pksunkara:
Looks like a merge conflict.

klokane:
rebased

test/fixtures/mson/enum-variants.json
```diff
@@ -1634,70 +1634,6 @@
...
-                      "content": 72
-                    }
-                  ]
-                }
-              ]
-            }
-          ]
-        }
-      },
-      "content": "duplicit value in enumeration"
```
tjanc: I think this warning is correct and shouldn't be removed. The only thing that needed changing is that the "mson example value" of an `enum` is added to `enumerations` _after_ all it's nested members are. What do you think?

tjanc: Although, perhaps it makes better sense to give a warning in the exact opposite example, where we introduce a new type to `enumerations` because the example does not match any nested type.

Issue #755, opened 2019-10-14
sigsegv


klokane:
following ADD goes to seg failt

```
# GET /

+ Response 200 (application/json)

        [
info:
  description: |
    ---
    # Description
```

Fix: add check of BytesRanges

https://github.com/apiaryio/drafter/blob/master/ext/snowcrash/ext/markdown-parser/src/ByteBuffer.cc#L140

```
    if (rangeSet.empty()) {
        return characterMap;
    }
```

PR #756, opened 2019-10-15
Closes issue #755

klokane:


kylef:
The contents of the JSON test fixture appears to be empty, is the content missing?

klokane:
Sry, my mistake. Forgot to add test among running tests

Issue #758, opened 2019-10-30
Include Drafter version in ABI


kylef:
Currently there isn't a public API method to get the version of Drafter, we should have one for runtime checks. I'd like to make use of the version in Draughtsman the Python binding (https://github.com/kylef/draughtsman/).

kylef:
@klokane suggested elsehwere (https://github.com/apiaryio/drafter/pull/762#pullrequestreview-324330045) we can bring the version number into CMake too.

kylef:
Turns out I missed this one: https://github.com/apiaryio/drafter/blob/master/src/drafter.h#L113-L116

PR #762, opened 2019-11-28
Drafter 4.1.0

kylef:


kylef:
@klokane that's a good idea, and it would also be good to expose the version in the ABI. It's not possible to get hold of version if you link against Drafter: https://github.com/apiaryio/drafter/issues/758

PR #766, opened 2020-01-08
(C API breaking) Move drafter options declarations to implementation

tjanc:
Avoid exposing options structures to allow adding options in an additive, non-breaking process.

To add an option:
* drafter.h: document the option and its default value in the comment above `drafter_init_parse_options` (resp. `drafter_init_serialize_options`) and add a declaration for a setter for the option
* drafter.cc: add the option to the `drafter_parse_options` (resp. `drafter_serialize_options`) structure itself, implement a setter for the option, add a default value to the initializer list in `drafter_init_parse_options` (resp. `drafter_init_serialize_options`)


kylef:
I'd perhaps leave the "DO NOT MERGE" label because once merged this will block other releases. We should perhaps have Protagonist/drafter.js changes ready to go.

src/drafter.cc
```diff
@@ -83,7 +127,7 @@ DRAFTER_API drafter_error drafter_parse_blueprint(
 
     sc::BlueprintParserOptions scOptions = sc::ExportSourcemapOption;
 
-    if (parse_opts.requireBlueprintName) {
+    if (parse_opts->requireBlueprintName) {
```
kylef: If `parse_opts` is NULL we'd crash here. Should we add a check above and provide an error? (or support `parse_opts` being NULL an `requireBlueprintName` goes to default value (false)

kylef: Allowing null options may be nicer as a user of the library, especially when getting started.

klokane:
What about linked list instead of adding new option to public API?

```
struct drafter_option {
   const char* name;
   option_type type; /* enum { number, string, boolean } */
   void* value;
   drafter_option* next;
};

```

Then add option in following way 
```
drafter_option* create_options();
result add_option(drafter_option* option, const char* name, option_type type, void* value);
/* return OK or ERR_UNKNOWN_OPTION_NAME .... */
```

src/drafter.cc
tjanc: Yes we can probably just use the default values on nullptr.

tjanc:
@klokane I would rather have a stronger contract (vs a list of string-void*). If we add an option to drafter, it seems ok to me that we also _extend_ the API with an appropriate setter; the goal of this PR is to avoid _breaking_ the API in such a case.
I also thought about uniting the two options structures into one, but the differentiating between parsing and serialisation felt right in the end.

tjanc:
This should be ready to merge now, what do you think @klokane

kylef:
@tjanc we'll need to update the documentation in the README at https://github.com/apiaryio/drafter#parsing-a-blueprint-to-a-json-or-yaml-string to use these newer APIs too.

kylef:
Upon release we should update apiblueprint.org's code example too: https://github.com/apiaryio/apiblueprint.org/blob/d461165b151992ec55683d7d92cef86c3ae10c60/source/developers.html.md#using-the-native-parser-interface-cc (https://apiblueprint.org/developers.html)

PR #767, opened 2020-02-12
Improve readability of ConversionContext; remove WrapperOptions

tjanc:
Some basic refactoring of ConversionContext:
- removed wrapper options, a needless abstraction
- moved some implementations to source files
- shortened two FarTooMuchLong method names

Stumbled upon this while adding an option to omit generating assets.

### DIFF
--- a/src/ConversionContext.h
+++ b/src/ConversionContext.h
@@ -8,42 +8,42 @@
 #ifndef DRAFTER_CONVERSIONCONTEXT_H
 #define DRAFTER_CONVERSIONCONTEXT_H
 
+#include <boost/container/vector.hpp>
+
 #include "refract/Registry.h"
-#include "snowcrash.h"
 #include "SourceMapUtils.h"
+
+namespace snowcrash
+{
+    struct SourceAnnotation;
+}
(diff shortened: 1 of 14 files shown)


src/ConversionContext.h
```diff
@@ -8,42 +8,42 @@
...
+namespace drafter
+{
     class ConversionContext
     {
-        refract::Registry registry;
-        const NewLinesIndex newLinesIndex;
-
     public:
-        ConversionContext(const char* source, const WrapperOptions& options);
+        using Warnings = boost::container::vector<snowcrash::SourceAnnotation>;
```
klokane: Is there some specific reason to prefer  boost vector?

tjanc: The standard doesn't say whether in `std::vector<T,...>` the type `T` must be defined or not. I think specifically Microsoft's implentation requires a definition. `boost::container::vector<T,...>` doesn't depend on the definition of T.
In my refactoring, I moved the snowcrash include out of the header, and forward declared `snowcrash::SourceAnnotation`.

PR #768, opened 2020-02-13
Implementation of URITemplate parsing based on pegtl

klokane:
- grammar is implemented as recoverable parsing. While identifi error,
it will roll back to start od token, then consume invalid part up to
begin of new one part of grammar

- it is will invoke different semantic action for valid and invalid
part, so we are able to report it

- it report invalid character in invalid token

- in curent version if cannot report location of error (altrought it
knows offset in URITemplate). Reason is, we receive sourcemap for
MarkdownNode instead of start of URI template.


test/fixtures/parse-result/blueprint.json
```diff
@@ -163,7 +163,7 @@
           ]
         }
       },
-      "content": "The URI template contains square brackets, please percent encode square brackets as %5B and %5D"
+      "content": "URI template contains invalid character '['. You probably will need to repace invalid chars by percent encoded character."
```
tjanc: typo: `repace` -> `replace`

test/fixtures/parse-result/blueprint.sourcemap.json
```diff
@@ -363,7 +363,7 @@
           ]
         }
       },
-      "content": "The URI template contains square brackets, please percent encode square brackets as %5B and %5D"
+      "content": "URI template contains invalid character '['. You probably will need to repace invalid chars by percent encoded character."
```
tjanc: typo: `repace` -> `replace`

test/fixtures/syntax/undisclosed-listitem.json
```diff
@@ -63,6 +63,70 @@
...
+                      "content": 10
+                    }
+                  ]
+                }
+              ]
+            }
+          ]
+        }
+      },
+      "content": "URI template contains invalid character ' '. You probably will need to repace invalid chars by percent encoded character."
```
tjanc: typo: `repace` -> `replace`

src/parser/recover.h
```diff
@@ -0,0 +1,138 @@
...
+//  apib::parser
+//
+//  Created by Jiri Kratochvil on 27-01-2020
+//  Copyright (c) 2020 Apiary Inc. All rights reserved.
+//
+
+#ifndef PARSER_RECOVER_H
+#define PARSER_RECOVER_H
+
+#include "../PEGTL/include/tao/pegtl.hpp"
```
tjanc: Fix include to something like `#include <tao/pegtl.hpp>` instead of referencing its exact relative path.

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
-
-    for (unsigned int i = 0; i < uriTemplate.length(); i++) {
-        if (uriTemplate[i] == '{')
-            openCount++;
-        if (uriTemplate[i] == '}')
-            closeCount++;
-    }
-    return openCount != closeCount;
-}
+#include "../../../src/parser/Uritemplate.h"
```
tjanc: If we merge this, we should move snowcrash into the `src/parser` directory ASAP so we can avoid this kind of ugliness.

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
-    return expressions;
-}
+    };
 
-static ClassifiedExpression ClassifyExpression(const Expression& expression)
-{
+}
 
-    VariableExpression variableExpression(expression);
+struct VariableValidityChecker {
```
tjanc: Is there a specific reason we are using a callable function object instead of lambda-wrapped regular functions? Example:
```
void checkVariableValidity(...);
algorithm(begin(...), end(...), [](const auto& a){ checkVariableValidity(a, ...); })
```
Mostly it's possible to avoid `struct Foo{ ... operator()(...); }`, so that plain old functions are available; a lambda can be instantiated out of those building blocks in-line.

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
+        ss << "URI template contains invalid " << state.message
+           << ". You probably will need to repace invalid chars by percent encoded character.";
+
+#if 0
+        /**
+         *  this is prepared to set exact location of invalid character
+         *  but in current version we receive souce map for full makdown node instead of position
+         *  URI template so we are not able to set exact location of invalid character
+         */
+        assert(sourceBlock.size() == 1);
```
tjanc: Are you sure this should be a debug assertion? Note that if `sourceBlock` is empty, line 168 comes with UB.

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
+                return "label";
+            case state::expression_type::path:
+                return "path segment";
+            case state::expression_type::path_param:
+                return "path segment parameter";
+            default:
+                return "reserved operator";
+        }
+        assert(0);
+        return std::string{};
```
tjanc: Suggestion:
```
            default:
                break;
        }
        return "reserved operator";

```

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
+    const size_t gSize = 5;
 
     if (uri.empty())
         return;
 
-    if (RegexCapture(uri, URI_REGEX, groups, gSize)) {
-        result.scheme = groups[1];
-        result.host = groups[3];
-        result.path = groups[4];
+    if (!RegexCapture(uri, URI_REGEX, groups, gSize)) {
```
tjanc: I fail to understand what we are doing here. Is it to ensure the URI template is also a valid URI?
The only special thing I feel we have to do when parsing URI templates in APIB/MSON is that they need to start from root (`/...`); just to be able to distinguish sections recognized by starting with a URI template. What am I missing?

ext/snowcrash/src/ParametersParser.h
```diff
@@ -16,6 +16,8 @@
 #include "StringUtility.h"
 #include "BlueprintUtility.h"
 
+#include "../../../src/parser/Uritemplate.h"
```
tjanc: If we merge this, we should move snowcrash into the `src/parser` directory ASAP so we can avoid this kind of ugliness.

src/parser/Uritemplate.h
```diff
@@ -0,0 +1,396 @@
...
+                                 pegtl::sor<
+                                  literals,
+                                  expression
+                                 >
+                             > {};
+
+            struct match_grammar : pegtl::seq<grammar, pegtl::eof>{};
+
+            
+            namespace state {
```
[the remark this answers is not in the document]
klokane: In general, we can, but in this case is probably also good idea move actions into its on headers.

But I mean it is small part and it is related to parsing so I personally prefer to hold it together. But if we agree on, I have no problem to split it into separated files

src/parser/Uritemplate.h
```diff
@@ -0,0 +1,396 @@
...
+            // try_catch because we want to continue by recovery - so we need local failure instead of global
+            struct pct_encoded : pegtl::try_catch< 
+                                   pegtl::if_must<
+                                     pegtl::one< '%' >,
+                                     pegtl::abnf::HEXDIG,
+                                     pegtl::abnf::HEXDIG
+                                   >
+                                 > {};
+
+            struct ucschar : pegtl::utf8::ranges<
```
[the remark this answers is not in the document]
klokane: Sure, I can add link to RFC

src/parser/Uritemplate.h
```diff
@@ -0,0 +1,396 @@
...
+        {
+
+            namespace pegtl = tao::pegtl;
+            
+            // try_catch because we want to continue by recovery - so we need local failure instead of global
+            struct pct_encoded : pegtl::try_catch< 
+                                   pegtl::if_must<
+                                     pegtl::one< '%' >,
+                                     pegtl::abnf::HEXDIG,
+                                     pegtl::abnf::HEXDIG
```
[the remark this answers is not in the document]
klokane: > pct-encoded    =  "%" HEXDIG HEXDIG

src/parser/Uritemplate.h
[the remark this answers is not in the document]
tjanc: Let's go with your recommendation.

src/parser/Uritemplate.h
```diff
@@ -0,0 +1,396 @@
...
+                                     pegtl::range<0x23, 0x24>,
+                                     pegtl::one<0x26>,
+                                     pegtl::range<0x28, 0x3b>,
+                                     pegtl::one<0x3d>,
+                                     pegtl::range<0x3f, 0x5b>,
+                                     pegtl::one<0x5d>,
+                                     pegtl::one<0x5f>,
+                                     pegtl::range<0x61, 0x7a>,
+                                     pegtl::one<0x7e>
+                                   > {};
```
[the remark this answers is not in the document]
klokane: See comment above. It is for discussion how to solve it.

ext/snowcrash/src/UriTemplateParser.cc
klokane: Yes it will be better solution, but in current state how is project splited `ext/snowcrash` vs `src/parser` it is not possible.

There is option add into cmake in snowcrash include paths `../../src` but it also is not good idea. 

So we have two options:
- move `src/paser` into `ext/snowcrash/`
- move snowcrash from `ext` to integral part of `drafter`

ext/snowcrash/src/UriTemplateParser.cc
klokane: It is called from `variant::visit` (resp. `mpark::visit`)
So it is common way how it is used. 

Or maybe I fully do not understand how do you mean your suggestion

ext/snowcrash/src/UriTemplateParser.cc
klokane: It is proposal how to correct implement location of invalid character.
It is also reason why it is closed in `#if 0/#endif` block - see comment above.

ext/snowcrash/src/UriTemplateParser.cc
klokane: It is taken from original implementation. 
```
    struct ParsedURITemplate {
        std::string scheme;
        std::string host;
        std::string path;

        Report report;
    };
```

It is reason why it is splited here. I didn't want to change code deeply into snowcrash in this phase of implementation.

ext/snowcrash/src/UriTemplateParser.cc
tjanc: An example of what I have in mind, in the context of `variant`:
* delaration of specific C functions: https://github.com/apiaryio/drafter/blob/master/src/refract/JsonValue.cc#L83
  * notice the specific declarations for each variant version
  * notice the last declaration in that block (L96) that declaras a generalized function
* definition of the generalized function then uses the specific C functions and an inline lambda: https://github.com/apiaryio/drafter/blob/master/src/refract/JsonValue.cc#L302

There is no reason to define a named type `VariableValidityChecker` when we can have `checkVaraibleValidity` and wrap that in an anonymous type via a lambda where needed. The following benefits come to mind:

* readability
  * one less symbol
  * state of execution is more explicit if carried in function parameters, rather than in the body of a callable structure
* usability - a C function is more useful than a C++ object with special operators
* abstraction - a C function can be made into a callable object without cost, whereas reducing a callable object to a C function depends on compiler optimizations to avoid abstraction cost

tjanc: Oh, and to avoid confusion: The referenced example uses a named callable structure, but that structure's `operator()` just delegates the call completely to overloaded C functions. The reason for that is C++11 compatibility, back then there was no `auto` in lambdas (`[...](const auto&){...}`). So that's just a compatibility workaround.

src/parser/Uritemplate.h
```diff
@@ -0,0 +1,396 @@
...
+             *  Original set is named `rfc_5670_ascii_allowed`
+             *  while modified is named `ascii_allowed`
+             *
+             *  From `ascii_allowed` are removed chars to be RFC1738 compatible
+             *  "~", "[" , "]"
+             *  %x7E / %x5B / %x5D 
+             *
+             */
+
+            struct rfc_5670_ascii_allowed : pegtl::sor<
```
kylef: The relevant part of RFC6570 is 2.1 (literals)  refers to utf-8 and unicode. Should we avoid calling it ascii as the spec specifies the character encoding as utf-8 not ascii.

> 2.1.  Literals
> 
>    The characters outside of expressions in a URI Template string are
>    intended to be copied literally to the URI reference if the character
>    is allowed in a URI (reserved / unreserved / pct-encoded) or, if not
>    allowed, copied to the URI reference as the sequence of pct-encoded
>    triplets corresponding to that character's encoding in UTF-8
>    [RFC3629].
> 
>      literals      =  %x21 / %x23-24 / %x26 / %x28-3B / %x3D / %x3F-5B
>                    /  %x5D / %x5F / %x61-7A / %x7E / ucschar / iprivate
>                    /  pct-encoded
>                         ; any Unicode character except: CTL, SP,
>                         ;  DQUOTE, "'", "%" (aside from pct-encoded),
>                         ;  "<", ">", "\", "^", "`", "{", "|", "}"

src/parser/recover.h
```diff
@@ -0,0 +1,138 @@
...
+#include <map>
+
+namespace apib
+{
+    namespace parser
+    {
+
+        //
+        // recovery parsing of list
+        // if list item is wrong defined it allows continue to prsing with next item
```
kylef:
```suggestion
        // if list item is wrong defined it allows continue to parsing with next item 
```

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
-            return;
-        }
+    state::uritemplate state;
 
-        if (HasNestedCurlyBrackets(result.path)) {
-            result.report.warnings.push_back(
-                Warning("The URI template contains nested expression brackets", URIWarning, sourceBlock));
-            return;
-        }
+    //tao::pegtl::memory_input<> in(result.path, "");
```
kylef:
```suggestion
```

ext/snowcrash/src/UriTemplateParser.cc
```diff
@@ -9,207 +9,251 @@
...
+    void operator()(const state::invalid& s) const {
+
+        tao::pegtl::memory_input<> in(s.content, "");
+        error_locator::state state;
+        tao::pegtl::parse<error_locator::grammar<varspec>, error_locator::report_action>(in, state);
+
+        std::stringstream ss;
+        ss << "URI template variable '" << s.content
+           << "' contains invalid " << state.message
+           <<". Allowed characters for expressions are A-Z a-z 0-9 _ and percent encoded characters";
```
kylef: Should we perhaps suggest how to percent encode the value? We could URI encode the `state.messge` and suggest that to the user.

"Replacing 'x' with %ab` etc

ext/snowcrash/test/test-UriTemplateParser.cc
```diff
@@ -54,7 +54,9 @@ TEST_CASE("Parse uri template for invalid format curly brackets (nested brackets
 
     parser.parse(uri, sourceBlock, result);
 
-    REQUIRE(result.report.warnings.size() == 1);
+    REQUIRE(result.report.warnings.size() == 2);
+    CHECK(result.report.warnings[0].message == "URI template variable 'id{' contains invalid nested expression bracket '{'. Allowed characters for expressions are A-Z a-z 0-9 _ and percent encoded characters");
+    CHECK(result.report.warnings[1].message == "URI template contains invalid character '}'. You probably will need to repace invalid chars by percent encoded character.");
```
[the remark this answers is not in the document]
klokane: It is not possible in current implementation. I know position of character in uri template, but from caller I do not receive begin of uri template. Instead I have full markdown node.

We need to repair it one level up.

src/parser/Uritemplate.h
klokane: This is just subset from ASCII range

Full definition of literals (including utf-8) is later https://github.com/apiaryio/drafter/blob/6c4a9c5f20f09c18b275338188f0ab685fb3cdf3/src/parser/Uritemplate.h#L197-L204

ext/snowcrash/src/UriTemplateParser.cc
klokane: Yes, we can prepare simple parser which will suggest replacement. I send it as one of proposals in internal discussion on slack.

It is question if we want to have it in this PR.

klokane: We know exact character, so it is very simple task

ext/snowcrash/src/UriTemplateParser.cc
klokane: Sry I miss it, will remove

ext/snowcrash/src/UriTemplateParser.cc
kylef: Yeah, I'd suggest we don't solve it here. We can make a follow up change to do this once this is merged so we don't snowball here.

ext/snowcrash/src/UriTemplateParser.cc
kylef: > It is proposal how to correct implement location of invalid character

Given the current behaviour is the same as the one in this PR. I'd descope this change from this changeset so we can move it forward. Improving the source maps can be done as part of a separate task which is already tracked in JIRA.

ext/snowcrash/src/ParametersParser.h
```diff
@@ -200,28 +197,49 @@ namespace snowcrash
...
+        std::for_each(result.begin(), result.end(), 
+            [&collect](const state::part& part) {
+                if (const auto expression = mpark::get_if<state::expression>(&part)) {
+                  std::transform(expression->variables.begin(), expression->variables.end(),
+                      collect,
+                      [](const state::expression::variable_type& var){
+                        if (const auto valid = mpark::get_if<state::variable>(&var)) {
+                          return valid->name;
+                        } else if (const auto invalid = mpark::get_if<state::invalid>(&var)) {
+                          // FIXME: what to do with invalid variable names? Shouldn't we ignore them?
```
kylef: I think that makes sense, we can filter them out.

klokane:
In this commit no.
pegtl is included from drafter, not in snowcrash.

UriTemplate is part of drafter and we inlude this headers in snowcrash.

I propose to resolve it in individual commit after discussion about solution

ext/snowcrash/src/ParametersParser.h
klokane: Okay, will remove check against invalid variables

klokane:
So, we can merge?
Or is there something to resolve?

kylef:
@klokane I think this can be merged and I don't think there is any further feedback.

PR #769, opened 2020-02-25
(Payload section -> APIE) Generalized mediaType handling

tjanc:
During the translation of an API Blueprint _Payload Section_ to corresponding _Asset Elements_,

* content types are fully interpreted (parsed, evaluated, serialized)
  * uses PEGTL grammar
  * implements media type serialization (unit tests needed)
* translation of user-supplied _Body Section_ & _Schema Section_ into _Asset Elements_ is performed in isolation from generating and attaching derived _Asset Elements_.
  * prerequisite to implement drafter options `generate-json-values` & `generate-json-schemas`

### Impact on tests
* Behaviour now corresponds to `generate-json-values=true`, `generate-json-schemas=true`
  * `test/fixtures/api/action-request-attributes.json`
    * adds generated JSON Schema
  * `test/fixtures/api/issue-702.json`
    * adds generated JSON
    * adds generated JSON Schema
  * `test/fixtures/api/schema-body.json`
    * fixes `class` of a JSON schema asset to equal `messageBodySchema` where previously it was `messageBody`
    * formats media type of that schema (removes white space not required)
  * `test/fixtures/api/schema-custom-json.json`
    * adds generated JSON Schema
  * `test/fixtures/api/schema-custom.json`
    * fixes `contentType` of a custom schema asset to equal `application/schema+xml` where previously it was `application/xml`
  * `test/fixtures/extend/many.json`
    * fixes propagation of media type parameters (`charset=utf-8`) where they were previously omitted
  * ` test/fixtures/render/action-request-attributes.json`
    * adds generated JSON Schema
  * `test/fixtures/render/content-type.json`
    * fixes propagation of media type parameters (`charset=utf-8`) where they were previously omitted


src/RefractAPI.cc
```diff
@@ -269,24 +246,89 @@ std::unique_ptr<IElement> HeaderToRefract(const NodeInfo<snowcrash::Header>& hea
...
+
+namespace
+{
+    using MediaType = apib::parser::mediatype::state;
+
+    apib::parser::mediatype::state addSchemaSubtype(apib::parser::mediatype::state m)
+    {
+        if (m.suffix.empty())
+            m.suffix = m.subtype;
+        m.subtype = "schema";
```
kylef: I’m not sure this is the right approach, while it works for `application/json` as JSON Schema is `application/schema+json` it might not work for all formats. For example XML DTD (Document Type Definition) has the mime type `application/xml-dtd` (https://www.w3.org/2006/02/son-of-3023/draft-murata-kohn-lilley-xml-04.html#anchor3) not `application/schema+xml`.

There might also be other attributes in the `MediaType` that might not make sense on our schemas if we are generating them. In the case of other unknown parameters. Even `charset` might be set as what the server is returning in its response, the charset for our generated JSON Schema or JSON Schema in the asset is going to be unicode regardless (atleast in current serialisation) as that’s already enforced by API Elements parse result being serialised as JSON (as that specifies unicode).

tjanc: I'm also not sure about the approach with adding `schema`. I will restrict the behavior to json-type'd media types.

However, I feel that the media type of an unknown XML schema is as much `application/schema+xml` as it is `application/xml` as it is `application/beer`. It's neither of these, in general. And in the code we can recognize that ambiguity. Wouldn't it be even better to produce `text/plain` or nothing at all in such cases?

Regarding media type parameters; `master` currently attempts propagating all parameters, I felt this PR has a diff large enough already, and this PR therefore just fully propagates any parameters down to their respective Body/Schema sections. Are you suggesting we should not propagate media type parameters on generated schemas? Or bodies? Because that is not current behavior.

kylef: > I'm also not sure about the approach with adding schema. I will restrict the behavior to json-type'd media types.

:+1:

> However, I feel that the media type of an unknown XML schema is as much application/schema+xml as it is application/xml as it is application/beer. It's neither of these, in general. And in the code we can recognize that ambiguity. Wouldn't it be even better to produce text/plain or nothing at all in such cases?

Perhaps, the assumption that a schema matches the content-type of the body isn't nessecerily correct but seems a close approximation that was done in the past. There could be cases where I am describing the body in a vastly different content type. Perhaps `text/plain` makes sense if we don't know the semantics behind it (or omitting the content-type in that case).

One example can be:

```
+ Request (application/yaml)
    + Body

            key: value

    + Schema

            {"type": "object"}
```

Where I've used JSON Schema to describe a YAML request. Although not sure in practise anyone would do such a thing.

> Are you suggesting we should not propagate media type parameters on generated schemas? Or bodies? Because that is not current behavior.

Looked like it was introduced based on the diff against some of the fixtures such as `test/fixtures/extend/many.json`.

test/fixtures/api/issue-702.json
```diff
@@ -155,6 +155,48 @@
...
+                                  ]
+                                }
+                              },
+                              "attributes": {
+                                "contentType": {
+                                  "element": "string",
+                                  "content": "application/json"
+                                }
+                              },
+                              "content": "{}"
```
kylef: Is this correct? According to how the parser has interpreted the body, it is `+ Body\n\n        {}\n\n+ Schema \n\n        {}\n` according to the asset above. We don't generate assets when the user have provided one.

test/fixtures/api/schema-body.json
```diff
@@ -130,15 +130,15 @@
                                   "content": [
                                     {
                                       "element": "string",
-                                      "content": "messageBody"
+                                      "content": "messageBodySchema"
```
kylef: I don't think this is correct, the response is itself a JSON Schema with content-type (application/schema+json ; hello=world). I-e, the API is expected to return a JSON Schema in the body.

```suggestion
                                      "content": "messageBody"
```

tjanc:
Closing in favor of #780 .

PR #771, opened 2020-03-04
Switch to JSON Schema Draft 7

klokane:
Implements transformation "one element enum" to "const" as defined in JSON schema draft 06

https://json-schema.org/draft-06/json-schema-release-notes.html


src/refract/JsonSchema.cc
```diff
@@ -138,9 +138,20 @@ namespace
         return schema;
     }
 
+    so::Object& addConst(so::Object& schema, so::Array value)
```
kylef: Anhy particular reason to use `so::Array` for value instead of `so::Value`? Since it can only be one value, I don't see why we are wrapping that in an array.

kylef:
If I understand correctly, there is no API Blueprint syntax which would trigger a `const` value in a JSON Schema. Should we update the JSON Schema optimisation logic at https://github.com/apiaryio/drafter/blob/klokane/json-schema-draft7/src/refract/JsonSchema.cc#L466 so that when `enum` is a single value it will use `const` instead?

src/refract/JsonSchema.cc
klokane: According how is `const` created - conversion form one member array i mean it is good way how to write it.

If you mean it is better just wrap one value I we can change 

line 151 to
```
addConst(schema, std::move(value.data.front());
```

And receive to addConst just one `so::Value`

klokane:
Maybe best solution check will be go thru commit 
https://github.com/apiaryio/drafter/pull/771/commits/2dc67272054d6debf64854489400e62ed7e64b45

If all changes are conforming to `const` definition.

I created separated commit, just with changes related to enum-const conversion

klokane:
I go thru and mean it is okay

src/refract/JsonSchema.cc
klokane: Maybe rename it `addConst()` to someting like `addEnumWithOneItem()`

kylef: I think we should make the change so that the `addConst` takes a single value. That's how `const` is described in JSON Schema spec, it takes one value. We should leave the details of extracting the single value to the use of `addConst` (so in `addEnum`). A secondary benefit is that it simplifies the safety of `addConst` because `addConst` knows it only has a single value and we don't have to assert length of an array.

Issue #630, opened 2016-08-12
Regex Insufficient Memory in Template Parser


abtris:
@bobdercole opened apiaryio/snowcrash#399

Hello,

I am using [API Blueprint Sublime Text Plugin](https://github.com/apiaryio/api-blueprint-sublime-plugin) and [Aglio](https://github.com/danielgtaylor/aglio). I'm having a common issue which I've narrowed down to Snowcrash's URI template parser. The parser seems to fail with very long URI templates.

Here is a working example:

```
[/test{?abcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcab}]
```

Here is a failing example (notice the additional character at the end):

```
[/test{?abcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabcabc}]
```

[...]

Issue #456, opened 2017-03-03
Reserved URI parameter "not found"


jackbentley:
This warning appears when using a parameter with `+` at the beginning (the *reserved value operator*). It shouldn't because [it's valid according to the spec](https://apiblueprint.org/documentation/specification.html#variable-with-reserved-characters-values)?

```
warning: (8)  parameter 'file' is not found within the URI template '/blah/{prefix}/files/{+file}'
```

pksunkara:
Yes, it's valid according to the spec. Unfortunately, our URI template implementation has a few bugs.

kylef:
As of [Drafter 5.0.0-rc.1](https://github.com/apiaryio/drafter/releases/tag/v5.0.0-rc.1), full support for URI Template has been added, including level 3/4.

URI Templates such as `/blah/{prefix}/files/{+file}` are now supported.

PR #776, opened 2020-03-16
Release 5.0.0-rc.1

kylef:
This PR creates a 5.0.0-rc.1 release, I think for final `5.0.0` release we need to do the following:

- Update the documentation in the README to count for the breaking changes to the Drafter API. The code examples/documentation in README is no longer current.

Drafter hard-coded the pre-release identifer as `pre`, I'd like to give better semantics for the release and use the recommended identifiers (alpha, beta, release candidates etc) and therefore I've updated the handling in the code base so that `DRAFTER_PRE_RELEASE_VERSION` takes the full pre-release identifier:

To aid review, here's how the Version.h is generated:

```c
#define DRAFTER_MAJOR_VERSION 5
#define DRAFTER_MINOR_VERSION 0
#define DRAFTER_PATCH_VERSION 0

#define DRAFTER_VERSION_IS_RELEASE 0
#define DRAFTER_PRE_RELEASE_VERSION "beta.1"

#if DRAFTER_VERSION_IS_RELEASE
#define DRAFTER_VERSION_STRING "v5.0.0"
#else
#define DRAFTER_VERSION_STRING "v5.0.0-beta.1"
#endif
```

---

### Breaking

* Parsing and serialisation options in the C API of Drafter contain breaking changes. Direct access to the option structures are no longer possible, a new API layer has been added for the options. See `drafter_init_parse_options` and `drafter_serialize_options` respectively.

### Enhancements

* Improved support for URI Template. Drafter supports up to, and including [level 4 of URI Template](https://tools.ietf.org/html/rfc6570#section-1.2). This includes support for the URI Template explode (`*`) modifier, and reserved value operator (`+`), and others.

  [#553](https://github.com/apiaryio/drafter/issues/553) [#456](https://github.com/apiaryio/drafter/issues/456) [#630](https://github.com/apiaryio/drafter/issues/630)

* Added support for JSON Schema Draft 7 in schema generation from MSON. JSON Schemas are now generated using Draft 7 instead of Draft 4.


kylef:
Ive' updated to 5.0.0-rc.1 as the version is correct.

PR #778, opened 2020-03-17
(doc) updated README to reflect recent C API changes

tjanc:


README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
-`drafter_serialize` function can be used to serialized the result as [API
-Elements](http://api-elements.readthedocs.io) in YAML or JSON.
+Tweaking `drafter_serialize_options` allows serialization into JSON.
 
+```c
+drafter_serialize_options* options = drafter_init_serialize_options();
+drafter_set_format(options, DRAFTER_SERIALIZE_JSON);
+
+char* jsonApie = NULL;
+if (0 == drafter_parse_blueprint_to(blueprint, &jsonApie, NULL, NULL)) {
```
kylef:
```suggestion
if (drafter_parse_blueprint_to(blueprint, &jsonApie, NULL, NULL) == DRAFTER_OK) {
```

What about using the drafter_error enum values, it perhaps a bit more clearer and readable for README purposes (especially if you are not familiar with C norms where 0 is truthy to success often). I don't think the target audience for Drafter is only C/C++ developers, there's people who want to try integrate it into their other languages with bindings/ffi and similiar so giving a bit of insight here can keep it clear for others.

README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
+    printf("%s\n", jsonApie);
+}
+
+free(jsonApie);
```
kylef: Should this go into the `if` block too? Would drafter always return an allocated block in `jsonApie` even if it failed?

README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
-    free(result);
+Without options, the resulting API Elements is serialized as YAML.
+
+```c
+char* yamlApie = NULL;
+if (drafter_parse_blueprint_to(blueprint, &yamlApie, NULL, NULL) == 0) {
+    printf("%s\n", yamlApie);
 }
+
+free(yamlApie);
```
kylef: Should this go into the `if` block too? Would drafter always return an allocated block in `yamlApie` even if it failed?

README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
 
-char* result = NULL;
-if (drafter_parse_blueprint_to(blueprint, &result, options) == 0) {
-    printf("%s\n", result);
-    free(result);
+Without options, the resulting API Elements is serialized as YAML.
+
+```c
+char* yamlApie = NULL;
+if (drafter_parse_blueprint_to(blueprint, &yamlApie, NULL, NULL) == 0) {
```
kylef: Ditto comment about `== DRAFTER_OK`

README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
-const char* blueprint =
-  "# My API\n"
-  "## GET /message\n"
-  "+ Response 200 (text/plain)\n"
-  "\n"
-  "      Hello World!\n";
+if (0 == drafter_check_blueprint(blueprint, result)) {
+    print("Understood.\n");
+} else {
+    print("Invalid API Blueprint.?\n");
```
kylef: Checking `drafter_error`, does this function return an error result if the API Blueprint is invalid, or does it pass success with the parse result?

Looking at the code I can see that `DRAFTER_EINVALID_OUTPUT` is if I don't set `result` to a pointer which is more of a programatic error than the API Blueprint being invalid.

`DRAFTER_EINVALID_INPUT` is if I pass `NULL` source, and I'd imagine that `DRAFTER_EUNKNOWN` is perhaps unexpected. So none of these cases appear to be invalid/errenous API Blueprint.

README.md
```diff
@@ -66,85 +66,118 @@ See [parse feature](features/parse.feature) for the details on using the `drafte
...
+}
+
+drafter_free_result(result);
```
kylef: Under what cases do we need to `free(result)`? It's not handled in any.

README.md
tjanc: This way - more safe.

README.md
tjanc: The user reading a README and developing his code upon only it, without looking into the header, is a user that should _always_ call `drafter_free_result`. It's not like it's not a `NOOP` when the user doesn't need the invocation. Also, notice the amount of negations in the last sentence.

kylef: 🤦‍♂ I didn't realise `result` was passed to `drafter_free_result`. Ignore this :)

README.md
tjanc: I usually think like this:
- Is the input of my interface a reference to a value?
  - If yes, can I use a reference (i.e. do I have C++)?
    - If yes, use a reference.
    - If not, **use a pointer and `assert` on it (first statement(s) of implementation)**.
  - If not, is the input of my interface a pointer to a value?
    - etc.

tjanc: I would remove the programming error values from `drafter_error` and just assert. The programmer wants an explosion, not a valid (error) value, on programatic errors, no?

What we are currently doing is basically lifting a programatic error into userland. And that's probably no business of libdrafter, which is nothing but a set of archives implementing text transformations.

README.md
kylef: Makes sense, I forgot passing NULL to `free` was safe.

PR #779, opened 2020-03-18
C API to support output parameters `== NULL`

tjanc:
Some pointers passed to the drafter C API have the character of output parameters. 

If set to `NULL`, their assignment could be skipped - perhaps the user is only testing for success.

This also contributes to simpler [examples](https://github.com/apiaryio/drafter/blob/tjanc/capi-allow-null-result/test/test-CAPI.c#L116).

### DIFF
--- a/src/drafter.cc
+++ b/src/drafter.cc
@@ -45,12 +45,7 @@ DRAFTER_API drafter_error drafter_parse_blueprint_to(const char* source,
         return DRAFTER_EINVALID_INPUT;
     }
 
-    if (!out) {
-        return DRAFTER_EINVALID_OUTPUT;
-    }
-
     drafter_result* result = nullptr;
-    *out = nullptr;
 
     drafter_error ret = drafter_parse_blueprint(source, &result, parse_opts);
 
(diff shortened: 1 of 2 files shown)


src/drafter.cc
```diff
@@ -45,20 +45,16 @@ DRAFTER_API drafter_error drafter_parse_blueprint_to(const char* source,
...
-    *out = nullptr;
 
     drafter_error ret = drafter_parse_blueprint(source, &result, parse_opts);
 
     if (!result) {
         return ret;
     }
 
-    *out = drafter_serialize(result, serialize_opts);
+    if(out)
```
opichals: All the rest of the lines use the block form of `if(out) { ... }`. What's the coding style rules on this?

tjanc: We don't have a rule on this; I tend to omit the brackets based on look & taste (TM).
If you think the missing brackets are misleading here, I should add them :) .

opichals: I'd vote to be consistent at least in a single file. So yes, please :)

tjanc: It's good you pointed this out, judging by the missing space after the `if` keyword and green CI earlier, our formatting CI test must be broken. Good to know, thanks for the insight from outside ;-)

PR #780, opened 2020-03-30
Parse media types during payload attachment

tjanc:
- Parse media types and handle them accordingly
- Prepare `RefractAPI.cc` for adding options dropping asset generation

## Affected tests
- `render/action-request-attributes`
  - generate schema where none was given
- `api/action-request-attributes`
  - generate schema where none was given
- `api/issue-702`
  - generate schema where none was given
- `api/schema-body`
  - generate schema where none was given
  - fix syntax of media type
- `api/schema-custom`
  - custom schema for unsupported media type - annotate `text/plain` instead of copying media type from value

### History
This branch is derived from #769 , and already (attempts to) address associated review comments.


src/RefractAPI.cc
```diff
@@ -98,39 +101,13 @@ namespace
         return !action.isNull() && !action.node->method.empty();
     }
 
-    NodeInfoByValue<snowcrash::Asset> renderPayloadBody(
-        const NodeInfo<snowcrash::Payload>& payload, RenderFormat format, const IElement& expanded)
+    NodeInfoByValue<snowcrash::Asset> renderPayloadBody(const NodeInfo<snowcrash::Payload>& payload)
```
kylef: Looks like this method is no longer used, shall we remove it?

```suggestion
```

src/RefractAPI.cc
```diff
@@ -98,39 +101,13 @@ namespace
...
-                    assert(false);
-            }
-            return NodeInfoByValue<snowcrash::Asset>{ ss.str(), NodeInfo<snowcrash::Asset>::NullSourceMap() };
-        }
         return NodeInfoByValue<snowcrash::Asset>{ payload.node->body, &payload.sourceMap->body };
     }
 
-    NodeInfoByValue<snowcrash::Asset> renderPayloadSchema(
-        const NodeInfo<snowcrash::Payload>& payload, RenderFormat format, const IElement& expanded)
+    NodeInfoByValue<snowcrash::Asset> renderPayloadSchema(const NodeInfo<snowcrash::Payload>& payload)
```
kylef:
```suggestion
```

Ditto, I don't see this used anywhere.

src/RefractAPI.cc
```diff
@@ -339,82 +379,53 @@ std::unique_ptr<IElement> PayloadToRefract( //
     if (!payload.node->description.empty())
         content.push_back(CopyToRefract(MAKE_NODE_INFO(payload, description)));
 
-    std::unique_ptr<IElement> payloadAttributeElement = payload.node->attributes.empty() ? //
-        nullptr :
+    auto unexpandedAttrs = payload.node->attributes.empty() ? //
```
kylef: I'd consider calling this something like `unexpandedDataStructures` because `attributes` is a bit of an overloaded term (API Blueprint syntax "attributes", MSON AST "attributes", API Elements "attributes") which confused me a for alittle bit.

src/backend/MediatypeS11n.cc
```diff
@@ -0,0 +1,89 @@
+//
+//  backend/MediatypeS11n.cc
```
kylef:
```suggestion
//  backend/MediaTypeS11n.cc
```

I find it weird that "type" is lowercased part of media type, I think this was a thing in the past implementation for Media type too though so perhaps it doesn't make sense to just change this here and leave it for consistency. Since "media type" is not one word.

src/backend/MediatypeS11n.cc
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    std::ostream& serialize_quoted(std::ostream& out, const std::string& v)
+    {
+        return serialize_quote_escaped(out << '"', v) << '"';
+    }
+
+    bool is_restricted(char c) noexcept
+    {
+        for (const auto& r : { '!', '#', '$', '&', '^', '_', '-', '.' })
```
kylef: Are you working from RFC 6838 or a different one?

I think we might be missing `+` perhaps `restricted-name-chars =/ "+"`? https://tools.ietf.org/html/rfc6838#section-4.2

tjanc: really good catch :clap:

tjanc:
Writing some unit tests, then we can merge.

tjanc:
Unit tests were indeed needed, had to add some fixes.

PR #782, opened 2020-04-14
Restructuring into `packages`

tjanc:
All dependencies including drafter are now contained flat in `./packages/`:

* `./packages/apib-parser/` - API Blueprint parser; contains `markdown-parser` and `snowcrash` sources; also contains new PEGTL grammars
* `./packages/apib/` - API Blueprint definition; only contains media type definition now
* `./packages/drafter/` - Drafter C library `libdrafter`, Drafter CLI `drafter-cli`
    * PEGTL grammars moved from `./src/parser/*` to `./packages/apib-parser/src/grammar/`
* `./packages/boost/` - external production dependency
* `./packages/variant/` - external production dependency
* `./packages/cmdline/` - external production dependency of `drafter-cli`
* `./packages/PEGTL/` - external production dependency of `apib-parser`
* `./packages/sundown/` - external production dependency of `apib-parser`
* `./packages/Catch/` - external test dependency
* `./packages/dtl/` - external test dependency

Intent on making the memory representation of our syntax tree available without dependence on parsing logic, I have introduced `./packages/apib/` to contain API Blueprint specific declarations and semantics. Currently it only contains media type definitions moved from `./src/parser`, but the intent is to declare new syntax definitions here instead of `./packages/apib-parser/grammar/`.


klokane:
I'm not sure if it is good idea to introduce `apib-parser` instead of `snowcrash`.
It can be also little bit confusing hold `sundown` separately from combo snowcrash/markdown-parser

tjanc:
@klokane 

All standalone projects, whether external dependencies or projects developed in this repository, should now be layed out flat in `./packages`. The build system should handle the structure of dependencies.

The API Blueprint parser `apib-parser` is a standalone project. Currently (before this PR) it is composed of:
* sundown, an external dependency
* markdown-parser, regular code somewhere in drafter
* snowcrash, regular code somewhere in drafter

We will be throwing away most of snowcrash and markdown-parser code during the PEGTL rewrite,  replacing it with grammars. I want to keep a clean set of dependencies for that precious new code :). So it only makes sense to isolate this logic into a seperate library.

To be honest, adding `apib` and `apib-parser` I think is very moderate, and I ended there to avoid bloating this PR.

We should continue splitting up the monolith of drafter introducing `apie`, `drafter-cli` next. There is even this old commit from Pavan (10d78fe7a) coming with the comment `librefract parts - will be separated into other project` in `./drafter.gyp` showing the intent of isolating API Elements logic even when it was developed.

Issue #694, opened 2019-05-08
fixed-type arrays produce anyOf in JSON Schema with single value


kylef:
### Source

```apib
# My API

## GET /

+ Response 200 (application/json)
    + Attributes (array[AnimalTarget], fixed-type)

## Data Structures

### AnimalTarget

+ type (required)
```

### Actual

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "array",
  "items": {
    "anyOf": [
      {
        "type": "object",
        "properties": {
          "type": {
            "type": "string"
          }
        },
        "required": [
          "type"
        ]
      }
    ]
  }
}
```

### Expected

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "type": {
        "type": "string"
      }
    },
    "required": [
      "type"
    ]
  }
}
```

PR #789, opened 2020-09-28
Optimise JSON Schemas generated for fixed-type arrays

kylef:
This changeset optimises the generation of JSON Schema of fixed-type array's with a single element inside them to be placed directly inside "items", instead of inside an "anyOf" inside "items". A user was asking about this behaviour in https://github.com/apiaryio/api-blueprint/issues/450.

In terms of JSON Schema validation these behave the same, the new behaviour produces a smaller, clearer JSON Schema.

While doing so, I noticed that Drafter would produce an invalid schema in the case where there is an empty fixed-type array. Drafter would emit a sub-schema such as `{"anyOf": []}` which breaks the clause in [JSON Schema 2019-09 Section 9.2.1.2](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.9.2.1.2):

> This keyword's value MUST be a non-empty array.

Instead Drafter will now emit a schema with `maxItems` set to 0 to indicate the array must be empty. This matches the intended behaviour from Drafter test specification:

> Given: An empty ArrayElement with fixedType attribute true
> When: a JSON Schema is generated from it
> Then: the schema matches an empty array

These two changes are separated into different commits thus I recommend reviewing per-commit.

Closes https://github.com/apiaryio/api-blueprint/issues/450
Closes https://github.com/apiaryio/drafter/issues/694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Drafter Changelog
 
+## TBD
+
+### Enhancements
+
+- JSON Schemas generated for `fixed-type` arrays with a single sub-type will
+  no longer be wrapped in an `anyOf` schema. Thus `array[Object]` as
+  `fixed-type` will now result in the following schema:
+
+  ```json
+  {
+    "type": "array",
+    "items": { "type": "object" }
+  }
+  ```
+
 ## 5.0.0 (2020-04-20)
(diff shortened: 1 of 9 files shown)


opichals (approved):
👍

